PR #17, opened 2026-02-26
docs: package contract layer design scaffold

henrycgbaker:
## Summary

- Adds design doc scaffolding the `core/` contract layer - canonical schemas, path conventions, config, + settings pattern
- ADR-0005 captures tooling rationale: Pydantic at boundaries, dataclasses for runtime, CSV interchange, per-tool settings

## Key decisions

- **Boundary schemas:** Pydantic models in `core/schema/` are SSOT; CSV/JSON are downstream serialisations
- **Runtime types:** dataclasses in `core/types/` - internal helpers, not contracts
- **Settings:** Per-tool settings modules (e.g. `querygen_settings.py`) with semantically scoped classes, bundled into `RunSettings`. Shared `ResolvableSettings` base class in `core/settings_base.py` provides recursive `resolve()` merge (overrides > config > defaults)
- **Paths:** `PathResolver` dataclass — single resolution point, tools compute their own output subdirs
- **Config:** Global only at `~/.chatboteval/config.yaml` for v1.0; optional (zero-config works out of the box)
- **Dependency direction:** `core/ ← api/ ← cli/`


docs/design/infra-package-contracts.md
```diff
@@ -0,0 +1,177 @@
...
+    └── commands/
+
+<base_dir>/                        # defaults to cwd
+  <tool>/
+    runs/
+      <run_id>/                    # per-run artefacts
+
+~/.chatboteval/
+  config.yaml                      # user config (optional)
+```
```
saschagobel: I would simplify this a bit. The goal here is to communicate the high-level contract layer structure, not every implementation detail. I’d suggest showing only `core/` with its main subpackages (`schemas/`, `types/`, `settings/`, `paths/`) and omitting files like `csv_io.py` as well as unrelated layers (`api/`, `cli/`). I’d also move the example config location under `<base_dir>` to align with the current workspace design rather than hardcoding `~/.chatboteval/`.

docs/design/infra-package-contracts.md
```diff
@@ -0,0 +1,177 @@
...
+
+Key API:
+
+- `WorkspacePaths.from_base_dir(path)` — normalises and resolves the workspace root
+- `tool_root(tool)` — returns `base_dir / tool`
+- `run_root(tool, run_id)` — returns `base_dir / tool / runs / run_id`
+- `under_base(path)` — resolves a path relative to `base_dir`
+- `coerce_path(str | Path)` — standalone helper for input normalisation
+
+Tool-specific path bundles (e.g. `QueryGenRunPaths` in `core/paths/querygen_paths.py`) are optional frozen dataclasses that pre-resolve all artefact paths for a run. A `resolve_<tool>_paths()` function constructs the bundle from `WorkspacePaths`. Tools with simple output structures can use `WorkspacePaths` directly.
```
saschagobel: This section reads more like an implementation specification than a design document. Listing specific method names and signatures introduces unnecessary maintenance overhead, since any small refactor would require updating the design doc. I’d keep this section at the conceptual level (workspace root, tool root, run root abstraction) and leave concrete APIs to the code and issue specs.

docs/design/infra-package-contracts.md
```diff
@@ -0,0 +1,177 @@
...
+### Tool-specific settings
+
+Each tool has a settings module in `core/settings/` containing:
+
+- **Semantically scoped settings classes** for logical groupings (e.g. `LlmSettings` for model config)
+- **A `RunSettings` class** that bundles the above plus run-level fields and inherits from `ResolvableSettings`
+
+`ResolvableSettings` in `core/settings/settings_base.py` is a Pydantic `BaseModel` (not `BaseSettings`) providing a `resolve()` classmethod that handles the precedence merge. Merge uses an `UNSET` sentinel to distinguish "not provided" from explicit values, with `deep_merge()` for recursive dict merging and `prune_unset()` to strip sentinel values. Each tool's `RunSettings` inherits this — no reimplementation per tool.
+
+The API entrypoint (`api/<tool>.py`) calls `RunSettings.resolve()` with caller-supplied overrides and config, then passes the resolved settings to the implementation layer.
```
saschagobel: Similar to the previous comment, this section documents fairly low-level implementation details (`UNSET`, `deep_merge`, `prune_unset`). That information already lives in the issue spec and code, and duplicating it here increases the maintenance surface. I would describe the conceptual mechanism (deterministic precedence resolution) and avoid detailing the exact helper functions.

saschagobel (changes requested):
Overall these are solid decision and design documents and the direction makes sense, but I think they currently go a bit too far into implementation detail and in a few places impose constraints that may be unnecessarily strict.

PR #18, opened 2026-02-27
feat: packaging invocation scaffold

### LINKED ISSUE #6: Scaffold packaging and invocation surfaces
**Goal**

Implement the packaging and invocation scaffold so the package is installable and the Python API and CLI invocation work end-to-end.

**Scope**

1. Create the package scaffold under `src/chatboteval/`:
- `api/`: stable high-level Python entrypoints (thin, user-facing orchestration surface)
- `cli/`: Typer app + commands (thin wrapper only)
- `core/`: internal implementation (may be empty placeholders for now)

2. Implement minimal runnable CLI:
- Add a Typer root app at `chatboteval/cli/app.py`
- Register at least one proof-of-wiring command:
  - `chatboteval --help` (Typer default)
  - `chatboteval --version` (prints version), implemented by calling an API function
- Wire a console-script entrypoint in `pyproject.toml`

3. Define the initial public Python surface:
- Top-level `chatboteval/__init__.py` should contain curated re-exports only

- Exactly one console-script entry point.
- Keep CLI thin: argument parsing + delegation only; no duplicated orchestration/business logic.
- Keep core/ non-public (do not re-export core symbols at top-level).
- Add one tiny real API function to validate wiring:
  - e.g., `chatboteval.api.info.get_version()` using `importlib.metadata` 

**Testing requirements**

- Import smoke: `import chatboteval` succeeds and curated symbol(s) exist
- CLI smoke: `chatboteval --help` exits 0
- CLI wiring: `chatboteval --version`exits 0 and matches expected output
- Boundary guard: lightweight import scan ensuring `chatboteval.cli` does not import `chatboteval.core` directly

**References**

- [ADR: Packaging and Invocation Surface](https://github.com/bertelsmannstift/chatboteval/blob/main/docs/decisions/0007-packaging-invocation-surface.md)
- [Design doc: Packaging and Invocation Surface](https://github.com/bertelsmannstift/chatboteval/blob/main/docs/design/packaging-invocation-surface.md)

lramir14:
**Brief summary**

Implements the packaging and invocation scaffold so the package is installable and both the Python API and CLI invocation work end-to-end.

**Detailed description**

Creates the full package scaffold under src/chatboteval/ with three layers: **api/** for stable high-level Python entrypoints, **cli/** as a thin Typer wrapper, and **core/** as a non-public internal implementation placeholder. 

Implements **api/info.py** with get_version() using importlib.metadata to validate wiring. Wires a single console-script entry point in pyproject.toml pointing to chatboteval.cli.app:app. The CLI delegates entirely to the API layer with no duplicated logic. A cli/commands/ directory is established as an empty package to enforce the convention that all future CLI subcommands live there, preventing logic from accumulating in app.py as the project grows. The top-level chatboteval/__init__.py exposes only curated re-exports via __all__. Adds pytest configuration and a dev dependency group for developer convenience.

**Tests cover:** import smoke, curated symbol existence, --help exit 0, --version wiring against the API, and an AST-based boundary guard ensuring cli never imports core directly.

**Status**

Ready for review

Closes #6


src/chatboteval/api/info.py
```diff
@@ -0,0 +1,15 @@
+"""Version resolution utilities for chatboteval."""
+
+from __future__ import annotations
+
+from importlib import metadata
+
+__all__ = ["get_version"]
```
saschagobel: `__all__ = ["get_version"]` in `api/info.py` is redundant since we already curate the API via `chatboteval/api/__init__.py` (and re-export top-level). I’d drop it to keep the module minimal.

tests/test_smoke.py
```diff
@@ -1,8 +1,58 @@
-"""Smoke tests."""
-
-
-def test_package_importable() -> None:
-    """Smoke test: the installed package can be imported."""
-    import chatboteval
-
-    assert chatboteval is not None
+"""Smoke tests."""
```
saschagobel: We repeatedly re-import `CliRunner`, `app`, and `chatboteval` across tests. Where there’s no strong need for per-test imports (I think there is for test_package_importable), I’d prefer module-level imports to reduce verbosity and repetition.

tests/test_smoke.py
```diff
@@ -1,8 +1,58 @@
...
+    assert result.exit_code == 0
+    assert get_version() in result.output
+
+
+def test_cli_does_not_import_core() -> None:
+    """Boundary guard: cli layer must not directly import core."""
+    import ast
+    import pathlib
+
+    cli_dir = pathlib.Path(__file__).parent.parent / "src" / "chatboteval" / "cli"
```
saschagobel: The path construction `Path(__file__).parent.parent / "src" / ...` hardcodes the repo layout. Consider deriving the CLI directory from the imported package instead (`Path(chatboteval.cli.__file__).resolve().parent`).

PR #25, opened 2026-03-05
feat(infra): add Argilla dev Docker Compose stack and Makefile

henrycgbaker:
## Goal

One-command dev environment for the Argilla annotation stack. Devs can start, stop, reset, and smoke-test the full stack w/ `make` targets. *NB this is for devs; user-facing production setup requires its own design.*

*NB: docs/design/annotation-interface.md diagram was surfaced in merge conflict, adds back in pending decision from SG*

## Scope

- `deploy/annotation/docker-compose.dev.yml` — 5-service stack (Argilla, worker, Postgres, Elasticsearch, Redis). Derived from [Argilla compose example](https://github.com/argilla-io/argilla/blob/main/examples/deployments/docker/docker-compose.yaml)
- `deploy/annotation/.env.dev.example` — dev environment defaults
- `Makefile` — lifecycle targets (setup, teardown, stop, logs, status, test-stack)
- `.gitignore` — allow `.env.*.example` files in version control
- `tests/infra/test_dev_stack.py` — structural tests (compose file, services, healthchecks, volumes, env vars, Makefile targets)
- `tests/infra/test_dev_stack_integration.py` — integration tests (boot stack, hit API, verify teardown)

## Implementation

- All services have health checks + named volumes
- `make setup` is idempotent: auto-copies env template, uses `--wait` to block until healthy (Argilla takes ~10-30s to ready — Postgres needs to initialise, ES needs to build indices etc)
- `make teardown` removes containers & volumes → clean state
- `make test-stack` runs full cycle: setup → authenticated API call to `/api/v1/me` via python3 urllib → teardown
- Argilla healthcheck uses `python3 -c` with urllib (not curl — removed from `argilla-server:latest` image). Hits `/api/v1/me` with API key header to exercise DB + auth
- Integration-marked tests auto-skip when Docker CLI is unavailable (via `conftest.py` fixture)

## Testing

*NB: integration tests have Docker dependency → install Docker Desktop to run locally*

### Automated (pytest)
- `pytest tests/infra/test_dev_stack.py` — 6 structural tests, no Docker required
- `pytest tests/infra/test_dev_stack_integration.py` — 3 integration tests, auto-skip without Docker

### Manual
- [x] `make setup` brings up all 5 services, Argilla UI reachable at localhost:6900
- [x] `make teardown && make setup` produces clean environment
- [x] `make stop` preserves volumes; `make setup` after stop resumes without data loss
- [x] `make test-stack` passes end-to-end

## References

- `docs/design/infra-package-contracts.md`

## NB 5 services

Argilla v2 requires server + PostgreSQL + Elasticsearch + Redis:
- PostgreSQL & Elasticsearch from before
- **Redis** also for background job queue
   - `ARGILLA_REDIS_URL` has default value (`redis://localhost:6379/0`) → server expects Redis instance on startup ([server config](https://docs.argilla.io/v2.4/reference/argilla-server/configuration/))
- **Worker** is separate process consuming from Redis for async tasks (bulk imports, distribution). Included in the [official Compose file](https://github.com/argilla-io/argilla/blob/main/examples/deployments/docker/docker-compose.yaml).
- Argilla deployment docs reference the 3 core components from before (server, database, search engine) but the official Compose file ships all 5, w/ Redis as hard dependency

### DIFF
--- a/deploy/annotation/docker-compose.dev.yml
+++ b/deploy/annotation/docker-compose.dev.yml
@@ -1,6 +1,6 @@
 services:
   argilla:
-    image: argilla/argilla-server:latest
+    image: argilla/argilla-server:v2.8.0
     restart: unless-stopped
     ports:
       - "${ARGILLA_PORT:-6900}:6900"
@@ -30,7 +30,7 @@ services:
       - argilla
 
   worker:
-    image: argilla/argilla-server:latest
+    image: argilla/argilla-server:v2.8.0
     restart: unless-stopped
     command: ["python", "-m", "argilla_server.cli", "worker", "--num-workers", "2"]
     environment:
--- a/tests/infra/test_dev_stack_integration.py
+++ b/tests/infra/test_dev_stack_integration.py
@@ -15,7 +15,7 @@ def test_stack_boots_healthy() -> None:
         resp = urllib.request.urlopen("http://localhost:6900/api/docs", timeout=10)
         assert resp.status == 200, f"Argilla docs returned {resp.status}"
     finally:
-        subprocess.run(["make", "teardown"], capture_output=True, timeout=60)
+        subprocess.run(["make", "teardown"], check=True, capture_output=True, timeout=60)
 
 
 @pytest.mark.integration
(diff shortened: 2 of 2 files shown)


pyproject.toml
```diff
@@ -30,6 +28,7 @@ build-backend = "hatchling.build"
 
 [tool.pytest.ini_options]
 testpaths = ["tests"]
+markers = ["integration: mark test as requiring a live Docker/Argilla stack"]
```
saschagobel: Small forward-looking thought: the current autouse skip logic is tied to the generic integration marker, so any future integration test would also inherit the Docker requirement semantics. It may be worth considering a more specific marker like annotation_integration if we expect other integration-test families later.

Also, note: the “auto-skip without Docker” behaviour appears to depend on the conftest.py added in PR #26 rather than this PR itself.

PR #42, opened 2026-03-11
feat(core): add annotation import boundary schemas

henrycgbaker:
## Summary
- Adds `core/schemas/annotation_import.py` with `Chunk` and `QueryResponsePair` Pydantic models
- Frozen, strict validation: non-empty text fields, chunk_rank >= 1, min 1 chunk
- Inlined `ConfigDict(frozen=True, extra="forbid")` per review feedback (no shared base class)

## Test plan
- [x] 21 unit tests covering construction, validation rejection, immutability, extra field rejection

## References
- Split from #33 per reviewer request
- Design: `docs/design/annotation-import-pipeline.md`

### DIFF
--- a/docs/design/annotation-user-management.md
+++ b/docs/design/annotation-user-management.md
@@ -6,7 +6,7 @@ Argilla user account provisioning and credential management for the annotation p
 
 **In scope:**
 - Create and assign annotator accounts in Argilla
-- Assign users to workspaces (workspace names are deployment config; defaults: `retrieval_grounding`, `generation`)
+- Assign users to workspaces (workspace names are deployment config; defaults: `retrieval`, `grounding`, `generation`)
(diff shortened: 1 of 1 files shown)


saschagobel (approved):
Looks good overall.

One small consistency thought: in `src/chatboteval/core/schemas/querygen_output.py` we already use `NonEmptyStr = Annotated[str, StringConstraints(strip_whitespace=True, min_length=1)]`. It might be worth moving that to a shared module such as `src/chatboteval/core/types.py` and reusing that pattern here instead of local validators, so we have one consistent idiom for “non-empty stripped string” across schemas.

Minor readability nit: the module helper `_non_empty` and the validator methods share the same name, which works but is slightly confusing when reading the file. This would become irrelevant if you adopt the shared `NonEmptyStr` pattern.

PR #51, opened 2026-03-13
feat(annotation): add dataset schemas and setup settings

henrycgbaker:
## Goal

Define the Argilla dataset schemas and annotation setup settings as pure Python objects — no Argilla server interaction needed. These are the contracts that the setup API (next PR) consumes.

## Scope

- `AnnotationSetupSettings` — workspace topology (3-workspace default), prefix, min_submitted
- `UserSpec` dataclass — user provisioning input
- Three `rg.Settings` constants with field visibility contract per ADR-0009
- Collapsible `<details>/<summary>` HTML template for supporting-context fields
- `TASK_SETTINGS` and `DATASET_NAMES` lookup dicts
- Full unit test coverage (43 tests, no Argilla server required)

## Implementation

**`api/annotation_task_config.py`** — Three `rg.Settings` module-level constants matching `annotation-interface.md`: primary fields as `TextField`, supporting context as `CustomField` with collapsible HTML. Binary `LabelQuestion`s per task with EN wording from the design doc. Hardcoded per ADR-0009.

**`core/settings/annotation_settings.py`** — `AnnotationSetupSettings(ResolveSettings)` with `workspace_prefix`, `workspace_dataset_map` (default: 1 workspace per task), `min_submitted`. `UserSpec` dataclass for user provisioning.

**`tests/annotation/conftest.py`** — Patches `rg.Argilla._default_client` so unit tests run without a live server.

**`pyproject.toml`** — Added `per-file-ignores` for test docstrings (D100-D103).

## Testing

```bash
python -m pytest tests/annotation/test_schemas.py tests/annotation/test_setup_settings.py -v
```

## References

- `docs/design/annotation-interface.md` — 3 tasks, field visibility, question wording
- `docs/decisions/0009-annotation-schema-configurability.md` — schemas hardcoded in code
- `docs/decisions/0010-annotation-multi-dataset-architecture.md` — one dataset per task
- Stacks on #50

### DIFF
--- a/src/chatboteval/core/schemas/querygen_realize.py
+++ b/src/chatboteval/core/schemas/querygen_realize.py
@@ -1,0 +1,20 @@
+"""Structured output contracts for LLM stage 2 query realization."""
+
+from pydantic import BaseModel, ConfigDict, Field
+
+
+class RealizedQuery(BaseModel):
+    """One realized query aligned to a stage 2 candidate."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    candidate_id: str = Field(description="Candidate identifier preserved from the stage 2 input blueprint.")
+    query: str = Field(description="Realized user query text for the stage 2 candidate.")
+
+
+class RealizedQueryList(BaseModel):
+    """Collection of realized queries aligned to stage 2 candidates."""
+
+    model_config = ConfigDict(extra="forbid")
+
+    queries: list[RealizedQuery] = Field(description="Realized queries aligned one-to-one with the stage 2 candidates.")
--- a/tests/test_querygen_realize.py
+++ b/tests/test_querygen_realize.py
@@ -1,0 +1,106 @@
+"""Unit tests for the stage 2 query realization structured output contract."""
+
+import pytest
+from pydantic import ValidationError
+
+from chatboteval.core.schemas.querygen_realize import RealizedQuery, RealizedQueryList
+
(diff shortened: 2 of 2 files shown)


henrycgbaker:
Note: this supersedes the earlier `feat/annotation-dataset-schemas` branch, which contains the full iterative commit history for this feature (now deleted).

src/chatboteval/core/settings/annotation_settings.py
```diff
@@ -0,0 +1,34 @@
...
+
+
+@dataclass
+class UserSpec:
+    """Specification for provisioning an Argilla user account."""
+
+    username: str
+    role: Literal["owner", "annotator"]
+    workspaces: list[str] = field(default_factory=list)
+    password: str | None = None
```
saschagobel: Structural question: `UserSpec` feels slightly different in kind from the rest of this module. `AnnotationSettings` is clearly runtime configuration, whereas `UserSpec` reads more like an operational data contract for provisioning users.

I can also see the argument for keeping it here, since it is closely tied to annotation setup and does provide defaults for that workflow. So I’m not strongly opposed to the current placement — just noting that it mixes “settings” and “specification” concepts a bit.

If we did want to separate those concerns more cleanly later, I could imagine `UserSpec` living in `core/schemas/annotation_user.py` if that makes more sense. Not blocking.

PR #59, opened 2026-03-15
refactor(core): move shared Pydantic model TypeVar M to core.types

saschagobel:
**Summary**

- move the shared `M = TypeVar("M", bound=BaseModel)` from `core/csv_io.py` to `core/types.py`
- keep `NonEmptyStr` and generic Pydantic typing utilities together in one shared module
- update `csv_io.py` to import M from `chatboteval.core.types`

**Why**

- `M` is a reusable core typing utility rather than a CSV-specific concern. Centralising it in `core/types.py` avoids duplicate generic model type definitions and gives future schema-related modules one canonical import location.

### DIFF
--- a/src/chatboteval/core/types.py
+++ b/src/chatboteval/core/types.py
@@ -1,4 +1,4 @@
-"""Shared type aliases andy typing utilities for core modules."""
+"""Shared type aliases and typing utilities for core modules."""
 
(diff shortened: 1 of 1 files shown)


henrycgbaker (approved):
fixed minor typo

PR #63, opened 2026-03-16
feat(annotation): add core setup implementation

henrycgbaker:
## Summary

- Add `core/annotation/argilla_ops.py` — idempotent create/lookup wrappers for workspaces, datasets, users
- Add `core/annotation/setup.py` — `setup_datasets()`, `provision_users()`, `teardown_resources()`

Stacked on #51.

## Changes from review

- Moved modules into `core/annotation/` package (SG comment #1)
- Fixed `provision_users()` idempotency — workspace membership reconciled on rerun, not just on creation (SG comment #2)
- Removed `include_users` from `teardown_resources()` — user deletion was too broad, scoped user lifecycle deferred (SG comment #3)
- Demoted "already exists" logs from warning to info (SG comment #4)
- Deferred `SetupResult._merge()` to PR #52 where its caller lives (SG comment #5)
- `SetupResult` stays in `core/` — `api/` can't define it without breaking dependency direction (SG comment #6)

## Scope

| File | Purpose |
|------|---------|
| `core/annotation/argilla_ops.py` | `apply_prefix`, `create_workspace`, `create_dataset`, `create_user` |
| `core/annotation/setup.py` | `SetupResult`, `setup_datasets`, `provision_users`, `teardown_resources` |
| `tests/annotation/test_setup_unit.py` | Pure Python tests (no server) |
| `tests/annotation/test_setup_integration.py` | Live Argilla tests (`@pytest.mark.integration`) |

## References

- `docs/design/annotation-user-management.md`
- `docs/design/annotation-interface.md`
- ADR-0009, ADR-0010

PR #64, opened 2026-03-16
feat(annotation): add core import implementation

henrycgbaker:
## Summary

- Add `core/annotation/record_builder.py` — record building and fan-out logic for the 3 annotation datasets
- Content-hash UUID derivation for idempotent upsert

Stacked on #63.

## Scope

| File | Purpose |
|------|---------|
| `core/annotation/record_builder.py` | `derive_record_uuid`, `build_*_records`, `fan_out_records` |
| `tests/annotation/test_import_unit.py` | Pure Python tests (no server) |

## Open question

`context_set` as supporting context in generation dataset — see PR comment for discussion.

## References

- `docs/design/annotation-import-pipeline.md`
- ADR-0010

### DIFF
--- a/tests/annotation/test_import_unit.py
+++ b/tests/annotation/test_import_unit.py
@@ -63,7 +63,7 @@ class TestDeriveRecordUuid:
     def test_returns_hex_string(self) -> None:
         result = derive_record_uuid(_make_pair())
         assert isinstance(result, str)
-        assert len(result) > 0
+        assert len(result) == 64
         assert all(c in "0123456789abcdef" for c in result)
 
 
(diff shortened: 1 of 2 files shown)


tests/annotation/test_import_unit.py
```diff
@@ -0,0 +1,194 @@
...
+        raw = _valid_raw()
+        pair_a = QueryResponsePair.model_validate(raw)
+        raw_reversed = {**raw, "chunks": list(reversed(raw["chunks"]))}
+        pair_b = QueryResponsePair.model_validate(raw_reversed)
+        assert derive_record_uuid(pair_a) == derive_record_uuid(pair_b)
+
+    def test_returns_hex_string(self) -> None:
+        result = derive_record_uuid(_make_pair())
+        assert isinstance(result, str)
+        assert len(result) > 0
```
saschagobel: Nit: since this is SHA-256, you could assert `len(result) == 64` instead of just > 0 to make the contract a bit tighter.

PR #71, opened 2026-03-19
chore: configure Dependabot

saschagobel:
**Summary**

Adds Dependabot configuration to keep Python dependencies and GitHub Actions up to date.

**Details**

- Configuration lives at `.github/dependabot.yml`
- Uses `uv` ecosystem
- Monthly update cadence to minimize churn and preserve evaluation stability
- Groups minor and patch updates into a single PR
- Leaves major updates ungrouped for explicit review
- Includes GitHub Actions dependency updates

**Status**

Ready for review.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -22,6 +22,7 @@ updates:
     directory: "/"
     schedule:
       interval: "monthly"
+    open-pull-requests-limit: 5
     commit-message:
       prefix: "chore(ci)"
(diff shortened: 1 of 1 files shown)


henrycgbaker:
non-blocking comment: could also assign a reviewer (otherwise auto-assigns Dependabot), I don't know if that's considered more transparent or noisy, but it means dependabot gets added as a contributor

PR #72, opened 2026-03-19
feat: implement LLM composition boundary for synthetic query generation

### LINKED ISSUE #61: Implement LLM composition boundary for synthetic query generation
**Goal**

Implement the LLM composition layer for synthetic query generation so planning and realization executors can construct schema-constrained LangChain runnables from prompt text and model settings without duplicating model setup logic.

**Scope**

- Implement the query generation LLM composition module in:
  - `src/chatboteval/core/querygen/llm.py`

- Deliverables include:
  - `_build_prompt_template()` internal prompt-template helper
  - `build_chat_runnable()` runnable composition with structured output and retry policy

- `_build_prompt_template()`
  - small internal helper for prompt construction
  - builds a `ChatPromptTemplate` from system and user text via `ChatPromptTemplate.from_messages()`

- `build_chat_runnable()`
  - main composition function, that accepts
    - `system_text`, `user_text`, `model_provider`, `model`, `api_key`, `output_schema` (Pydantic schema class), `requests_per_second`, `check_every_n_seconds`, `max_bucket_size`
  - create the chat model via `init_chat_model()`
    - inputs forwarded from `gen_queries` into composition layer as a plain dict
    - construct a kwargs dict for model initialization and call `init_chat_model(**init_kwargs)`
    - use dynamic kwargs assembly with conditional insertion where appropriate
    - implement rate limiting via `InMemoryRateLimiter`
  - apply `.with_structured_output()` using the provided Pydantic schema
  - apply `.with_retry()` for retry behaviour
  - build the prompt via `_build_prompt_template()`
  - compose prompt + structured model into a runnable, e.g. `prompt | structured_chat_model`
  - return the composed runnable without invoking it

- wire rate limit settings into settings and API
  - add to `gen_queries()`
    - `requests_per_second: float | object = UNSET`
    - `check_every_n_seconds: float | object = UNSET`
    - `max_bucket_size: int | object = UNSET`
  - expose these settings in `QueryGenRunSettings`
    

**Testing requirements**

- Unit tests required covering:
  - `_build_prompt_template()` returns a `ChatPromptTemplate` with the expected message structure
  - `build_chat_runnable()` constructs the runnable pipeline correctly
  - `init_chat_model()` is called with the expected kwargs
  - structured output wrapping is applied
  - retry wrapping is applied
  - the returned object is a composed runnable and no invocation occurs during construction
- avoid real provider calls in unit tests

saschagobel:
**Summary**

Introduce the LLM composition boundary for synthetic query generation and wire rate limiting through settings and API.

**Key changes**
- Add `core/querygen/llm.py`:
  - `_build_prompt_template()` for prompt construction
  - `build_llm_runnable()` for composing:
    - `init_chat_model(...)`
    - structured output (`.with_structured_output`)
    - retry policy (`.with_retry`)
    - `InMemoryRateLimiter`

- Extend LlmSettings with rate limiting options:
  - `requests_per_second`
  - `check_every_n_seconds`
  - `max_bucket_size`

- Wire rate limit settings through:
  - `QueryGenRunSettings` via `LlmSettings`
  - `gen_queries()` API

- Add unit tests for LLM composition:
  - prompt template construction
  - model init kwargs assembly
  - structured output and retry wrapping
  - construction-only runnable composition

**Status**

Ready for review.

Closes #61

### DIFF
--- a/tests/unit/core/querygen/test_llm.py
+++ b/tests/unit/core/querygen/test_llm.py
@@ -176,6 +176,7 @@ def test_build_llm_runnable_propagates_init_error(
             model_kwargs={},
         )
 
+
 @pytest.mark.parametrize(
     "reserved_key",
     ["api_key", "base_url", "model", "model_provider", "rate_limiter"],
--- a/tests/unit/core/settings/test_querygen_settings.py
+++ b/tests/unit/core/settings/test_querygen_settings.py
@@ -2,6 +2,7 @@
 
 from pathlib import Path
 from typing import Any
+
 from pydantic import ValidationError
 
 from pragmata.core.settings.querygen_settings import LlmSettings, QueryGenRunSettings
@@ -119,6 +120,7 @@ def test_querygen_run_settings_model_kwargs_merge_semantics() -> None:
         "top_p": 0.9,
         "max_tokens": 300,
     }
+
 
 def test_llm_settings_rejects_invalid_rate_limiter_values() -> None:
     """LlmSettings rejects invalid rate limiter configuration values."""


pyproject.toml
```diff
@@ -15,6 +15,7 @@ dependencies = [
     "typer>=0.9",
     "pydantic>=2.0,<3.0",
     "pyyaml>=6.0",
+    "langchain>=1.0.8",
```
lramir14: One thing I’d like to clarify is the dependency/support story for providers. The querygen settings expose provider-specific configurations and the default provider is `mistralai`. I am thinking whether `langchain` alone is sufficient for the providers we intend to support here, or whether the relevant provider integrations should also be declared/documented explicitly so installation and runtime behavior stay aligned.

lramir14 (commented):
I like the new LLM composition boundary overall. A couple of things I’d flag are that `model_kwargs` currently works as a side door for overriding reserved settings.

Because the function first defines the canonical init args and then does:

```python
init_kwargs.update(model_kwargs)

So then, these settings (model, model_provider, api_key, rate_limiter, or base_url) can change with kwargs after being defined. 

Additionally, I think it would be good to validate the new rate-limiter settings at the settings layer as well.

Right now, requests_per_second, check_every_n_seconds, and max_bucket_size are added as plain numeric fields, so invalid values could pass through settings resolution and only fail later during runtime.

pyproject.toml
saschagobel: Good point. In this PR we only build the runnable (no invocation yet), so `langchain` is sufficient.

Provider integrations will be added in a follow-up PR that actually invokes the models. Plan is to include the default provider as a core dependency and expose others via optional extras.

PR #87, opened 2026-03-25
test: align annotation integration tests with marker taxonomy and module-level marking

saschagobel:
**Summary**

Aligns existing annotation integration tests with the new pytest marker taxonomy and module-level marking pattern.

**Key changes**

- Apply module-level markers to annotation integration tests
  - Add `pytestmark = [pytest.mark.integration, pytest.mark.annotation]` across annotation-related integration modules
  - Remove per-test `@pytest.mark.integration` decorators

- Enforce dual-marker taxonomy
  - Ensure all annotation integration tests are consistently marked with both `integration` and `annotation`

- Update test module docstrings
  - Reflect new marker-based invocation pattern:
    - `-m "integration and annotation"`

- Maintain compatibility with updated conftest gating
  - Annotation tests correctly trigger stack availability checks
  - Non-annotation integration tests (e.g. querygen) remain unaffected

**Note**

Requires PR #86 to be merged first, then rebase onto `main`.

**Status**

Ready for review.

### DIFF
--- a/tests/integration/test_annotation_import.py
+++ b/tests/integration/test_annotation_import.py
@@ -1,6 +1,6 @@
 """Integration tests for annotation import against a live Argilla server.
 
-Run with: pytest tests/annotation/test_import_integration.py -m "integration and annotation" -v
+Run with: pytest tests/integration/test_annotation_import.py -m "integration and annotation" -v
 Requires: make setup (Argilla stack running on localhost:6900)
 """
 


henrycgbaker:
@saschagobel 

Frustratingly I lost my previous comments on this which were more comprehensive, however I'll raise the core points/clarifications here.

- would it not make more sense to have a distinct `docker` marker. Here `integration` is doing double work.
- so I would suggest that we have unit and integration (scope-level), then domain-level (annotation, querygen, eval), and layer on top of that a infra dep related `docker` marker
- then default exclusion becomes `-m "not docker"` rather than `-m "not integration"`. It's another marker to maintain but it's more explicit (and I think clearer).

---

Additionally this PR breaks annotation tests briefly as `annotation` marker is not registered. End of last week this was rectified by a follow up PR that introduced it, but I don't see that anywhere this week - has it been closed?

PR #94, opened 2026-03-26
refactor: introduce typed UNSET sentinel and remove casts

saschagobel:
**Summary**

Replace the untyped `UNSET = object()` sentinel with a dedicated typed singleton and update API surfaces to use `Unset` for override parameters. Refactor config loading to use `isinstance`-based narrowing, removing all `cast(...)` usage.

**Key changes**
- Introduce `_UnsetType` sentinel and `type Unset = _UnsetType`
- Replace `| object` with `| Unset` across API functions
- Remove `cast(...)` usage in config loading
- Switch to `isinstance(..., (str, Path))` for mypy-safe narrowing
- Apply changes consistently across querygen and annotation APIs

No behavioral changes; improves type safety and readability.

**Status**

Ready for review.

### DIFF
--- a/src/pragmata/core/settings/settings_base.py
+++ b/src/pragmata/core/settings/settings_base.py
@@ -16,6 +16,9 @@ class _UnsetType:
     def __repr__(self) -> str:
         return "UNSET"
 
+    def __bool__(self) -> bool:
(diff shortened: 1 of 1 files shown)


saschagobel:
> non-blocking: prune_unset still uses `is not UNSET` as identity check; rest of PR moves to `isinstance` narrowing -> worth switching to isinstance(item, _UnsetType) here too?

Yes, adjusted.

PR #100, opened 2026-03-27
feat(infra): support external backing services via Compose profiles

henrycgbaker:
## Goal

Allow Argilla to run against externally managed backing services (Postgres, Elasticsearch, Redis) on a per-service basis, while keeping the default all-bundled dev workflow unchanged (zero config setup principle).

## Scope

Infrastructure only - no Python code changes.

- `deploy/annotation/docker-compose.dev.yml`
- `deploy/annotation/.env.dev.example`
- `Makefile`

### Named Compose profiles

Rather than binary bundled/external switch, this further adds backing services assigned to named profiles that model the realistic deployment scenarios:

| Make target | Profile | Postgres | Elasticsearch | Redis |
|---|---|---|---|---|
| `make docker-up` | `all-bundled` | bundled | bundled | bundled |
| `make docker-up-external-pg` | `external-pg` | **external** | bundled | bundled |
| `make docker-up-external-es` | `external-es` | bundled | **external** | bundled |
| `make docker-up-external` | _(none)_ | **external** | **external** | **external** |

`external-redis` is omitted deliberately. Redis is a job queue not data store -> least likely service to be centrally managed (named profiles reflect realistic deployment scenarios rather than every possible combo).

The rationale for `external-pg` and `external-es` specifically: Postgres is ubiquitous and most orgs already run it centrally; Elasticsearch holds the annotation record data and is therefore the highest-value service to persist externally.

### Profile assignments in Compose

```yaml
postgres:      profiles: [all-bundled, external-es]
elasticsearch: profiles: [all-bundled, external-pg]
redis:         profiles: [all-bundled, external-pg, external-es]
```

`argilla` and `worker` have no profile - they always start. `depends_on` entries use `required: false` so they start regardless of whether backing containers are present.

### Env var interpolation

Backing service URLs use `${VAR:-default}` on argilla and worker - external URLs override bundled container addresses when set in `.env`.

### Management targets

`docker-down`, `docker-stop`, `docker-logs`, `docker-status` all pass all three profiles via `ALL_PROFILES` so they cover whichever containers are running regardless of which `up` variant was used.

## Testing

- `make docker-up` starts all 5 containers (unchanged)
- `make docker-up-external-pg` starts argilla + worker + ES + Redis
- `make docker-up-external-es` starts argilla + worker + Postgres + Redis
- `make docker-up-external` starts argilla + worker only
- `.env.dev.example` documents override vars and maps each to the correct make target

## References

- `docs/design/annotation-interface.md` - Argilla as annotation backend
- Requires Docker Compose v2.20.2+ (`depends_on required: false` support)


saschagobel (approved):
All good, thanks, just two questions (non-blocking).

saschagobel (approved):
All good now, thanks!

PR #103, opened 2026-03-31
feat: implement stage-2 realization executor for synthetic query generation

### LINKED ISSUE #99: Implement stage-2 realization executor for synthetic query generation
**Goal**

Implement the stage-2 realization executor for the synthetic query generation workflow.

**Scope**

- Implement the realization stage executor in:
  - `src/pragmata/core/querygen/realization.py`
- Deliverables:
  - `_build_realization_prompt_vars()` helper for constructing the stage-2 invoke payload from selected stage-1 candidate blueprints
  - `run_realization_stage()` executor function for one realization-stage invocation
- Batching, filtering/deduplication, and final synthetic-query assembly are out of scope.

- `_build_realization_prompt_vars()`
  - implement as `_build_realization_prompt_vars(candidates: list[QueryBlueprint]) -> dict[str, object]`
  - consumes the selected stage-1 candidate blueprints for this single realization invocation
  - returns a dict with a single key `"query_blueprints"`
  - the value for `"query_blueprints"` must be a human-readable formatted string representation of the candidate blueprints (not raw dict/JSON)
  - the formatted representation must:
    - include all required blueprint fields (including candidate_id)
    - follow a consistent field order across candidates
    - include optional fields only when present

- `run_realization_stage()`
  - implement as `run_realization_stage(candidates: list[QueryBlueprint], llm_settings: LlmSettings, api_key: str) -> list[RealizedQuery]`
  - calls `_build_realization_prompt_vars()`
  - constructs the stage-2 runnable using the existing realization prompt assets and `build_llm_runnable()`
  - passes `output_schema=RealizedQueryList` when calling `build_llm_runnable()`
  - invokes the runnable exactly once via `.invoke(...)`, passing the invoke payload
  - extracts and returns `RealizedQueryList.queries` as list[RealizedQuery]`

**Testing requirements**

- Unit tests required covering, for example:
  - prompt-payload construction from selected `QueryBlueprint`s
  - correct wiring of realization-stage prompt texts and realization-model settings into `build_llm_runnable()` using standard mocking/patching
  - passing `output_schema=RealizedQueryList`
  - one-shot invocation behavior of `run_realization_stage()` (optionally use LangChain fake chat models for deterministic execution tests)
  - extraction of `.queries` and return type `list[RealizedQuery]`

saschagobel:
**Summary**

Implement the stage-2 realization executor for synthetic query generation, including blueprint prompt payload construction, LLM invocation wiring, and structured realized-query extraction.

**Key changes**

- Add `core/querygen/realization.py`:
  - lightweight blueprint formatting for prompt injection with:
     - deterministic field order
    - inclusion of all required blueprint fields
    - conditional inclusion of optional fields only when present
  - `_build_realization_prompt_vars()` for constructing the stage-2 prompt payload from selected `QueryBlueprints`
  - `run_realization_stage()` for:
    - composing the realization runnable via `build_llm_runnable(...)`
    - enforcing `RealizedQueryList` structured output
    - `.invoke(...)` execution
    - returning `list[RealizedQuery]`
  - introduce `RealizationStageError` and wrap `llm_runnable.invoke(...)` failures with a clear, user-facing error

- Add comprehensive unit tests for realization stage:
  - prompt payload construction from selected blueprints
  - correct formatting of single and multiple blueprint blocks
  - conditional handling of optional blueprint fields
  - exact placeholder mapping for stage-2 prompt variables
  - correct wiring of realization prompt assets and LLM settings
  - structured output schema usage (`RealizedQueryList`)
  - single-invocation execution behavior
  - realized-query extraction and return type validation
  - error propagation vs. wrapping (`LlmInitializationError`, `RealizationStageError`)

**Status**

Ready for review.

Note: Requires merging PR #101 

Closes #99

### DIFF
--- a/src/pragmata/core/querygen/realization.py
+++ b/src/pragmata/core/querygen/realization.py
@@ -1,0 +1,113 @@
+"""Stage-2 realization executor for synthetic query generation."""
+
+from pragmata.core.querygen.llm import LlmInitializationError, build_llm_runnable
+from pragmata.core.querygen.prompts import SYSTEM_PROMPT_REALIZATION, USER_PROMPT_REALIZATION
+from pragmata.core.schemas.querygen_plan import QueryBlueprint
+from pragmata.core.schemas.querygen_realize import RealizedQuery, RealizedQueryList
+from pragmata.core.settings.querygen_settings import LlmSettings
+
+
+class RealizationStageError(RuntimeError):
+    """Raised when a realization-stage invocation fails."""
+
+
+def _format_blueprint(
+    candidate: QueryBlueprint,
+) -> str:
+    """Format one query blueprint for prompt injection.
+
+    Args:
+        candidate: Structured candidate query blueprint selected for realization.
+
+    Returns:
+        A human-readable multiline representation of the blueprint with
+        deterministic field order and optional fields included only when present.
+    """
+    lines = [
+        f"- candidate_id: {candidate.candidate_id}",
+        f"  domain: {candidate.domain}",
+        f"  role: {candidate.role}",
+        f"  language: {candidate.language}",
+        f"  topic: {candidate.topic}",
+        f"  intent: {candidate.intent}",
+        f"  task: {candidate.task}",
+    ]
+
+    if candidate.difficulty is not None:
+        lines.append(f"  difficulty: {candidate.difficulty}")
+
+    if candidate.format is not None:
+        lines.append(f"  format: {candidate.format}")
+
+    lines.extend(
+        [
+            f"  user_scenario: {candidate.user_scenario}",
+            f"  information_need: {candidate.information_need}",
+        ]
+    )
+
+    return "\n".join(lines)
+
+
+def _build_realization_prompt_vars(
+    candidates: list[QueryBlueprint],
+) -> dict[str, object]:
+    """Build invoke-time prompt variables for one realization-stage batch.
+
+    Args:
+        candidates: Selected stage-1 candidate blueprints for this single
+            realization invocation.
+
+    Returns:
+        A dict containing a single ``query_blueprints`` key whose value is a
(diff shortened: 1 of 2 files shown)


lramir14 (approved):
I think it is a nice implementation. I do have one doubt: RealizedQueryList guarantees the output shape, but not that the realized queries still align with the input candidates. Since run_realization_stage() returns llm_output.queries directly, we could silently accept malformed stage-2 outputs if the model returns fewer items, duplicate candidate_ids, or unexpected IDs. I am thinking if we should validate one-to-one alignment with the selected candidates before returning?

saschagobel:
> I think it is a nice implementation. I do have one doubt: RealizedQueryList guarantees the output shape, but not that the realized queries still align with the input candidates. Since run_realization_stage() returns llm_output.queries directly, we could silently accept malformed stage-2 outputs if the model returns fewer items, duplicate candidate_ids, or unexpected IDs. I am thinking if we should validate one-to-one alignment with the selected candidates before returning?

Good point, and I agree this is a real failure mode in principle.

In our current design, we deliberately keep `run_realization_stage()` as a thin execution unit (prompt construction → LLM invocation → structured extraction), and handle alignment deterministically in the downstream filtering step. Specifically, the candidate-ID alignment filtering introduced in PR #107 removes outputs with missing, duplicated, or misordered IDs based on positional agreement.

So we intentionally do not enforce one-to-one alignment inside the stage itself, as that would mix concerns and potentially be wasteful (we’d reject entire batches rather than recover valid items).

For the second class of issues (semantically incorrect realizations despite correct IDs), we currently treat this as part of the expected residual error of LLM generation. Given the overall pipeline is designed around human annotation as ground truth , these cases are caught downstream. If this turns out to be frequent, we’d likely address it with an additional automated review/validation stage rather than tightening this executor.

PR #106, opened 2026-04-01
feat(annotation): add structured logging across annotation tool

henrycgbaker:
## Goal

Add structured Python logging across annotation tool (querygen handled separately) so operations are observable via CLI output and container logs (Docker captures stderr by default).

## Scope

- **Library root** — `NullHandler` on `pragmata` logger (standard lib convention)
- **CLI** — `--verbose` / `-v` flag (`-v`=INFO, `-vv`=DEBUG, default=WARNING) via `logging.basicConfig` to stderr
- **Core** — loggers + log calls in `export_runner` (CSV write success/failure, rollback) and `loaders` (record count after file load). Existing loggers in `argilla_ops`, `setup`, `record_builder`, `export_fetcher` unchanged.
- **API** — orchestration-level summaries: import completion + validation warnings, setup counts, teardown start, export totals. No duplication of core-level logs.

## Implementation

Follows Python logging best practice for libs:
- `core/` emits logs at point of context (where errors occur, where I/O happens)
- `api/` emits orchestration summaries only
- Logger hierarchy propagates automatically — no manual error forwarding
- CLI configures the root handler; library consumers configure their own

## Testing

- [x] All files compile (`py_compile`)
- [x] Ruff lint passes (pre-commit hook)
- [ ] Existing test suite (CI)

## References

- Python logging HOWTO: library pattern with `NullHandler`
- Architecture: `docs/design/packaging-invocation-surface.md` (cli → api → core dependency direction)

### DIFF
--- a/src/pragmata/api/_error_log.py
+++ b/src/pragmata/api/_error_log.py
@@ -1,4 +1,4 @@
-"""Scoped file handler that writes ERROR+ to base_dir/annotation/errors.log."""
+"""Scoped file handler that writes ERROR+ to a caller-supplied directory."""
 
 import logging
 from collections.abc import Generator
@@ -9,11 +9,10 @@ _FMT = "%(asctime)s | %(name)s | %(levelname)s | %(message)s"
 
 
 @contextmanager
-def error_log(base_dir: Path) -> Generator[None]:
+def error_log(log_dir: Path) -> Generator[None]:
     """Attach a file handler for the duration of a block, then clean up."""
-    log_dir = Path(base_dir).expanduser().resolve() / "annotation"
     log_dir.mkdir(parents=True, exist_ok=True)
-    handler = logging.FileHandler(log_dir / "errors.log", delay=True)
+    handler = logging.FileHandler(log_dir / "errors.log", delay=True, encoding="utf-8")
     handler.setLevel(logging.ERROR)
     handler.setFormatter(logging.Formatter(_FMT))
     root = logging.getLogger("pragmata")
--- a/src/pragmata/api/_error_log.py
+++ b/src/pragmata/api/_error_log.py
@@ -11,7 +11,6 @@ _FMT = "%(asctime)s | %(name)s | %(levelname)s | %(message)s"
 @contextmanager
 def error_log(log_dir: Path) -> Generator[None]:
     """Attach a file handler for the duration of a block, then clean up."""
-    log_dir.mkdir(parents=True, exist_ok=True)
     handler = logging.FileHandler(log_dir / "errors.log", delay=True, encoding="utf-8")
     handler.setLevel(logging.ERROR)
     handler.setFormatter(logging.Formatter(_FMT))
--- a/src/pragmata/api/_error_log.py
+++ b/src/pragmata/api/_error_log.py
@@ -1,0 +1,25 @@
+"""Scoped file handler that writes ERROR+ to base_dir/annotation/errors.log."""
+
+import logging
+from collections.abc import Generator
+from contextlib import contextmanager
+from pathlib import Path
+
+_FMT = "%(asctime)s | %(name)s | %(levelname)s | %(message)s"
+
+
+@contextmanager
+def error_log(base_dir: Path) -> Generator[None]:
+    """Attach a file handler for the duration of a block, then clean up."""
+    log_dir = Path(base_dir).expanduser().resolve() / "annotation"
+    log_dir.mkdir(parents=True, exist_ok=True)
+    handler = logging.FileHandler(log_dir / "errors.log", delay=True)
+    handler.setLevel(logging.ERROR)
+    handler.setFormatter(logging.Formatter(_FMT))
+    root = logging.getLogger("pragmata")
+    root.addHandler(handler)
+    try:
+        yield
+    finally:
+        root.removeHandler(handler)
+        handler.close()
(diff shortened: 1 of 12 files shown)


src/pragmata/api/_error_log.py
saschagobel: Small design question: this helper is named generically (`error_log` / `_error_log.py`), but the implementation is annotation-specific via. Could we either make it truly generic and usable across tools (preferrable) or rename it to reflect its actual scope? Right now the name suggests reusable API logging infrastructure, but the behavior is tool-specific.

src/pragmata/api/_error_log.py
```diff
@@ -0,0 +1,25 @@
...
+
+_FMT = "%(asctime)s | %(name)s | %(levelname)s | %(message)s"
+
+
+@contextmanager
+def error_log(base_dir: Path) -> Generator[None]:
+    """Attach a file handler for the duration of a block, then clean up."""
+    log_dir = Path(base_dir).expanduser().resolve() / "annotation"
+    log_dir.mkdir(parents=True, exist_ok=True)
+    handler = logging.FileHandler(log_dir / "errors.log", delay=True)
```
saschagobel: Tiny robustness tweak: should we set `encoding="utf-8"` explicitly on the `FileHandler`?

src/pragmata/api/_error_log.py
```diff
@@ -0,0 +1,25 @@
...
+import logging
+from collections.abc import Generator
+from contextlib import contextmanager
+from pathlib import Path
+
+_FMT = "%(asctime)s | %(name)s | %(levelname)s | %(message)s"
+
+
+@contextmanager
+def error_log(base_dir: Path) -> Generator[None]:
```
saschagobel: Possible future consideration if we ever run annotation operations concurrently: this scoped file-handler approach may need revisiting for overlapping invocations. Not a concern for the current usage model though. For querygen this presents no problem since we have run-specific dirs there.

saschagobel (changes requested):
Nice addition overall, thanks! Just a design tweak revolving around `_error_log.py / error_log()` and workspace path resolution. Else only few nits.

src/pragmata/api/annotation_import.py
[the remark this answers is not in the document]
henrycgbaker: addresse core concern here: error_log() now accepts resolved log_dir: Path (instead of deriving base_dir / "annotation" internally) -> all call sites (import, setup, export) resolve paths in api layer via WorkspacePaths.from_base_dir() and pass workspace.tool_root("annotation")

Didn't intro resolve_import_paths here though (unlike export, import doesn't have run_id or write any run-specific artifacts beyond error log; adding full path bundle would mean inventing import_id just to use resolve). 

Happy to add it if/when import starts writing run-specific output / you feel is necessary now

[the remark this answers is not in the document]
henrycgbaker: @saschagobel let me know

src/pragmata/api/annotation_export.py
```diff
@@ -44,7 +48,16 @@ def export_annotations(
     )
     resolved_id = resolve_export_id(settings, export_id if isinstance(export_id, str) else None)
     workspace = WorkspacePaths.from_base_dir(settings.base_dir)
-    paths = resolve_export_paths(workspace=workspace, export_id=resolved_id).ensure_dirs()
+    annotation_paths = resolve_annotation_paths(workspace=workspace).ensure_dirs()
+    export_paths = resolve_export_paths(workspace=workspace, export_id=resolved_id).ensure_dirs()
```
saschagobel: Non-blocking: Why not add `tool_root: Path` also to `AnnotationExportPaths` so we can onyl call `export_paths = resolve_export_paths(workspace=workspace, export_id=resolved_id).ensure_dirs()` here and dispense of `annotation_paths = resolve_annotation_paths(workspace=workspace).ensure_dirs()`?

saschagobel (approved):
Looks solid now, thanks! Left just one minor non-blocking comment.

src/pragmata/core/annotation/loaders.py
```diff
@@ -147,7 +150,7 @@ def _load_hf_dataset(dataset: Dataset) -> list[dict[str, Any]]:
 
 def _load_dataframe(df: pd.DataFrame) -> list[dict[str, Any]]:
     """Convert a pandas DataFrame to list[dict]."""
-    return df.to_dict("records")
+    return cast(list[dict[str, Any]], df.to_dict("records"))
```
saschagobel: Non-blocking:
This hides a real type mismatch `(dict[Any, Any] → dict[str, Any])`, no? Suggestion: normalizing keys (str(k)) or validating instead of casting?

src/pragmata/core/paths/annotation_paths.py
```diff
@@ -8,59 +8,58 @@
...
     export_dir: Path
+    tool_root: Path
     retrieval_annotation_csv: Path
     grounding_annotation_csv: Path
     generation_annotation_csv: Path
 
     def ensure_dirs(self) -> Self:
         """Create the export directory scaffold."""
         self.export_dir.mkdir(parents=True, exist_ok=True)
+        self.tool_root.mkdir(parents=True, exist_ok=True)
```
saschagobel: Non-blocking:
Isn't that line redundant since `export_dir.mkdir(parents=True)` already creates it?

saschagobel (approved):
Looks good, just two more nits here

PR #108, opened 2026-04-02
fix(annotation): use users.list() for export user lookup

henrycgbaker:
## Summary
- `client.users()` requires a username or ID argument in Argilla v2 — calling it with no args raises `ArgillaError`
- Fixes export pipeline crash by using `client.users.list()` to enumerate all users

## References
- Discovered during UAT of export pipeline (`annotation_uat/03_export.py`)

## Test plan
- [ ] Run `annotation_uat/03_export.py` against a live Argilla instance with submitted annotations


saschagobel (approved):
All good, thank you!

PR #109, opened 2026-04-02
refactor(test): move test_annotation_paths to match source layout

henrycgbaker:
## Goal

Align test file location with source layout - `test_annotation_paths.py` was under `tests/unit/core/annotation/` but the source lives at `src/pragmata/core/paths/annotation_paths.py`.

## Scope

File move only, no code changes.


saschagobel (approved):
Perfect, thanks!

PR #110, opened 2026-04-02
feat: add deterministic deduplication for query blueprints

### LINKED ISSUE #95: Implement deterministic blueprint deduplication with semantic near-duplicate detection
**Goal**

Implement deterministic deduplication of candidate query blueprints, including exact duplicate removal and semantic near-duplicate detection using embeddings.

**Scope**

- Implement deduplication logic in:
  - `src/pragmata/core/querygen/deduplication.py`
- Deliverables:
  - `_blueprint_content_key()` helper for exact duplicate detection
  - `_load_embedding_model()` helper for loading the embedding model
  - `_embed_blueprints()` helper for batch embedding of blueprints
  - `_select_non_duplicate_indices()` helper for deterministic near-duplicate selection over a similarity matrix
  - `deduplicate_blueprints()` aggregate function orchestrating exact and near deduplication

- `_blueprint_content_key()`
  - implement as `_blueprint_content_key(candidate: QueryBlueprint) -> str`
  - constructs a deterministic serialization of all blueprint fields except `candidate_id`
  - hashes the serialized representation using `hashlib.sha256`
  - returns the hex digest as the exact-duplicate key

- `_load_embedding_model()`
  - implement as `_load_embedding_model(checkpoint: str = "all-MiniLM-L6-v2") -> SentenceTransformer`

- `_embed_blueprints()`
  - implement as `_embed_blueprints(candidates: list[QueryBlueprint]) -> NDArray[np.float32]`
  - collapses each blueprint (excluding `candidate_id`) into a deterministic string representation in fixed field order
  - loads the embedding model via `_load_embedding_model()`
  - encodes all representations in one batch using `model.encode(...)` 
  - should use normalized embeddings suitable for cosine similarity
  - returns embeddings aligned with input order

- `_select_non_duplicate_indices()`
  - implement as `_select_non_duplicate_indices(similarities, threshold: float = 0.95) -> list[int]`
  - consumes the full similarity matrix
  - applies deterministic selection:
    - process candidates from top to bottom in their original list order
    - always keep the first candidate that has not already been removed
    - after keeping candidate `i`, compare it to all later candidates `j > i`:
      - if `similarities[i, j] >= threshold`, mark candidate j as removed
      - continue until all candidates have been processed
  - returns retained indices in original order

- `deduplicate_blueprints()`
  - implement as `deduplicate_blueprints(candidates: list[QueryBlueprint]) -> list[QueryBlueprint]`
  - performs two passes:
    - Exact-duplicate pass
      - compute `_blueprint_content_key` for all candidates
      - retain first occurrence per key
      - preserve original order
    - Near-duplicate pass
      - compute embeddings once via `_embed_blueprints`
      - compute full similarity matrix once via `model.similarity(embeddings, embeddings)`
      - select retained indices via `_select_non_duplicate_indices`
  - returns order-preserving `list[QueryBlueprint]` corresponding to retained, deduplicated blueprints

- Dependency handling
  - `sentence-transformers` must be treated as an optional dependency under the `querygen` extra

**Testing requirements**

- Unit tests required covering, for example:
  - deterministic construction of `_blueprint_content_key()` and correct exact deduplication behavior
  - order preservation and first-occurrence retention in `deduplicate_blueprints()`
  - correctness of `_select_non_duplicate_indices()` 
  - correct wiring of embedding + similarity matrix into selection logic (use mocked or stub embeddings)
- Tests must not require sentence-transformers by default:
  - mock `_load_embedding_model()` or `_embed_blueprints()`
  - use deterministic fake embeddings or similarity matrices

lramir14:
## Brief summary

This PR adds deterministic deduplication for candidate query blueprints.

It includes exact duplicate removal, semantic near-duplicate filtering, unit tests for the new helpers, and a `querygen` optional dependency for `sentence-transformers`.

## Detailed description

The new deduplication logic lives in `src/pragmata/core/querygen/deduplication.py` and is implemented as a two-pass workflow.

First, exact duplicates are removed using a deterministic serialization of blueprint content, excluding `candidate_id`. This preserves the first occurrence of each unique blueprint in original order.

Second, semantic near-duplicates are removed using normalized embeddings generated from the same deterministic content representation. A full similarity matrix is computed once, and retained candidates are selected deterministically by scanning from top to bottom and removing later candidates whose similarity to an already kept candidate is above the threshold.

To keep the implementation simple and consistent with the existing querygen helper modules, the logic is split into focused helpers for content-key generation, model loading, embedding, deterministic index selection, and final orchestration.

The PR also adds unit tests covering exact deduplication, near-duplicate selection, direct embedding-helper behavior, order preservation, and empty-input handling. The embedding-related tests use mocked helpers so they do not require `sentence-transformers` by default.

Ready for review

Closes #95


src/pragmata/core/querygen/deduplication.py
```diff
@@ -0,0 +1,131 @@
...
+    serialized = _serialize_blueprint_content(candidate)
+    return hashlib.sha256(serialized.encode("utf-8")).hexdigest()
+
+
+def _select_non_duplicate_indices(
+    similarities: NDArray[np.float32],
+    threshold: float = 0.95,
+) -> list[int]:
+    """Select non-duplicate indices deterministically from a similarity matrix."""
+    matrix = np.asarray(similarities, dtype=np.float32)
```
saschagobel: Minor optional cleanup: `_select_non_duplicate_indices()` already accepts `NDArray[np.float32]`, and callers currently coerce similarity outputs to `np.float32` before passing them in. The additional `np.asarray(..., dtype=np.float32)` inside the helper is therefore a bit redundant. You could remove one coercion point to tighten the implementation slightly.

src/pragmata/core/querygen/deduplication.py
```diff
@@ -0,0 +1,131 @@
...
+            "sentence-transformers is required for blueprint deduplication. Install pragmata with the 'querygen' extra."
+        ) from exc
+
+    return SentenceTransformer(checkpoint)
+
+
+def _embed_blueprints(candidates: list[QueryBlueprint]) -> NDArray[np.float32]:
+    """Embed blueprint content in one normalized batch."""
+    if not candidates:
+        return np.empty((0, 0), dtype=np.float32)
```
saschagobel: Minor optional cleanup: `_embed_blueprints()` currently returns `np.empty((0, 0), dtype=np.float32)` for empty input. Since `deduplicate_blueprints()` already exits early on empty candidate lists, this branch is not needed on the current execution path. It would be reasonable to drop it if you want the helper to stay narrowly focused on the non-empty case.

src/pragmata/core/querygen/deduplication.py
```diff
@@ -0,0 +1,131 @@
...
+    for candidate in candidates:
+        content_key = _blueprint_content_key(candidate)
+        if content_key in seen_content_keys:
+            continue
+
+        seen_content_keys.add(content_key)
+        exact_deduplicated.append(candidate)
+
+    if len(exact_deduplicated) <= 1:
+        return exact_deduplicated
```
saschagobel: Minor optional cleanup: after the earlier non-empty guard, if `len(exact_deduplicated) <= 1` could be tightened to `== 1`, since 0 should not be reachable here. The current condition is correct, just slightly looser than necessary.

src/pragmata/core/querygen/deduplication.py
```diff
@@ -0,0 +1,131 @@
...
+        retained_indices.append(index)
+
+        for later_index in range(index + 1, matrix.shape[0]):
+            if matrix[index, later_index] >= threshold:
+                removed_indices.add(later_index)
+
+    return retained_indices
+
+
+@lru_cache(maxsize=None)
```
saschagobel: Minor optional cleanup: `_load_embedding_model()` currently uses `@lru_cache(maxsize=None)`, which is correct, but likely broader than needed for the current design. Since the workflow uses a single deduplication checkpoint by default, `maxsize=1` could communicate intent a bit more clearly while still avoiding repeated model loads within the process.

src/pragmata/core/querygen/deduplication.py
```diff
@@ -0,0 +1,131 @@
+"""Deduplication helpers for synthetic query blueprints."""
```
saschagobel: Minor wording nit: the module docstring says "Deduplication helpers for synthetic query blueprints.", but this module implements the actual blueprint deduplication logic, not just auxiliary helpers. I’d suggest using a broader description such as "Deterministic deduplication for synthetic query blueprints."

tests/unit/core/querygen/test_deduplication.py
```diff
@@ -0,0 +1,318 @@
+"""Unit tests for synthetic query blueprint deduplication helpers."""
```
saschagobel: Minor wording nit: the module docstring says "Unit tests for synthetic query blueprint deduplication helpers.", but this test module covers the actual deduplication functionality as well, not just helpers. I’d suggest a broader description such as "Unit tests for synthetic query blueprint deduplication."

tests/unit/core/querygen/test_deduplication.py
saschagobel, line 1: Consider adding a unit test for the `len(exact_deduplicated) == 1` short-circuit in `deduplicate_blueprints()`. For example, provide multiple exact duplicates and assert that the function returns the first occurrence without calling `_embed_blueprints()` or `_load_embedding_model()`. That would pin down the intended early-return behavior directly.

pyproject.toml
saschagobel, line 21: `numpy` should be added to core dependencies, not just relied on transitively via the `querygen` extra. `deduplication.py` imports numpy directly, so this is a real runtime dependency of our code.

saschagobel (changes requested):
Requesting changes for one blocking packaging issue: `deduplication.py` imports numpy directly, but numpy is not declared as a direct runtime dependency. Aside from that, I only have minor cleanup nits .

pyproject.toml
lramir14: good catch! Done.

src/pragmata/core/querygen/deduplication.py
lramir14: Agreed, changed.

lramir14:
Thanks for the review. I've addressed the requested changes and follow-up comments, including the direct numpy dependency, the deduplication cleanups, and the added short-circuit test. CI is now passing, so this should be ready for a final review.

saschagobel (approved):

PR #116, opened 2026-04-08
feat(annotation): add core IAA metric implementations and schemas

henrycgbaker:
## Goal

Add the core computation layer for Inter Annotator Agreement (IAA) metrics. NumPy implementations to keep pragmata MIT-clean (avoids GPL-licensed `krippendorff` package). This is the foundation - orchestration and API follow in a stacked PR.

## Scope

- **Core metrics** (`core/annotation/iaa.py`): Krippendorff's alpha (nominal), Cohen's kappa, bootstrap CIs, percentage agreement. All pure NumPy, no external metric libraries.
- **Report schema** (`core/schemas/iaa_report.py`): Frozen Pydantic models - `LabelAgreement`, `AnnotatorPair`, `TaskAgreement`, `IaaReport`.
- **Path bundle** (`core/paths/annotation_paths.py`): `IaaPaths` derived from `AnnotationExportPaths`, scoped under `export_dir/iaa/`.
- **Dependency**: numpy>=1.26 added as explicit `annotation` optional dependency.

## Implementation

- Alpha validated against `fast-krippendorff` library reference values
- Metrics nullable (`None` in schema) when insufficient overlap data exists
- Bootstrap supports seeded RNG for reproducibility

## Testing

- 24 unit tests for core metrics (edge cases, reference validation, bootstrap reproducibility)
- Full suite: 419 passed

## References

- Design: `docs/design/annotation-export-pipeline.md` (IAA as downstream consumer)
- Research: `.hb-scratch/research/iaa/`
- Prior research: `.hb-scratch/research/ADR0014-iaa-metric.md`


henrycgbaker:
NB: our implementation of krippendorff is not nearly as efficient as the python lib (no vectorisation, nominal-only etc), but this is tradeoff we make to stay within MIT license / shouldn't be an issue given annotation size (we are O(N * M^2 + C^2) where N is items, M is annotators, C is expected disagreement)

src/pragmata/core/annotation/iaa.py
```diff
@@ -0,0 +1,171 @@
+"""Pure-NumPy implementations of inter-annotator agreement metrics."""
+
+from __future__ import annotations
```
saschagobel: Minor: we’re on Python 3.12+, and this module doesn’t rely on postponed evaluation (no forward refs), so `from __future__ import annotations` doesn’t seem necessary here. I’d remove it for consistency.

src/pragmata/core/annotation/iaa.py
```diff
@@ -0,0 +1,171 @@
...
+            d_e += expected_counts[v] * expected_counts[w]
+    d_e = (2.0 * d_e) / (n_total * (n_total - 1))
+
+    if d_e == 0:
+        return 1.0
+
+    return 1.0 - d_o / d_e
+
+
+def cohen_kappa(labels_a: NDArray[np.integer | np.bool_], labels_b: NDArray[np.integer | np.bool_]) -> float:
```
saschagobel, line 71: The current implementation of cohen_kappa() is only correct for binary labels encoded as 1 vs not-1, but the signature/docstring reads more generally.

Two options:

- If binary-only is intended, make that explicit in the contract (docstring + type expectations).
- If this is meant to be general nominal kappa, the expected agreement (p_e) needs to be computed over all observed label classes rather than hardcoding the positive class.

src/pragmata/core/schemas/iaa_report.py
```diff
@@ -0,0 +1,58 @@
...
+class IaaReport(BaseModel):
+    """Full IAA report across all analysed tasks."""
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    export_id: str
+    created_at: datetime
+    tasks: list[TaskAgreement]
+    n_bootstrap_resamples: int
+    ci_level: float
```
saschagobel: Minor: n_bootstrap_resamples and ci_level could use constrained types to enforce validity at the schema level (e.g. positive int, and 0 < ci_level < 1) instead of relying on upstream validation.

tests/unit/core/annotation/test_iaa.py
saschagobel, line 1: Test coverage looks good overall. One thing I’d add is a cohen_kappa() test that makes the intended label contract explicit: either a non-{0,1} binary encoding / multiclass case if this is meant to be general nominal kappa, or an input-validation test if the function is intentionally binary-only. Right now the tests only cover {0,1} / bool, so they wouldn’t catch that distinction.

saschagobel (approved):
Looks solid overall, thanks — nice, clean implementation and good test coverage. Approved with a few minor comments around cohen_kappa scope/contract clarity and small consistency nits.

src/pragmata/core/annotation/iaa.py
henrycgbaker: Agree. Intent here's binary-only - iaa_runner always passes per-label 0,1 arrays, + our annotation schemas are binary by design (ADR9).

tests/unit/core/annotation/test_iaa.py
henrycgbaker: added test_binary_contract_bool_matches_int locking the bool ≡ {0,1} equivalence

henrycgbaker (commented):
implemented all suggested fixes

PR #118, opened 2026-04-08
feat(annotation): add IAA runner, API, filtering, and public re-exports

henrycgbaker:
## Goal

Wire up the IAA computation layer (from #116) into the full pipeline: CSV reading, orchestration, API entrypoint, filtering, and public facade re-exports.

## Scope

- **Runner** (`core/annotation/iaa_runner.py`): Single-pass pivot per task, computes all metrics, writes structured JSON report. `TASK_LABELS` derived from export schemas (SSOT). Reuses `TASK_CSV_ATTR`/`TASK_SCHEMA` from export_runner (made public here).
- **Filtering**: `exclude_annotators`, `after`, `before` params on both `run_iaa()` and `compute_iaa()` - enables rerunning IAA on same export with different subsets without re-exporting.
- **API** (`api/annotation_iaa.py`): `compute_iaa()` entrypoint following existing settings resolution pattern.
- **Re-exports** (`annotation/__init__.py`): `compute_iaa` and `IaaReport` added to public facade.

## Implementation

- Runner handles missing/empty CSVs gracefully (skip with warning)
- NaN metrics converted to `None` via `_or_none()` helper before Pydantic serialisation
- Pairwise kappa pairs omitted when degenerate
- Single-pass `_pivot_task()` builds all label matrices + records structure in one row iteration
- `TASK_LABELS` derived programmatically from export schema bool fields

## Testing

- 10 unit tests for runner (report structure, disk write, missing/empty CSVs, multi-task, 3-annotator pairwise, annotator exclusion, date filtering)
- Full suite: 429 passed (1 pre-existing failure unrelated)

## References

- Depends on #116 (core metrics, schemas, path bundle)
- Design: `docs/design/annotation-e2e-flow.md` (QA step)

## Out of scope

- CLI command (Dedicated phase/PR)
- Drift monitoring / windowed alpha (deferred)

### DIFF
--- a/src/pragmata/core/annotation/iaa_runner.py
+++ b/src/pragmata/core/annotation/iaa_runner.py
@@ -11,6 +11,7 @@ from pathlib import Path
 
 import numpy as np
 
+from pragmata.core.annotation.export_runner import TASK_CSV_ATTR, TASK_SCHEMA
 from pragmata.core.annotation.iaa import (
     bootstrap_alpha,
     cohen_kappa,
@@ -29,21 +30,8 @@ from pragmata.core.schemas.iaa_report import (
 logger = logging.getLogger(__name__)
 
 TASK_LABELS: dict[Task, list[str]] = {
-    Task.RETRIEVAL: ["topically_relevant", "evidence_sufficient", "misleading"],
-    Task.GROUNDING: [
-        "support_present",
-        "unsupported_claim_present",
-        "contradicted_claim_present",
-        "source_cited",
-        "fabricated_source",
-    ],
-    Task.GENERATION: ["proper_action", "response_on_topic", "helpful", "incomplete", "unsafe_content"],
-}
-
-_TASK_CSV: dict[Task, str] = {
-    Task.RETRIEVAL: "retrieval_annotation_csv",
-    Task.GROUNDING: "grounding_annotation_csv",
-    Task.GENERATION: "generation_annotation_csv",
+    task: [name for name, info in schema.model_fields.items() if info.annotation is bool]
+    for task, schema in TASK_SCHEMA.items()
 }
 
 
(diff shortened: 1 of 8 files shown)


src/pragmata/core/annotation/iaa_runner.py
```diff
@@ -0,0 +1,220 @@
+"""Orchestrate IAA computation from exported annotation CSVs."""
+
+from __future__ import annotations
```
saschagobel: Minor: we’re on Python 3.12+, and this module doesn’t rely on postponed evaluation (no forward refs), so from __future__ import annotations doesn’t seem necessary here. I’d remove it for consistency.

src/pragmata/core/annotation/iaa_runner.py
```diff
@@ -0,0 +1,220 @@
...
+
+
+def _read_csv(path: Path) -> list[dict[str, str]]:
+    """Read a CSV file into a list of row dicts."""
+    with path.open(newline="", encoding="utf-8") as f:
+        return list(csv.DictReader(f))
+
+
+def _to_bool(value: str) -> bool:
+    return value.lower() == "true"
```
saschagobel: Minor: could this reuse `core.csv_io.read_csv(csv_path, TASK_SCHEMA[task])` instead of reintroducing `_read_csv()` / `_to_bool()` locally? We already have a typed CSV boundary for Pydantic models, and that would avoid the current stringly parsing where any non-"true" value silently becomes False.

src/pragmata/core/annotation/iaa_runner.py
```diff
@@ -0,0 +1,220 @@
...
+            created = datetime.fromisoformat(row["created_at"])
+            if after and created < after:
+                continue
+            if before and created > before:
+                continue
+        filtered.append(row)
+    return filtered
+
+
+def run_iaa(
```
saschagobel, line 122: Minor architectural note: run_iaa() already reads as a full orchestration layer (I/O + control flow + writing), so compute_iaa() ends up being a thin wrapper over another orchestrator. Not blocking, but worth keeping an eye on so we don’t blur the api/core boundary further — ideally core stays focused on composable logic and API owns the top-level orchestration.

pyproject.toml
```diff
@@ -26,7 +26,7 @@ type = [
     "mypy>=1.19",
     "types-PyYAML",
 ]
-annotation = ["argilla>=2.0,<3.0",]
+annotation = ["argilla>=2.0,<3.0", "numpy>=1.26"]
```
saschagobel: Minor: numpy is already listed as a core dependency. Specifying a version constraint (>=1.26) is a good idea, though — I’d move that constraint to the main dependency instead of duplicating it here.

tests/unit/core/annotation/test_iaa_runner.py
saschagobel, line 1: Minor: it may be worth adding a test for the “degenerate pairwise kappa omitted” case, since that behavior is called out in the PR body and runner logic but doesn’t seem to be asserted here.

saschagobel (approved):
Thanks, this is in nice shape. Approved with a few minor comments/questions.

src/pragmata/core/annotation/iaa_runner.py
henrycgbaker: **re: using core.csv_io.read_csv()** - slight blocker here we might fix with a quick PR beforehand:

export CSVs aren't just schema dumps - `write_export_csv` also appends 2 cols (`constraint_violated` + `constraint_details`) that aren't in annotation schemas. Since our annotation schemas are extra="forbid", `read_csv(path, RetrievalAnnotation)` would raise on every row.

So to resue - we'd need to define `RetrievalExportRow` (etc.) extending annotation schema w/ the 2 constraint cols - I actually think this is more honest than current (mirrors actual on-disk format as csv becomes superset of annotation record -> we shouldn't pretend otherwise) +would give future consumers (QA tooling etc) a typed handle on the constraint cols.

So I propose in new fix PR (before this PR):
- intro RetrievalExportRow / GroundingExportRow / GenerationExportRow in core/schemas/annotation_export.py 
- extend each w/ `constraint_violated: bool` and `constraint_details: str`
- swap TASK_SCHEMA in export_runner.py + iaa_runner.py to these row schemas
==> then here: replace _read_csv/_to_bool with csv_io.read_csv(csv_path, TASK_EXPORT_ROW[task])

src/pragmata/core/annotation/iaa_runner.py
henrycgbaker: Not sure I've got this yet.

Was considering giving `run_iaa` full I/O + control flow + writing and moving up into api/ => keep `compute_iaa` in core/ as pure compute, but that contradicts our `run_export` pattern (orchestration runner lives in core/ and api/ only handles settings/path resolution)...

Alt split would be within core/: extract `compute_task_agreement(rows, task) -> TaskAgreement` as pure compute => `run_iaa` narrowed to read-CSVs > loop > write-report. Keeps `run_iaa` symmetric with `run_export` but removes composable logic-orchestration ?

src/pragmata/core/annotation/iaa_runner.py
saschagobel: Yes, I agree, strong refinement.

src/pragmata/core/annotation/iaa_runner.py
saschagobel: I would not change this in this PR. My comment was mainly directional: compute_iaa() and run_iaa() are both doing orchestration at different levels, so I just wanted to flag the boundary blur, not suggest a refactor right now. Given the existing run_export pattern, keeping run_iaa() as the core runner is fine; if we want to tighten the split later, I’d rather extract a smaller pure helper like compute_task_agreement(...) than move the whole runner up into api/.

And yes, this is not specific to run_iaa(). The same pattern already exists with run_export: the API handles entrypoint concerns like settings/path resolution, while the core runner handles workflow orchestration, which by our architecture should also live in the API. My point was just that we should be conscious of that boundary drift and, if we ever tighten it, do so consistently across runners rather than piecemeal.

saschagobel (approved):

PR #119, opened 2026-04-08
feat(annotation): add annotation CLI commands

henrycgbaker:
## Goal

Add Typer CLI commands for the annotation pipeline, wiring up the `api/` entrypoints.

## Scope

- `annotation setup|teardown|import|export|iaa` subcommands
- CLI is a thin execution layer: defers `from pragmata import annotation` inside each command function, passes kwargs through to the API
- Handles `None` alpha gracefully in IAA display (shows `n/a (insufficient overlap)`)

## Implementation

- `cli/commands/annotation.py` — full annotation subcommand group
- `cli/app.py` — registers `annotation_app`
- Tests mock `compute_iaa` at the facade level, testing CLI display logic only

**Review rework** (addresses [original review feedback](https://github.com/bertelsmannstift/pragmata/pull/119#pullrequestreview-4112171759)):

| Review comment | Resolution |
|---|---|
| Blocker: eager argilla import on root CLI path | `from pragmata import annotation` deferred inside each command function; module-scope imports are argilla-free |
| Client construction should move to API | Deleted `_client()` and `import argilla as rg` — CLI passes `api_url`/`api_key` as kwargs; API owns client construction (#143) |
| Import `Task` from facade, not `pragmata.core` | CLI uses `parse_tasks()` from `cli/parsing.py` which defers the facade import; test imports `Task` from `pragmata.annotation` |
| Centralise `_parse_tasks()` | Uses `parse_tasks` from `cli/parsing.py` (#148) |
| Centralise `--users` JSON parsing | Uses `parse_user_specs` from `cli/parsing.py` (#148) |

## Testing

- [x] 3 unit tests for IAA CLI display (alpha formatting, None handling, empty report)
- [x] 543/543 unit tests pass
- [x] CLI module verified argilla-free at import time

## References

- Stacked on #143 (`refactor/annotation-client-api`) — rebase onto main after #143 merges
- Depends on #144 (lazy facade, merged) and #148 (parsing helpers, merged)


src/pragmata/cli/commands/annotation.py
```diff
@@ -0,0 +1,187 @@
+"""CLI commands for annotation pipeline."""
+
+from pathlib import Path
+
+import argilla as rg
+import typer
+
+from pragmata import annotation
+from pragmata.api import UNSET
+from pragmata.core.schemas.annotation_task import Task
```
saschagobel: Minor architecture point: this command module imports Task from pragmata.core, which crosses the intended CLI → API → core dependency direction:

"- `pragmata.cli` is a thin execution layer over the API. May import `api`, but must not import `core` directly."

I’d prefer the CLI to depend only on the API/public surface here and keep core fully hidden from command modules.

src/pragmata/cli/commands/annotation.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _client(api_url: str | None, api_key: str | None) -> rg.Argilla:
+    """Build an Argilla client from explicit args or environment defaults."""
+    kwargs: dict = {}
+    if api_url is not None:
+        kwargs["api_url"] = api_url
+    if api_key is not None:
+        kwargs["api_key"] = api_key
+    return rg.Argilla(**kwargs)
```
saschagobel: I’d prefer this client construction to move into the API in a separate refactor. Right now the CLI has to know how to build an Argilla client from api_url / api_key, which makes it a bit thicker than necessary; ideally the CLI would just pass those inputs through and the API would own client initialization.

See my comments on lines R5 to R8.

src/pragmata/cli/commands/annotation.py
```diff
@@ -0,0 +1,187 @@
...
+
+
+def _parse_tasks(raw: str | None) -> list[Task] | None:
+    """Parse comma-separated task names, or None for all tasks."""
+    if raw is None:
+        return None
+    return [Task(t.strip()) for t in raw.split(",")]
```
saschagobel: Small consistency note: _parse_tasks() feels like the same class of CLI coercion/parsing helper as the logic in cli/parsing.py. Not a blocker, but I’d consider centralizing typed CLI parsing helpers in one place.

src/pragmata/cli/commands/annotation.py
```diff
@@ -0,0 +1,187 @@
...
+        " 'workspaces' and 'password' are optional.",
+    ),
+) -> None:
+    """Create Argilla workspaces, datasets, and (optionally) user accounts."""
+    import json
+
+    users = None
+    if users_json is not None:
+        raw = json.loads(Path(users_json).read_text())
+        users = [annotation.UserSpec(**u) for u in raw]
```
saschagobel: Small consistency point: the --users JSON loading/parsing feels like the same kind of CLI input normalization we already handle in cli/parsing.py. I’d consider extracting this into a parsing helper so the command stays focused on wiring arguments into the API.

src/pragmata/cli/commands/annotation.py
```diff
@@ -0,0 +1,187 @@
+"""CLI commands for annotation pipeline."""
+
+from pathlib import Path
+
+import argilla as rg
+import typer
+
+from pragmata import annotation
```
saschagobel: Blocker: annotation is an optional extra, but app.py now pulls the annotation command onto the root CLI import path. If any module on that path imports Argilla at module scope, the whole pragmata CLI can fail to import even for non-annotation commands. I think this also points to the broader design cleanup mentioned in my comment on lines R15 to R22: in a separate refactor, the annotation API should accept api_url / api_key rather than a prebuilt client, and client initialization should move behind a runtime-only core helper invoked from the API orchestration layer. That would let us remove the CLI-side Argilla client construction, keep the CLI thinner, and avoid eager Argilla imports on the base CLI path while preserving annotation as an optional extra.

saschagobel (changes requested):
Thanks — the CLI surface and tests are generally in good shape, but I’m requesting changes because the current annotation wiring appears to pull an optional Argilla dependency onto the root pragmata CLI import path, which should be addressed in a separate refactor before this can merge. Beyond that, I left a few minor architecture and consistency notes.

henrycgbaker:
Refactor PRs addressing this review are up:

- #143 — API accepts `api_url`/`api_key`, settings-layer env resolution consolidated, CLI parsing helpers centralised
- #144 — stacked on #143: lazy `pragmata.annotation` facade so CLI imports stay argilla-free

Will rebase this PR on top of both once they merge. After the rebase, the #119 diff reduces to just the annotation command module + its tests.

saschagobel (approved):
Looks all good now, thanks.

PR #121, opened 2026-04-08
ci: add workflow_dispatch trigger for manual CI runs

henrycgbaker:
## Summary

- Add `workflow_dispatch` trigger to CI workflow for manual runs against any branch
- Add local Make targets mirroring CI: `make lint`, `make type-check`, `make ci`

## Test plan

- [ ] Verify "Run workflow" button appears in Actions tab after merge
- [ ] `make lint` runs ruff format check + linter
- [ ] `make type-check` runs mypy
- [ ] `make ci` runs all three (lint + type-check + unit tests)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,7 @@ on:
     branches: [ "main" ]
   pull_request:
     branches: [ "main" ]
+  workflow_dispatch:
 
 env:
   UV_VERSION: "0.9.30"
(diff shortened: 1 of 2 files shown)


saschagobel (approved):
Nice addition — having `workflow_dispatch` and local Make targets is definitely useful.

One small note: the Make targets don’t fully mirror CI yet. In CI we also run `uv lock --check` and install deps via `uv sync --extra ... --frozen`, which ensures the environment is reproducible and aligned with the lockfile. The current Make targets only run the check commands on whatever local env happens to be active.

That’s totally fine for lightweight local use, but for `make ci` the naming suggests closer parity with CI.

PR #141, opened 2026-04-15
refactor(annotation): introduce export-row schemas for on-disk CSV format

henrycgbaker:
## Goal

Prep PR for #118. Model the export CSV on-disk row as explicit Pydantic schemas, so the IAA runner can swap its local `_read_csv` + `_to_bool` helpers (with their silent-False bug) for `csv_io.read_csv(path, TASK_EXPORT_ROW[task])`.

Addresses SG's suggestion on #118 comment thread: *"introduce RetrievalExportRow / GroundingExportRow / GenerationExportRow extending each annotation schema with `constraint_violated: bool` and `constraint_details: str`, swap `TASK_SCHEMA` in `export_runner.py` + `iaa_runner.py` to these row schemas, and replace `_read_csv`/`_to_bool` with `csv_io.read_csv(csv_path, TASK_EXPORT_ROW[task])`."* 

## Scope

1. **New export-row schemas** in `core/schemas/annotation_export.py`: `RetrievalExportRow`, `GroundingExportRow`, `GenerationExportRow`. Each extends its annotation schema with `constraint_violated: bool` (required) and `constraint_details: str = ""`. Inherits `frozen=True, extra="forbid"` via `AnnotationBase`.
2. **Split `TASK_SCHEMA`** in `core/annotation/export_runner.py` into two explicit mappings:
   - `TASK_ANNOTATION_SCHEMA`: annotation-only, used downstream for IAA label derivation (bool-field filter stays meaningful, won't pick up `constraint_violated` metadata)
   - `TASK_EXPORT_ROW`: full on-disk row, drives CSV header + construction
3. **Rewrite `write_export_csv`** to build export-row instances from `(annotation, violations)` tuples and serialise via the schema. Removes the out-of-schema column append. `extra="forbid"` on the export row now validates every write.

## Implementation

- `RetrievalExportRow(RetrievalAnnotation)` etc. Pydantic preserves parent-then-child field order in `model_fields`, so the CSV header is byte-identical to the old manual `list(schema.model_fields) + ["constraint_violated", "constraint_details"]` construction.
- `write_export_csv` now routes through `row_cls(**annotation.model_dump(), constraint_violated=bool(violations), constraint_details=";".join(violations))`. The single `model_dump(mode="json")` at the end feeds `_to_csv_value` for scalar serialisation, unchanged.
- Atomicity (`.tmp` + rename + cleanup on failure) preserved. `csv_io.write_csv` is not used directly — it's a no-op on empty input, but `write_export_csv` must always emit a header.

## On `constraint_details: str` vs `list[str]`

Modelled as `str` to preserve the existing on-disk format (`";".join(violations)`) exactly. The honest type is `list[str]`, and semicolon-joining is lossy if a violation message ever contains `;`.

BUT migrating to `list[str]` would require (a) deciding a structured serialisation format, (b) extending `csv_io` with list-field support, and (c) changing on-disk CSV bytes. Potential follow-up PR.

## Testing

- **Existing `TestWriteExportCsv` assertions unchanged and passing** — these check specific on-disk bytes (bool `"true"/"false"`, `;`-joined constraint_details, header-only on empty input, `.tmp` cleanup). Byte-identity confirmed.
- **New `test_export_runner.py::TestExportRowRoundTrip`** — writes CSVs, reads back via `csv_io.read_csv(path, TASK_EXPORT_ROW[task])`, asserts typed `bool` constraint_violated and correct values. Proves the downstream #118 swap will work end-to-end.
- **New tests in `test_annotation_export.py`** — construction, frozen, extra-forbid, required constraint_violated, parametrised field-order invariant across all three row classes.
- Full suite: 491 pass.

## References

- Downstream consumer: #118 (will be rebased onto this; the `_read_csv`/`_to_bool` swap lives there)
- SG's review thread on #118, `src/pragmata/core/annotation/iaa_runner.py` line 43

## Follow-up (flagged, not in scope)

- `constraint_details: list[str]` + `csv_io` list-field support (see above)
- Legacy `TASK_SCHEMA` alias not added; #118's rebase will replace its single import with the two new names

### DIFF
--- a/src/pragmata/core/annotation/export_runner.py
+++ b/src/pragmata/core/annotation/export_runner.py
@@ -33,6 +33,14 @@ TASK_CSV_ATTR = {
     Task.GENERATION: "generation_annotation_csv",
 }
 
+TASK_ANNOTATION_SCHEMA: dict[
+    Task, type[RetrievalAnnotation] | type[GroundingAnnotation] | type[GenerationAnnotation]
+] = {
+    Task.RETRIEVAL: RetrievalAnnotation,
+    Task.GROUNDING: GroundingAnnotation,
+    Task.GENERATION: GenerationAnnotation,
+}
+
 TASK_EXPORT_ROW: dict[Task, type[RetrievalExportRow] | type[GroundingExportRow] | type[GenerationExportRow]] = {
     Task.RETRIEVAL: RetrievalExportRow,
     Task.GROUNDING: GroundingExportRow,
--- a/src/pragmata/core/annotation/export_runner.py
+++ b/src/pragmata/core/annotation/export_runner.py
@@ -15,8 +15,11 @@ from pragmata.core.paths.annotation_paths import AnnotationExportPaths
 from pragmata.core.schemas.annotation_export import (
     AnnotationBase,
     GenerationAnnotation,
+    GenerationExportRow,
     GroundingAnnotation,
+    GroundingExportRow,
     RetrievalAnnotation,
+    RetrievalExportRow,
 )
 from pragmata.core.schemas.annotation_task import Task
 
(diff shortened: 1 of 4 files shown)


src/pragmata/core/annotation/export_runner.py
```diff
@@ -25,18 +27,26 @@
...
     Task.GROUNDING: GroundingAnnotation,
     Task.GENERATION: GenerationAnnotation,
 }
 
+TASK_EXPORT_ROW: dict[Task, type[RetrievalExportRow] | type[GroundingExportRow] | type[GenerationExportRow]] = {
+    Task.RETRIEVAL: RetrievalExportRow,
+    Task.GROUNDING: GroundingExportRow,
+    Task.GENERATION: GenerationExportRow,
+}
+
```
saschagobel, line 49: Something to keep in mind for later: these mapping annotations are now quite tightly coupled to the current fixed task set. That is fine for now, but if we ever add more task types or move toward user-defined tasks, the unions will become cumbersome because the type annotation has to be updated alongside the mapping contents. Not a change I’d ask for in this PR, just a maintainability note for future extension.

saschagobel (approved):
Nice work, thanks! I left a forward-looking comment.

PR #143, opened 2026-04-15
refactor(annotation): API accepts api_url/api_key, consolidate env-var settings resolution

henrycgbaker:
## Goal

Prerequisite for #119 (annotation CLI rebase). Moves Argilla client construction out of the CLI into a `core/` helper invoked by the API, so the API accepts `api_url`/`api_key` kwargs instead of a prebuilt client and `ARGILLA_API_KEY` comes from env.

Addresses [#119 review feedback](https://github.com/bertelsmannstift/pragmata/pull/119#pullrequestreview-4112171759) ahead of the #119 rebase.

## Scope

- Rename `resolve_provider_api_key` → `resolve_api_key` and `PROVIDER_API_KEY_ENV_VARS` → `API_KEY_ENV_VARS`; add `argilla` entry
- Add nested `ArgillaSettings(api_url)` on `AnnotationSettings` — URL only; the API key stays out of settings and is resolved from `ARGILLA_API_KEY` via `resolve_api_key("argilla")`
- Add `core/annotation/client.py::resolve_argilla_client()` with deferred `argilla` import and env fallback
- Swap `setup`, `teardown`, `import_records`, `export_annotations` to take `api_url`/`api_key` kwargs (drops `client: rg.Argilla` first arg); `compute_iaa` unchanged
- Core-layer signatures (`setup_datasets`, `provision_users`, `teardown_resources`, `fan_out_records`, `run_export`) are unchanged — the API layer now owns client construction before delegating down

**Not included** (dropped from original PR, per discussion):
- `__env_map__` / generalised env-var resolution on `ResolveSettings` — doesn't handle the nested/list shapes in `QueryGenSpec`, so dropped in favour of inline resolution
- Querygen settings migration (tied to `__env_map__`)
- Parsing helpers (already merged as #148)
- Lazy facade (already merged as #144)

Credential resolution order:
- `api_url`: kwarg > config (`settings.argilla.api_url`) > Argilla SDK default
- `api_key`: kwarg > `ARGILLA_API_KEY` env var (secrets never live in config)

## Testing

- [x] 542/542 unit tests pass
- [x] New tests for `resolve_argilla_client` (kwarg forwarding, env fallback, `MissingSecretError` on missing key)
- [x] New tests for `ArgillaSettings` nested-model (overrides, config, precedence, extra-field forbidden)
- [x] API unit tests use a shared `mock_client` fixture that patches `resolve_argilla_client` at the consumer namespace
[...]

PR #144, opened 2026-04-15
refactor(annotation): lazy facade so CLI imports stay argilla-free

henrycgbaker:
## Goal

Second of four PRs; makes the `pragmata.annotation` facade lazy so `import pragmata` stays argilla-free for users who installed without the `[annotation]` extra.

## Scope

- `pragmata/annotation/__init__.py` becomes a PEP 562 lazy facade. Re-exports resolve on attribute access; no module-scope imports of annotation-dependent types.
- `pragmata/__init__.py` gains `from . import annotation` + `"annotation"` in `__all__` so `pragmata.annotation` is reachable the same way as `pragmata.querygen` (addresses @saschagoebel's observation that `from pragmata import annotation` failed).
- Test suite: `tests/unit/annotation/test_facade_lazy.py` covers the import-time invariants (subprocess-based for clean module state).

## Implementation

PEP 562 pattern: `__all__` lists the public surface; `_LAZY` maps each name → (module, attr); `__getattr__` resolves on first access and caches into `globals()`; `TYPE_CHECKING` block provides IDE/mypy visibility.

Lazy invariant verified: `import pragmata`, `from pragmata import annotation`, and `import pragmata.annotation` all succeed without triggering `import argilla`. Accessing an annotation-dependent attribute (e.g. `annotation.SetupResult`) is the trigger point.

## Testing

- 541/541 unit tests pass
- Facade laziness verified empirically (argilla absent from `sys.modules` after facade-only imports)
- Existing annotation unit tests untouched

## References

- Closes @saschagoebel's observation re `pragmata.__init__` missing `from . import annotation`
- Part of progression: #148 (merged) → **this PR** → PR 4 (#143 reduced scope) → #119 rebase

### DIFF
--- a/tests/unit/annotation/test_facade_lazy.py
+++ b/tests/unit/annotation/test_facade_lazy.py
@@ -1,0 +1,121 @@
+"""Unit tests for the lazy pragmata.annotation facade.
+
+Some tests run a subprocess so sys.modules isolation is guaranteed — by the
+time pytest reaches these tests, other suites have already loaded
+pragmata.api.*, argilla, and most of core/annotation, so in-process sys.modules
+assertions cannot verify laziness.
+"""
+
+import importlib
+import subprocess
+import sys
+import textwrap
+
+import pytest
+
+
+@pytest.fixture(autouse=True)
+def _fresh_facade() -> None:
+    """Force a fresh facade import per test so __getattr__ runs cleanly."""
+    for name in list(sys.modules):
+        if name == "pragmata.annotation" or name.startswith("pragmata.annotation."):
+            del sys.modules[name]
+
+
+def _run_isolated(script: str) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        [sys.executable, "-c", textwrap.dedent(script)],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+
+def test_importing_facade_does_not_load_argilla_or_api_modules() -> None:
+    """Importing the facade must not eagerly pull argilla-dependent modules."""
+    result = _run_isolated(
+        """
+        import sys
+        import pragmata.annotation  # noqa: F401
+        leaks = [
+            m for m in (
+                "pragmata.api.annotation_setup",
+                "pragmata.api.annotation_import",
+                "pragmata.api.annotation_export",
+                "pragmata.api.annotation_iaa",
+                "pragmata.core.annotation.setup",
+                "pragmata.core.annotation.export_runner",
+                "argilla",
+            )
+            if m in sys.modules
+        ]
+        if leaks:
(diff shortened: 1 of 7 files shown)


tests/unit/test_facade_lazy.py
```diff
@@ -0,0 +1,113 @@
...
+
+def test_unknown_attribute_raises_attribute_error() -> None:
+    """Lookups outside __all__ surface a clean AttributeError."""
+    facade = importlib.import_module("pragmata.annotation")
+
+    with pytest.raises(AttributeError, match="pragmata.annotation"):
+        _ = facade.does_not_exist
+
+
+def test_dir_returns_public_surface() -> None:
```
saschagobel, line 112: Non-blocking: it may be worth adding a small invariant test that `set(__all__) == set(_LAZY)` so the declared public surface and the lazy resolver cannot drift apart over time.

tests/unit/annotation/test_facade_lazy.py
```diff
@@ -0,0 +1,113 @@
...
+def _run_isolated(script: str) -> subprocess.CompletedProcess:
+    return subprocess.run(
+        [sys.executable, "-c", textwrap.dedent(script)],
+        capture_output=True,
+        text=True,
+        check=False,
+    )
+
+
+def test_importing_facade_does_not_load_argilla_or_api_modules() -> None:
```
saschagobel: Non-blocking: since this PR also changes top-level reachability via pragmata.__init__, I’d add isolated import tests for import pragmata and from pragmata import annotation as well, not just import pragmata.annotation.

tests/unit/test_facade_lazy.py
```diff
@@ -0,0 +1,113 @@
...
+        import pragmata.annotation as facade
+        _ = facade.Task
+        if "argilla" in sys.modules:
+            raise SystemExit("accessing Task pulled argilla")
+        """
+    )
+    assert result.returncode == 0, result.stderr or result.stdout
+
+
+def test_accessing_setup_loads_api_module() -> None:
```
saschagobel, line 78: Non-blocking: the subprocess-based laziness tests are great; for consistency, test_accessing_setup_loads_api_module() could also be isolated, since in-process module state can make this a bit less informative depending on prior imports.

tests/unit/test_facade_lazy.py
saschagobel, line 1: Non-blocking: for a single test file, I’d be inclined to keep this flat under tests/unit/ rather than introducing tests/unit/annotation/ + __init__.py. Mirroring src/ in tests can be useful once a test area grows, but for one file this may be a bit more structure than we need right now.

saschagobel (approved):
This looks good, thanks. I left a few non-blocking test comments.

saschagobel (approved):

PR #149, opened 2026-04-16
feat: add querygen run logging

saschagobel:
**Summary**

Add lean run-level logging to the query generation API and suppress noisy dependency loggers in the CLI logging configuration.

**Key changes**

- Update `api/querygen.py`:
  - add a module logger via `logging.getLogger(__name__)`
  - wrap the staged query-generation run in a single run-scoped `error_log(...)` context
  - log high-level run lifecycle milestones for:
    - run start
    - stage 1 completion
    - stage 2 completion
    - run completion

- Update `cli/app.py`:
  - suppress noisy third-party logger output during CLI runs
  - raise `httpx` logging to `WARNING`
  - raise `sentence_transformers` logging to `WARNING`

**Status**
Ready for review.

### DIFF
--- a/src/pragmata/api/querygen.py
+++ b/src/pragmata/api/querygen.py
@@ -1,11 +1,13 @@
 """API orchestration for the synthetic query generation workflow."""
 
+import logging
 from itertools import islice
 from pathlib import Path
 from typing import Any
 
 from pydantic import BaseModel, ConfigDict, PositiveInt
 
+from pragmata.api._error_log import error_log
 from pragmata.core.paths.paths import WorkspacePaths
 from pragmata.core.paths.querygen_paths import QueryGenRunPaths, resolve_querygen_paths
 from pragmata.core.querygen.assembly import assemble_queries_meta, assemble_query_rows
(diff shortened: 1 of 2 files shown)


src/pragmata/api/querygen.py
```diff
@@ -163,73 +167,101 @@ def gen_queries(
...
-                api_key=api_key,
-                batch_candidate_ids=batch_candidate_ids,
+    logger.info(
+        "Starting query generation run %s (n_queries=%d, batch_size=%d)",
+        settings.run_id,
+        settings.n_queries,
+        settings.batch_size,
+    )
+
+    with error_log(paths.run_dir):
```
henrycgbaker, line 177: log scope is per-run, not tool-wide, cf. annotation APIs scope error_log to paths.tool_root so all runs append to one tool_root/errors.log. I think this makes more sense for querygen (runs are discrete, errors stay with their artefacts), but just flagging not symmetrical.

henrycgbaker (approved):
1 non-blocking comment

src/pragmata/api/querygen.py
saschagobel: Good point. This is intentionally not symmetrical with annotation: I used run_dir so each querygen run keeps its errors with its own artifacts. Unless we want strict consistency across tools, I’d keep the run-scoped behavior here.

PR #150, opened 2026-04-18
test: mark annotation facade laziness smoke tests as packaging

saschagobel:
**Summary**

Exclude hermetic annotation-facade laziness smoke tests from the default pytest run by marking them as packaging/import-surface tests.

**Key changes**

- update `tests/.../test_facade_lazy.py`:
  - mark subprocess-based cold-state laziness checks with `@pytest.mark.packaging`
  - leave cheap in-process facade behavior tests in the default unit test suite
- update `pyproject.toml`:
  - add a dedicated `packaging` pytest marker
  - exclude `packaging` tests from the default run

**Why**

These tests validate packaging/import-surface behavior in a fresh interpreter and are not ordinary unit tests. Marking them separately keeps the default test run fast while preserving the hermetic smoke tests for explicit execution.

**Status**

Ready for review.


henrycgbaker (approved):
Approved - non-blocking question(s)

Would it make sense to move this out of unit, but then to add to a blocking CI step `uv run pytest -m packaging -q`. This allows quick default runs locally (= dev inner loop) while still running the packaging tests in ci as a separate step. When i ran these tests at head of this pr branch I got:
- default (pkg tests excluded) ~5s
- just pkg tests was ~10s
= combined ~15s

If that's too slow for PR CI, perhaps a scheduled workflow on main (fails loudly in actions tab)?

saschagobel:
> Approved - non-blocking question(s)
> 
> Would it make sense to move this out of unit, but then to add to a blocking CI step `uv run pytest -m packaging -q`. This allows quick default runs locally (= dev inner loop) while still running the packaging tests in ci as a separate step. When i ran these tests at head of this pr branch I got:
> 
> * default (pkg tests excluded) ~5s
> * just pkg tests was ~10s
>   = combined ~15s
> 
> If that's too slow for PR CI, perhaps a scheduled workflow on main (fails loudly in actions tab)?

Thanks — I went ahead and applied that split.

I moved the lazy facade test out of `unit`, added a separate `packaging` marker to the package/import-surface tests, and wired a dedicated CI step to run `uv run pytest -m packaging -q`. So the default local run stays fast, while the packaging/import-surface checks still run as an explicit blocking CI lane.

PR #151, opened 2026-04-19
feat(annotation): discard contract (enum, schema, fetcher, hidden questions)

henrycgbaker:
## Goal
Data-layer foundation for annotator-driven record discard. Captures a structured reason + optional notes on Argilla's native discarded response status, so the data flows through to export.

## Scope
Stacked on #120 (\`feature/dataset-id\`). **Backend only** — no user-visible UI changes (that's #152, stacked on this).

## Implementation
- **\`DiscardReason\` StrEnum** in \`core/schemas/annotation_task\`: \`low_quality_query\`, \`duplicate\`, \`unclear\`, \`beyond_domain_knowledge\`. Re-exported from the public annotation facade.
- **Hidden questions on all three tasks** (retrieval, grounding, generation): \`discard_reason\` (\`LabelQuestion\` over \`DiscardReason\`) and \`discard_notes\` (\`TextQuestion\`), both \`required=False\`. They carry discard metadata on the response without affecting submitted-response flow.
- **Export schema** (\`AnnotationBase\`) gains \`response_status\`, \`discard_reason\`, \`discard_notes\`. Query filter widens from submitted-only to \`[\"submitted\", \"discarded\"]\`. Per-user response grouping preserves \`response_status\`. Discarded rows have null task-label columns and skip constraint validation.

## Testing
- Unit: \`test_argilla_task_definitions\`, \`test_export_fetcher\`, \`test_annotation_export\` — 80 tests green
- Manual: not needed for this PR — stacked UI PR #152 does the end-to-end test

## References
- \`docs/design/annotation-export-pipeline.md\` — updated shared metadata columns
- ADR-0009 (schemas hardcoded), ADR-0010 (one dataset per task linked by \`record_uuid\`)

### DIFF
--- a/docs/design/annotation-export-pipeline.md
+++ b/docs/design/annotation-export-pipeline.md
@@ -73,7 +73,7 @@ Export schemas define what downstream pipelines need. They are not a mirror of w
 | `record_status` | string | Argilla record status: `pending` or `completed` (whether the record met its `TaskDistribution` overlap target) |
 | `response_status` | string | Annotator response status: `submitted` or `discarded` |
 | `discard_reason` | string\|null | Discard reason selected by annotator; null for submitted responses |
-| `discard_notes` | string | Optional free-text notes from the discard panel; empty string if absent |
+| `discard_notes` | string\|null | Optional free-text notes from the discard panel; null if absent |
 
 **Not exported:**
 - `_server_id` — Argilla-internal UUID; `record_uuid` already serves as the cross-dataset identifier
--- a/docs/design/annotation-export-pipeline.md
+++ b/docs/design/annotation-export-pipeline.md
@@ -37,7 +37,7 @@ Argilla PostgreSQL
 
 ## Inputs
 
-Three Argilla datasets, accessed via Argilla SDK. Filter: `status == "submitted"` only (exclude draft, discarded). 
+Three Argilla datasets, accessed via Argilla SDK. Filter: `status in ["submitted", "discarded"]` — both submitted and annotator-discarded responses are exported; draft records are excluded.
 
 
 | Dataset | Records |
@@ -71,10 +71,14 @@ Export schemas define what downstream pipelines need. They are not a mirror of w
 | `inserted_at` | datetime | When the record was loaded into Argilla, tracks batch provenance |
 | `created_at` | datetime | Response submission timestamp |
 | `record_status` | string | Argilla record status: `pending` or `completed` (whether the record met its `TaskDistribution` overlap target) |
+| `response_status` | string | Annotator response status: `submitted` or `discarded` |
+| `discard_reason` | string\|null | Discard reason selected by annotator; null for submitted responses |
+| `discard_notes` | string | Optional free-text notes from the discard panel; empty string if absent |
 
 **Not exported:**
-- `response_status` — we filter to `submitted` only; column would be constant
 - `_server_id` — Argilla-internal UUID; `record_uuid` already serves as the cross-dataset identifier
+
+**Discarded rows:** label columns are null; constraint validation is skipped.
 
 ### Task 1: Retrieval — `retrieval.csv`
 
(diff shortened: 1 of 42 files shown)


docs/design/annotation-export-pipeline.md
```diff
@@ -71,11 +71,15 @@ Export schemas define what downstream pipelines need. They are not a mirror of w
 | `inserted_at` | datetime | When the record was loaded into Argilla, tracks batch provenance |
 | `created_at` | datetime | Response submission timestamp |
 | `record_status` | string | Argilla record status: `pending` or `completed` (whether the record met its `TaskDistribution` overlap target) |
+| `response_status` | string | Annotator response status: `submitted` or `discarded` |
```
saschagobel, line 74: Small schema question: we model `discard_reason` as `string | null`, but `discard_notes` as string with empty string as the “absent” case. Is there a reason for this asymmetry? Would it be cleaner to use `string | null` for both fields and treat absence uniformly as null?

docs/design/annotation-export-pipeline.md
```diff
@@ -37,7 +37,7 @@ Argilla PostgreSQL
 
 ## Inputs
 
-Three Argilla datasets, accessed via Argilla SDK. Filter: `status == "submitted"` only (exclude draft, discarded). 
+Three Argilla datasets, accessed via Argilla SDK. Filter: `status in ["submitted", "discarded"]` — both submitted and annotator-discarded responses are exported; draft records are excluded.
```
saschagobel, line 40: Design question on export behavior: we now include both submitted and discarded responses by default.

For downstream evaluation/training pipelines, we’d typically want only submitted rows — including discarded ones by default could lead to accidental misuse.

Should we consider adding a flag (follow-up PR) to control this (e.g., include_discarded: bool, default False) so users can opt in explicitly?

tests/unit/core/annotation/test_argilla_task_definitions.py
```diff
@@ -70,6 +70,17 @@ def test_notes_question(self):
...
+        q = _get_question(_RETRIEVAL, "discard_reason")
+        assert isinstance(q, rg.LabelQuestion)
+        assert q.required is False
+        assert set(q.labels) == {r.value for r in DiscardReason}
+
+    def test_discard_notes_question(self):
+        q = _get_question(_RETRIEVAL, "discard_notes")
+        assert isinstance(q, rg.TextQuestion)
+        assert q.required is False
+
```
saschagobel: Minor test-structure question: we now duplicate the same `test_discard_reason_question / test_discard_notes_question` logic across retrieval, grounding, and generation.

Would it make sense to parametrize this over tasks instead, to avoid repetition and keep the contract centralized in one place?

tests/unit/core/annotation/test_iaa_runner.py
```diff
@@ -87,6 +88,26 @@ def _make_row(
     return (annotation, [])
 
 
+_RETRIEVAL_LABELS = ("topically_relevant", "evidence_sufficient", "misleading")
+
+
+def _make_discarded_retrieval_row(record_uuid: str, annotator_id: str) -> tuple:
+    schema_cls = TASK_EXPORT_ROW[Task.RETRIEVAL].__bases__[0]
```
saschagobel: Small test robustness question: `_make_discarded_retrieval_row()` derives the schema via `TASK_EXPORT_ROW[Task.RETRIEVAL].__bases__[0]`.

Is there a reason to depend on the inheritance structure here? This feels a bit brittle — using the concrete export row class directly (or the mapping itself) might be more stable against refactors.

src/pragmata/annotation/__init__.py
```diff
@@ -14,6 +14,7 @@
 from typing import TYPE_CHECKING
 
 __all__ = [
+    "DiscardReason",
```
saschagobel: Quick API-surface question: I see we now re-export `DiscardReason` from `pragmata.annotation` and cover it in the facade lazy-loading test. What is the intended user-facing use case for exposing this enum publicly, rather than keeping it internal for Argilla/task-definition/export wiring?

saschagobel (approved):
Approved — looks solid, thanks. I left a few minor questions around public API exposure (DiscardReason) and some small consistency/robustness points, but nothing blocking.

PR #152, opened 2026-04-19
feat(annotation): discard panel custom field and DOM integration

henrycgbaker:
## Goal
User-facing discard UX: a collapsible red-bordered panel with reason dropdown + notes, wired up via CustomField iframe to Argilla's native discard — so a structured reason is always captured.

## Scope
Stacked on #151 (\`feature/reject-flow\`) which supplies the backend contract (\`DiscardReason\`, hidden questions, export schema). This PR adds the CustomField UX and its DOM integration.

## Implementation
**Custom field template** (\`discard_flow.html\`) — srcdoc iframe, same-origin with the parent page. Renders a \`<details>\` panel (red border, distinct styling) containing:
- reason dropdown over \`DiscardReason\` enum values
- optional notes textarea
- \"Discard\" button (disabled until a reason is picked)

**Task definitions** — each of the three tasks gains a \`rg.CustomField(name=\"discard_flow\", ..., required=False)\`. Records populate \`\"discard_flow\": {\"text\": \"\"}\` so the iframe actually renders (Argilla skips fields without a record value).

**DOM integration** (from inside the iframe, via \`window.parent.document\`):
- On click: look up \`.title-area\` elements matching \`Discard reason\` / \`Discard notes (optional)\`, walk to \`.closest('.wrapper')\`, click the matching label chip, fill the textarea, then trigger \`.button--discard\`.
- Inject parent CSS hiding \`.button--discard\` so the panel is the only discard path.
- Find \`.title-area\` elements matching our hidden questions and hide their \`.closest('.wrapper')\` — keeps the annotator's view clean. Polls every 300ms to re-apply after Argilla re-renders on record navigation.
- Hide Argilla's rendered field title (\"Discard this record\") by hiding all iframe siblings inside \`.custom_field_component\`.

**iframe sizing bug fix** (also affects \`collapsible_field.html\`): \`document.documentElement.scrollHeight\` is clamped to at least the iframe viewport, so once the iframe grows it can't shrink back on collapse. Fix: reset \`frameElement.style.height\` to \`0\` first, measure in next frame, then set. Makes the context-set / query grey box hug its summary when closed.

## Testing
- Unit: \`test_argilla_task_definitions\` (CustomField presence), \`test_import\` (discard_flow field populated) — 107 tests green overall
- Manual: verified on retrieval / grounding / generation datasets — panel collapses cleanly, reason+notes submit propagates to the discarded response, native discard button hidden, underlying question cards hidden

## References
- \`docs/design/annotation-interface.md\` — Record discard UX section
- \`.planning/phase-6-reject-flow/plan.md\` (local) — original design with CustomField Option C

## Residual fragility
Selectors \`.wrapper\`, \`.title-area\`, \`.button--discard\`, \`.custom_field_component\` are Argilla-internal — re-verify on Argilla upgrade (~5 min work). Structural ~60px padding inside \`.custom_field_component\` is unavoidable without patching Argilla itself.

### DIFF
--- a/src/pragmata/core/annotation/argilla_task_definitions.py
+++ b/src/pragmata/core/annotation/argilla_task_definitions.py
@@ -16,7 +16,7 @@ from string import Template
 
 import argilla as rg
 
-from pragmata.core.schemas.annotation_task import Task
+from pragmata.core.schemas.annotation_task import DiscardReason, Task
 
 DATASET_NAMES: dict[Task, str] = {
     Task.RETRIEVAL: "retrieval",
@@ -34,6 +34,18 @@ def _collapsible_field(name: str, title: str, template_text: str) -> rg.CustomFi
         advanced_mode=True,
         required=True,
     )
+
+
+def _discard_questions() -> list[rg.LabelQuestion | rg.TextQuestion]:
+    return [
+        rg.LabelQuestion(
+            name="discard_reason",
+            title="Discard reason",
+            labels=[r.value for r in DiscardReason],
+            required=False,
+        ),
+        rg.TextQuestion(name="discard_notes", title="Discard notes (optional)", required=False),
+    ]
 
 
 @functools.cache
@@ -72,6 +84,7 @@ def build_task_settings() -> dict[Task, rg.Settings]:
                     required=True,
                 ),
                 rg.TextQuestion(name="notes", title="Notes (optional)", required=False),
+                *_discard_questions(),
             ],
             metadata=[
                 rg.TermsMetadataProperty("record_uuid", visible_for_annotators=False),
@@ -120,6 +133,7 @@ def build_task_settings() -> dict[Task, rg.Settings]:
                     required=True,
                 ),
                 rg.TextQuestion(name="notes", title="Notes (optional)", required=False),
+                *_discard_questions(),
             ],
             metadata=[
                 rg.TermsMetadataProperty("record_uuid", visible_for_annotators=False),
@@ -165,6 +179,7 @@ def build_task_settings() -> dict[Task, rg.Settings]:
                     required=True,
                 ),
                 rg.TextQuestion(name="notes", title="Notes (optional)", required=False),
+                *_discard_questions(),
             ],
             metadata=[
                 rg.TermsMetadataProperty("record_uuid", visible_for_annotators=False),
--- a/src/pragmata/core/annotation/argilla_task_definitions.py
+++ b/src/pragmata/core/annotation/argilla_task_definitions.py
@@ -56,6 +56,14 @@ def build_task_settings() -> dict[Task, rg.Settings]:
     (or with a mock client in tests). Cached after first call.
     """
     template_text = files("pragmata.core.annotation").joinpath("collapsible_field.html").read_text(encoding="utf-8")
+    discard_template = files("pragmata.core.annotation").joinpath("discard_flow.html").read_text(encoding="utf-8")
+    discard_field = rg.CustomField(
+        name="discard_flow",
+        title="Discard this record",
+        template=discard_template,
+        advanced_mode=True,
+        required=False,
+    )
 
     return {
         Task.RETRIEVAL: rg.Settings(
@@ -63,6 +71,7 @@ def build_task_settings() -> dict[Task, rg.Settings]:
                 rg.TextField(name="query", title="Query", required=True),
                 rg.TextField(name="chunk", title="Chunk", required=True),
                 _collapsible_field("generated_answer", "Generated answer", template_text),
+                discard_field,
             ],
             questions=[
                 rg.LabelQuestion(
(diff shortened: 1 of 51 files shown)


src/pragmata/core/annotation/discard_flow.html
```diff
@@ -0,0 +1,193 @@
...
+        ta.focus();
+        ta.value = notes;
+        ta.dispatchEvent(new Event('input', { bubbles: true }));
+        ta.dispatchEvent(new Event('change', { bubbles: true }));
+      }
+    }
+
+    setTimeout(function() {
+      hideStaleQuestions();
+      var discardBtn = pdoc.querySelector(NATIVE_DISCARD_SEL);
```
saschagobel: Have you verified that Argilla will only ever expose a single relevant `.button--discard` in the DOM for the active record? Right now this uses a global `pdoc.querySelector(...)`, so I’m a bit unsure how robust this is if multiple discard buttons exist transiently during re-renders.

Would it be feasible to scope the lookup to the same card/container as the hidden discard fields? For example:

```
var reasonCard = cardFor(REASON_TITLE);

// Fallback to document if we can’t localize (keeps current behavior)
var scope = reasonCard || pdoc;

var discardBtn = scope.querySelector(NATIVE_DISCARD_SEL);
if (discardBtn) discardBtn.click();
```

src/pragmata/core/annotation/discard_flow.html
```diff
@@ -0,0 +1,193 @@
...
+
+  <label>Reason:
+    <select id="reason">
+      <option value="">&#8212; select &#8212;</option>
+      <option value="low_quality_query">Low quality / unrealistic query</option>
+      <option value="duplicate">Duplicate</option>
+      <option value="unclear">Unclear query or answer</option>
+      <option value="beyond_domain_knowledge">Beyond my domain knowledge</option>
+    </select>
+  </label>
```
saschagobel, line 49: The current discard reasons feel a bit ambiguous and potentially inconsistent in how annotators would apply them. In particular, I’m not sure the distinction between `low_quality_query` and `unclear` is sharp enough — “low quality / unrealistic” could easily include “unclear,” which might lead to inconsistent labeling. Would it make sense to either collapse these into one reason or separate them more explicitly?

I’m also a bit unsure about including `duplicate` here. Unless there’s a very clear operational definition and annotators have enough context to reliably detect duplicates, this might introduce more noise than signal. Should we consider removing it?

Finally, the wording of `beyond_domain_knowledge` feels slightly awkward. Maybe something more neutral and task-centered like “Outside my domain expertise” or “Requires domain expertise I don’t have” would read more naturally for annotators.

Concretely, something along these lines might be clearer:

```html
<option value="invalid_or_unrealistic">Invalid or unrealistic record</option>
<option value="not_evaluable">Cannot be reliably evaluated</option>
<option value="outside_reviewer_expertise">Outside my domain expertise</option>
```

src/pragmata/core/annotation/argilla_task_definitions.py
```diff
@@ -56,13 +56,22 @@ def build_task_settings() -> dict[Task, rg.Settings]:
...
     """
     template_text = files("pragmata.core.annotation").joinpath("collapsible_field.html").read_text(encoding="utf-8")
+    discard_template = files("pragmata.core.annotation").joinpath("discard_flow.html").read_text(encoding="utf-8")
+    discard_field = rg.CustomField(
+        name="discard_flow",
+        title="Discard this record",
+        template=discard_template,
+        advanced_mode=True,
+        required=False,
+    )
```
saschagobel: We’re reusing the same `rg.CustomField` instance (`discard_field`) across multiple `rg.Settings`. Do we know if Argilla treats field definitions as purely declarative, or could it mutate/bind them internally per task? If it’s the latter, it might be safer to instantiate a fresh field per task to avoid any unintended cross-task coupling.

saschagobel (approved):
Thanks for putting this together. Approving with a few non-blocking questions on DOM robustness, discard taxonomy clarity, and safe field instantiation.

src/pragmata/core/annotation/discard_flow.html
```diff
@@ -0,0 +1,197 @@
...
+  <p style="margin-top:8px;font-size:13px;color:#555">
+    Use this if the record is unsuitable for annotation
+    (e.g. invalid, unclear, or outside your expertise).
+  </p>
+
+  <label>Reason:
+    <select id="reason">
+      <option value="">&#8212; select &#8212;</option>
+      <option value="invalid_or_unrealistic">Invalid or unrealistic record</option>
+      <option value="unclear">Unclear query or answer</option>
```
saschagobel: query-answer pair does not always apply. Consider adjusting to:

"<option value="unclear">Unclear query/context/answer relationship</option>"

saschagobel (approved):

PR #157, opened 2026-04-21
feat: expose near-duplicate tolerance in API

saschagobel:
**Summary**

Expose `near_duplicate_tolerance` as a user-facing querygen knob in the library/API path so callers can control how strictly semantic near-duplicates are removed during stage-1 blueprint selection.

**Key changes**

- Update `core/querygen/deduplication.py`:
  - replace the hard-coded semantic near-duplicate cutoff with an explicit `near_duplicate_tolerance` parameter
  - require and thread that value through the semantic deduplication path

- Update `core/settings/querygen_settings.py`:
  - add `near_duplicate_tolerance` to `QueryGenRunSettings`

- Update `api/querygen.py`:
  - expose `near_duplicate_tolerance` on `gen_queries()`
  - thread it into `QueryGenRunSettings.resolve(...)`
  - pass the resolved value into `deduplicate_blueprints(...)` during stage-1 blueprint selection

- Add/update unit tests covering:
  - explicit `near_duplicate_tolerance` handling in deduplication tests
  - `near_duplicate_tolerance` in querygen run settings resolution
  - API wiring from `gen_queries()` through resolved settings into blueprint deduplication

**Out of scope**

- CLI wiring in `src/pragmata/cli/commands/querygen.py`
- CLI docs/examples
- CLI parsing/forwarding tests

These are intentionally deferred to a follow-up PR that brings the querygen CLI surface back into sync with the API/settings surface in one pass.

**Status**

Draft

### DIFF
--- a/src/pragmata/api/querygen.py
+++ b/src/pragmata/api/querygen.py
@@ -104,9 +104,10 @@ def gen_queries(
             values use fewer, bigger calls; smaller values split
             generation into more repeated calls. Defaults to 25.
         near_duplicate_tolerance: Similarity tolerance used for semantic
-            near-duplicate blueprint removal. Lower values deduplicate more
-            aggressively; higher values allow more similar blueprints, including
-            close paraphrases, to remain. Defaults to 0.95.
+            near-duplicate blueprint removal. Must be in the range (0, 1],
+            where lower values deduplicate more aggressively and higher values
+            allow more similar blueprints, including close paraphrases, to remain.
+            Defaults to 0.95.
         enable_planning_memory: Whether to enable planning memory for the run.
             Defaults to True. When enabled, an additional LLM updates and persists a
             compact summary of prior blueprint generation across batches and compatible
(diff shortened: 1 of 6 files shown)


src/pragmata/api/querygen.py
```diff
@@ -102,6 +103,10 @@ def gen_queries(
         batch_size: Number of queries to generate per LLM call. Larger
             values use fewer, bigger calls; smaller values split
             generation into more repeated calls. Defaults to 25.
+        near_duplicate_tolerance: Similarity tolerance used for semantic
```
lramir14, line 106: In this docstring, I am thinking if it would be beneficial to establish the range of the function (0,1].

lramir14 (approved):
Thanks for the work. It looks good to me. The new near_duplicate_tolerance setting is wired cleanly through the current querygen flow. Only a small clarity note: the valid range is enforced in settings, but the docstrings could state it more explicitly as (0, 1]. Approved.

PR #159, opened 2026-04-22
feat(annotation): add --include-discarded flag to export

henrycgbaker:
## Goal

Address PR #151 review comment: avoid accidentally polluting downstream evaluation/training pipelines with annotator-discarded responses by making them opt-in.

## Scope

- Default `pragmata annotation export` to fetching submitted responses only
- Add opt-in flag (`--include-discarded` CLI / `include_discarded=True` API) for analyses that need discard data (e.g. studying rejection reasons)
- CSV schema unchanged: `discard_reason` and `discard_notes` columns still present (always-null when flag is off)

## Implementation

Threads a single bool through the existing layers, applied at the Argilla query filter:

```
cli --include-discarded
   v
api.export_annotations(include_discarded=False)
   v
core.run_export(..., include_discarded=...)
   v
core.fetch_task(..., include_discarded=...)
   v
rg.Filter([("response.status", "in", ["submitted"] | ["submitted", "discarded"])])
```

- `core/annotation/export_fetcher.py`: `fetch_task` builds the status list conditionally
- `core/annotation/export_runner.py`: `run_export` threads the kwarg through
- `api/annotation_export.py`: public kwarg, defaults to `False`, docstring updated
- `cli/commands/annotation.py`: `--include-discarded` Typer option, passes through
- `docs/design/annotation-export-pipeline.md`: documents default behaviour and opt-in

Filter applied at fetch level (vs post-fetch) so discarded responses never enter the pipeline when not requested - cheaper over the wire and aligns with the `Inputs` section of the design doc.

## Testing

- 2 new tests in `test_export_fetcher.py` asserting `query.filter.conditions` contents for both flag values
- Updated 3 existing discard tests to pass `include_discarded=True` explicitly (matches real-world callsite)
- 261 annotation/api/facade tests pass; 21 CLI tests pass

## References

- Design doc: `docs/design/annotation-export-pipeline.md`
- Follows discard contract from PR #151
- Per-invocation flag (not deployment config) - lives as API kwarg + CLI flag, not in `AnnotationSettings`

### DIFF
--- a/src/pragmata/api/annotation_export.py
+++ b/src/pragmata/api/annotation_export.py
@@ -25,11 +25,14 @@ def export_annotations(
     tasks: list[Task] | None = None,
     dataset_id: str | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
+    include_discarded: bool = False,
 ) -> ExportResult:
-    """Fetch submitted annotations from Argilla and write flat CSVs per task.
+    """Fetch annotations from Argilla and write flat CSVs per task.
 
-    Queries each task dataset for submitted-only responses, groups by annotator,
-    applies constraint validation, and writes atomic CSVs.
+    By default, queries each task dataset for submitted-only responses. Set
+    ``include_discarded=True`` to also include responses the annotator
+    discarded; their label columns are null and constraint validation is
+    skipped, but the row carries ``discard_reason`` and ``discard_notes``.
 
     Credential resolution:
     - ``api_url``: kwarg > ``ARGILLA_API_URL`` env > config (``argilla.api_url``)
@@ -44,6 +47,9 @@ def export_annotations(
         tasks: Tasks to export. Defaults to all three tasks.
         dataset_id: Suffix identifying which datasets to export from.
         config_path: Path to YAML config file for settings resolution.
+        include_discarded: If True, include discarded responses alongside
+            submitted ones. Defaults to False to avoid polluting downstream
+            evaluation pipelines.
 
     Returns:
         ExportResult with file paths, row counts, and constraint summary.
@@ -65,7 +71,7 @@ def export_annotations(
     resolved_tasks = tasks if tasks is not None else list(Task)
 
     with error_log(export_paths.tool_root):
-        result = run_export(client, settings, export_paths, resolved_tasks)
+        result = run_export(client, settings, export_paths, resolved_tasks, include_discarded=include_discarded)
 
     logger.info(
         "Export complete: %d task(s), %d total rows",
--- a/src/pragmata/api/annotation_export.py
+++ b/src/pragmata/api/annotation_export.py
@@ -25,7 +25,7 @@ def export_annotations(
     tasks: list[Task] | None = None,
     dataset_id: str | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
-    include_discarded: bool = False,
+    include_discarded: bool | Unset = UNSET,
 ) -> ExportResult:
     """Fetch annotations from Argilla and write flat CSVs per task.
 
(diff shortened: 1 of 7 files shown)


src/pragmata/api/annotation_export.py
```diff
@@ -25,11 +25,14 @@ def export_annotations(
     tasks: list[Task] | None = None,
     dataset_id: str | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
+    include_discarded: bool = False,
```
saschagobel: I’d prefer not to put the False default directly in the API signature here. For consistency with the rest of the settings-resolution pattern, this should be `include_discarded: bool | Unset = UNSET`, with the default living in `AnnotationSettings` and then passed through after resolution. That keeps API kwargs as override inputs rather than the source of runtime defaults.

src/pragmata/cli/commands/annotation.py
```diff
@@ -135,6 +140,7 @@ def export_command(
         tasks=parse_tasks(tasks),
         dataset_id=UNSET if dataset_id is None else dataset_id,
         config_path=UNSET if config is None else config,
+        include_discarded=include_discarded,
```
saschagobel: Same default-resolution point as above. Since this flag is absent unless the user opts in, I’d pass `include_discarded=UNSET if not include_discarded else True` here rather than always passing False. Otherwise the CLI’s default value becomes an explicit override and would prevent AnnotationSettings / config defaults from taking effect later.

src/pragmata/core/annotation/export_fetcher.py
```diff
@@ -97,8 +97,14 @@ def fetch_task(
     settings: AnnotationSettings,
     task: Task,
     user_lookup: dict[UUID, str],
+    *,
+    include_discarded: bool = False,
```
saschagobel: Same default-source concern one layer deeper: I’d avoid giving `fetch_task()` its own False default. Once settings are resolved, this should receive the effective value explicitly, e.g. `include_discarded: bool`, so `AnnotationSettings` remains the only place defining the runtime default.

src/pragmata/core/annotation/export_runner.py
```diff
@@ -110,6 +110,8 @@ def run_export(
     settings: "AnnotationSettings",
     paths: AnnotationExportPaths,
     tasks: list[Task],
+    *,
+    include_discarded: bool = False,
```
saschagobel: Same default-source issue here too: I’d avoid repeating `include_discarded: bool = False` at each internal layer.

tests/unit/core/annotation/test_export_fetcher.py
```diff
@@ -298,7 +298,7 @@ def test_discarded_response_included(self) -> None:
         ]
         record = _make_record(fields=_RETRIEVAL_FIELDS, metadata=_BASE_METADATA, responses=responses)
         client = _mock_client_with_records([record])
-        rows = fetch_task(client, _SETTINGS, Task.RETRIEVAL, {_UID1: "alice"})
+        rows = fetch_task(client, _SETTINGS, Task.RETRIEVAL, {_UID1: "alice"}, include_discarded=True)
```
saschagobel, line 301: Since `fetch_task()` should receive the resolved `include_discarded` value explicitly rather than own the default, this test (and the following) should not exercise an omitted argument or call it the default behavior.

saschagobel (changes requested):
Thanks for addressing the export behavior. Changes requested because `include_discarded=False` is currently duplicated across API/core layers instead of being resolved once through `AnnotationSettings` and passed explicitly through the internal call chain.

saschagobel (approved):

PR #160, opened 2026-04-22
feat(annotation): track n_annotators per task in ExportResult

henrycgbaker:
## Goal

Extend `ExportResult` with per-task annotator counts. Foundational for export-manifest sidecar (planned next), which will record this as provenance metadata alongside the CSVs.

## Scope

- Add `n_annotators: dict[Task, int]` field to `ExportResult`
- Populate it in `run_export` by aggregating distinct `annotator_id` values from already-fetched rows
- No behaviour change: CSVs, constraint checks, Argilla fetch are unchanged

## Implementation

One line in `run_export`:

```python
n_annotators = {task: len({row.annotator_id for row, _ in task_rows[task]}) for task in tasks}
```

Zero cost: `annotator_id` is already populated on every annotation row in the existing fetcher, so we just reshape. The early-return path for `tasks=[]` returns an empty dict to keep the shape consistent.

Chose `dict[Task, int]` (matching the shape of `row_counts` and `files`) rather than a flat total so per-task analysis stays possible downstream.

## Testing

- Updated `TestExportResult` to construct with the new field
- Added `TestNAnnotators`:
  - `test_counts_distinct_annotators_per_task` - two annotators on one record -> `{RETRIEVAL: 2}`
  - `test_empty_task_reports_zero` - no records -> `{RETRIEVAL: 0}`
  - `test_no_tasks_returns_empty_dict` - `tasks=[]` -> `{}`
- All 262 annotation + api + facade tests pass; 21 CLI tests pass

## References

- Prep for follow-up PR: export manifest JSON sidecar (mirrors querygen's `synthetic_queries.meta.json`)
- Related: #159 (`include_discarded` flag, also destined for the manifest)

### DIFF
--- a/docs/design/annotation-export-pipeline.md
+++ b/docs/design/annotation-export-pipeline.md
@@ -37,7 +37,7 @@ Argilla PostgreSQL
 
 ## Inputs
 
-Three Argilla datasets, accessed via Argilla SDK. Filter: `status in ["submitted", "discarded"]` — both submitted and annotator-discarded responses are exported; draft records are excluded.
+Three Argilla datasets, accessed via Argilla SDK. Filter: `response.status == "submitted"` by default; pass `--include-discarded` (CLI) or `include_discarded=True` (API) to also include `discarded` responses. Draft responses are always excluded. Discarded responses are off by default to keep downstream evaluation/training pipelines clean — opt in when you specifically want to analyse discard reasons.
 
 
 | Dataset | Records |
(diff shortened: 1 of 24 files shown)


saschagobel (approved):
Thanks, this looks good to me. The implementation reuses the already-fetched rows, preserves the existing export behavior, and the tests cover the relevant result shapes, including empty task and no-task cases.

PR #161, opened 2026-04-22
feat(annotation): write export provenance sidecar JSON

henrycgbaker:
> **Stacked PR** - this branch stacks on #160, which stacks on #159. The PR diff on GitHub only shows this PR's own changes once the base is set to `feature/export-result-annotator-counts`.

## Goal

Write a JSON provenance sidecar alongside the task CSVs on every export run, so downstream consumers can inspect run-level metadata (filter flags, counts, dataset scope) without polluting the CSV schema. Mirrors querygen's `synthetic_queries.meta.json` pattern.

## Scope

- New sidecar file: `annotation_export.meta.json` written next to the three task CSVs
- New Pydantic schema `AnnotationExportMeta` with a validator that enforces per-task dict keys (`row_counts`, `n_annotators`) cover exactly the exported tasks
- New helpers `assemble_export_meta(...)` and `write_export_meta(...)` in the export runner
- `run_export` writes the sidecar in both the empty-tasks and populated-tasks branches
- Sidecar section added to the export-pipeline design doc

## Implementation

Mirrors querygen (`core/querygen/export.py::export_queries`) as closely as possible:

```
exports/{export_id}/
├── retrieval.csv
├── grounding.csv
├── generation.csv
└── annotation_export.meta.json
```

Sidecar fields:
| Field | Source |
|-------|--------|
| `export_id` | `paths.export_dir.name` |
| `created_at` | `datetime.now(UTC)` at assembly time |
| `dataset_id` | `settings.dataset_id` (coerced `""` -> `null`) |
| `tasks` | caller-supplied |
| `include_discarded` | caller-supplied (from #159) |
| `row_counts` | computed from `task_rows` |
| `n_annotators` | computed from `task_rows` (from #160) |
| `constraint_summary` | computed from `task_rows` |

`extra="forbid"` on the model; dates serialise as ISO 8601 via `model_dump(mode="json")`. Direct write (no `.tmp` dance) matches querygen.

## Testing

- 8 schema tests (`tests/unit/core/schemas/test_annotation_export_meta.py`): constructable, `dataset_id` null path, extras forbidden, per-task key validators, negative count rejection, ISO 8601 serialisation, JSON round-trip
- 4 integration tests in `TestExportMetaSidecar`: sidecar written alongside CSVs, captures `include_discarded=True`, written when `tasks=[]`, `dataset_id` null when unset
- Updated 3 existing `AnnotationExportPaths(...)` constructions to include the new required field
- 276 annotation + api + facade + schema tests pass; 21 CLI tests pass

## References

- Design doc: `docs/design/annotation-export-pipeline.md` (new **Sidecar Manifest** section)
- Mirrors querygen sidecar pattern: `core/querygen/export.py`, `core/schemas/querygen_output.py::SyntheticQueriesMeta`
- Depends on: #159 (`include_discarded` field in sidecar), #160 (`n_annotators` field in sidecar)


src/pragmata/core/schemas/annotation_export_meta.py
saschagobel, line 1: The schema itself looks right to me, but I’d keep it in `schemas/annotation_export.py` rather than introducing a separate `annotation_export_meta.py`. That matches the querygen pattern, where the row schema and sidecar metadata schema live together in `querygen_output.py`, and avoids splitting one export contract across multiple schema modules.

tests/unit/core/annotation/test_export_runner.py
saschagobel, line 1: These cases cover the sidecar behavior well. Small test-structure nit: `json` and `export_annotations` are imported repeatedly inside each test; unless that is intentional for this file, I’d move them to module-level imports to keep the sidecar tests a bit cleaner.

tests/unit/core/schemas/test_annotation_export_meta.py
saschagobel, line 1: If we move `AnnotationExportMeta` into `schemas/annotation_export.py`, I’d also fold these tests into the existing `tests/unit/core/schemas/test_annotation_export.py` so the export row/schema contract stays tested in one place.

saschagobel (approved):
Approved, thanks — the sidecar is a clean addition and the implementation/test coverage looks solid. I only left non-blocking maintainability comments about keeping the annotation export metadata schema and tests colocated with the existing annotation export schema/tests.

saschagobel (approved):
All good now, thanks.

PR #162, opened 2026-04-23
docs(config): config and settings design doc

henrycgbaker:
## Goal

Add [`docs/design/config-and-settings.md`](docs/design/config-and-settings.md) - the design document for shared config resolution, settings precedence, secrets handling, CLI surface, and first-use error UX across all three tools (`annotation`, `querygen`, `eval`).

Annotation-specific stack lifecycle (Docker, compose, bootstrap) is in a companion PR #175.

## Scope

- **Install model** - optional extras (`pragmata[annotation]`, `[querygen]`, `[eval]`), lazy import guards scoped narrowly to the actual import site, clear `ImportError` messages naming the missing package and the extra that provides it
- **Config system** - builds on the existing `ResolveSettings.resolve` chain; adds auto-discovery of project-level config (`./pragmata.yaml` or `pyproject.toml [tool.pragmata]`, first-match-wins walk from cwd) and user-level config (`platformdirs.user_config_dir("pragmata")`). One new core dep: `platformdirs`
- **Precedence chain** - `CLI flag/kwarg > env var > explicit config_path (if passed) > auto-discovered project config > auto-discovered user config > defaults`. Explicit always beats implicit
- **Secrets** - no pragmata-owned credential store; LLM provider keys stay env-only via existing `resolve_api_key()`; Argilla delegates to its own `~/.cache/argilla/credentials` after env
- **Entrypoint signatures** - public API accepts tool-specific settings as kwargs with `UNSET` defaults; resolution always internal (existing pattern from `api/querygen.py`)
- **CLI surface** - `annotation setup`/`teardown`/`import`/`export`/`iaa`/`up`/`down`, `querygen gen-queries`, `eval run`; all commands non-interactive by default
- **First-use error UX** - never raw tracebacks; clear "X is missing, run Y to fix" messages; annotation-specific infra failures in companion PR #175

## Key decisions

- Zero-config: defaults resolve on every run via the same chain - no separate first-run path
- No config wizard, no `questionary`; `annotation setup` keeps its existing headless/flag-driven Argilla provisioning semantics
- Both `pragmata.yaml` and `pyproject.toml [tool.pragmata]` supported as project-config formats (first-match-wins, ruff pattern) - see open question below
- `PRAGMATA_<TOOL>_<KEY>` env prefix for non-secret tool settings; `PRAGMATA_CONFIG_DIR` as escape hatch for the platformdirs location

## Open question

**Two project-config formats?** Supporting both `./pragmata.yaml` and `pyproject.toml [tool.pragmata]` (first-match-wins, one file ever active per run - no divergence risk) vs. picking one. `pragmata.yaml` matches user-level file shape for easy copy-paste; `pyproject.toml` keeps pragmata alongside ruff/mypy/pytest for repos that already centralise config there.

## Implementation

Docs-only. No code changes.

## References

- [ADR-0007 - Packaging & invocation surface](docs/decisions/0007-packaging-invocation-surface.md)
- Companion PR #175 - annotation bootstrap and containerisation design docs


docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+Related: 
+- ADR-0007 (packaging & invocation surface), 
+- ADR-0012 (install & bootstrap UX) [draft]
+
+## Purpose
+
+- define how users install, configure, and first-run `pragmata`. 
+- holistically cover install model, bootstrap commands, config system, and error UX - across all three tools (`annotation`, `querygen`, `eval`)
+
+**Top-level shape** - two buckets w/ orthogonal concerns:
```
saschagobel: The §1/§2 split makes sense, but I would sharpen the boundary: §1 should cover shared settings/config resolution and entrypoint UX only, while operational bootstrap/setup concerns should remain annotation-specific unless a concrete need exists for querygen/eval.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+2. **Install is side-effect free.** No prompts, no Docker, no network calls at `pip install` time -> configuration happens on first explicit invocation (if at all, see principle 4). follows PyPI policy and `gh auth login` / `supabase` convention.
+3. **Fail clearly -> point to fix** - e.g. if optional extras not installed. Never raw traceback on a first-use error. Always: "X is not Y. Run `pragmata ... ` to fix." Follows `gh`, `supabase`, `vercel`, `dbt`, `fly`, `railway`.
+4. **Zero config OOTB.** Setup is optional; defaults work out of the box. Each tool must be usable immediately after install via opinionated defaults - no `setup` required for the happy path. `setup` exists for experienced users who need to override defaults (custom Argilla URL, alternate LLM provider, non-default workspace, etc.). First-run behaviour synthesises a working config from defaults + env vars; the wizard is a convenience, not a gate. Follows `ruff` / `black` (zero-config by design) and `supabase start` (sensible local defaults, no wizard required)
```
saschagobel: Zero config OOTB” is a good target, but I would avoid framing normal settings resolution as “first-run config synthesis.” Defaults + env + config + explicit args should be resolved consistently on every run.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+5. **Flags suppress prompts.** Interactive wizard by default; any required flag present short-circuits the corresponding prompt. Same command, both use cases. Follows [`gh auth login`](https://cli.github.com/manual/gh_auth_login).
```
saschagobel: “Flags suppress prompts” makes sense only within explicitly interactive commands such as annotation setup. I would not make interactivity the default for normal execution commands; run commands should be deterministic and scriptable unless the user explicitly enters a setup/init flow.

Edit: I noticed that setup has a different meaning here. See my comment on lines R95 to R96.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+6. **Tool-scoped complexity.** Per-tool setup commands (`pragmata annotation setup`, `pragmata querygen setup`, `pragmata eval setup`). No global `pragmata setup`. Setup is diff per tool (annotation needs Docker + Argilla creds; querygen needs LLM provider creds; eval tbc).
```
saschagobel: I agree with avoiding a global setup command, but I would not infer that every tool needs its own setup command. annotation setup seems justified because it bootstraps infrastructure; querygen credentials and provider selection should probably stay in args/env/config unless we identify a concrete setup workflow.

Edit: I noticed that setup has a different meaning here. See my comment on lines R95 to R96.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+```
+pip install pragmata                     # bare install - CLI skeleton only
+pip install pragmata[annotation]         
+pip install pragmata[querygen]           
+pip install pragmata[eval]               
+```
+
+### Lazy imports at the package boundary
+
+- optional-extra packages (`argilla`, `langchain`, etc.) must not be imported at module scope on the root import path.
```
saschagobel: I would not define extras purely by tool ownership. A dependency being used only by querygen (e.g., `langchain`) does not automatically mean it belongs in [querygen]; extras should mainly capture heavy, optional, provider-specific, or infrastructure/runtime-sensitive dependencies. Lightweight core dependencies can remain regular deps if they support a first-class package capability.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+pip install pragmata                     # bare install - CLI skeleton only
+pip install pragmata[annotation]         
+pip install pragmata[querygen]           
+pip install pragmata[eval]               
+```
+
+### Lazy imports at the package boundary
+
+- optional-extra packages (`argilla`, `langchain`, etc.) must not be imported at module scope on the root import path. 
+- each tool's public API entry point wraps its optional dep in a `try/except ImportError` and re-raises with the exact `pip install` command. Pattern borrowed from [`transformers` error handling](https://github.com/huggingface/transformers/issues/24147):
```
saschagobel: I agree with lazy optional-dependency guards at tool entrypoints, but we should catch missing direct dependencies narrowly, i.e., where they are imported e.g., in core,, so we do not mask unrelated import errors. The error should name the missing package and then point to the relevant extra.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+> `ImportError: pragmata.annotation requires the 'annotation' extra. Install with: pip install 'pragmata[annotation]'`
+> ^^^ this is the "fail loudly and point to fix" principle
+
+Failure modes must be distinguished:
+
+| Failure | Cause | Error message hint |
+|---|---|---|
+| **Extra not installed** | `pip install pragmata` (bare) -> user calls `pragmata annotation setup` | "Install with: `pip install pragmata[annotation]`" |
+| **Docker missing (annotation only)** | Has extra + config -> no Docker daemon | "Docker is required for `pragmata annotation`. Install Docker" |
+| *TODO confirm:* **"Installed but unconfigured"** would not occur happen (due to zero-config principle)| Has extra -> hasn't run setup | "Run `pragmata annotation setup` first" |
```
saschagobel: Agree: under the zero-config principle, “installed but unconfigured” should not be a normal failure mode.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+
+> NB: new deps proposed in this doc
+>
+> | Package | Where it goes | Why | Notes |
+> |---|---|---|---|
+> | [`platformdirs`](https://platformdirs.readthedocs.io/) | core deps | OS-appropriate user config dir (`~/.config/pragmata/` etc.) | ~50KB, zero deps, PyPA-endorsed (vendored in `pip`) |
+> | [`questionary`](https://questionary.readthedocs.io/) | `[annotation]` extra (and/or `[querygen]`, `[eval]` if they gain wizards -> core dep?) | Interactive setup wizard prompts (`pragmata annotation setup` etc.) | See Open question §Q-wizard-lib in this doc: commit to Questionary or use a lighter prompt lib? Prompt-toolkit conflicts largely resolved in 2.x |
```
saschagobel: `questionary` seems fine as an annotation-only extra if we keep an explicit annotation setup wizard, but I would not make it a core dependency or attach it to querygen/eval unless those tools gain a concrete interactive setup flow.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+- `overrides`: call-site kwargs (CLI flags land here as kwargs)
+- `env`: environment-derived layer
+- `config`: YAML loaded via `load_config_file(config_path)`; caller passes `config_path` explicitly
+- `defaults`: pydantic model defaults
+- Secrets resolved separately via `resolve_api_key()` (env-only, never read from `config`)
+
+*Proposed additions (non-breaking - same layer order, just adds fallbacks where layers are currently empty):*
+
+1. **Auto-discover `config_path`** when the caller doesn't pass one -> walk up from cwd for a project-local `./pragmata.yaml` or `pyproject.toml [tool.pragmata]`; fall back to `platformdirs.user_config_dir("pragmata") / "config.yaml"`. Explicit `config_path` still wins, so every current caller behaves identically. See §1.1.3.
+2. **Add a credentials-file fallback for `resolve_api_key()`** below env vars -> if the env var is unset, read from `~/.config/pragmata/credentials` before raising `MissingSecretError`. Env still wins.
```
saschagobel: I’m comfortable with auto-discovery for non-secret config, but I’d separate that from persistent credentials. For secrets, args/env should remain the primary supported path; adding `~/.config/pragmata/credentials` is a bigger security and lifecycle decision and should probably be deferred.

I don’t think a user-level credentials file is inherently wrong — it is common in CLIs — but it is a larger product/security surface (we manage a local secret store, cross-platform) than non-secret config discovery. If we add it, we should explicitly define permissions, redaction, rotation/update behavior, cross-platform handling, and precedence. For now I’d prefer env/kwargs as the primary secret path, with provider-native stores delegated where available.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+Precedent: [ruff](https://docs.astral.sh/ruff/configuration/) (supports both `ruff.toml` and `pyproject.toml [tool.ruff]`, first match wins), [dbt](https://docs.getdbt.com/docs/core/connect-data-platform/profiles.yml) (project `dbt_project.yml` + user `~/.dbt/profiles.yml`), [black](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html#configuration-via-a-file) (walks up for `pyproject.toml`).
+
+### 1.1.4 Precedence chain
+
+Universal pattern (aws/terraform/kubectl/dbt/pip), extended with a project layer (§1.1.3):
+
+```
+CLI flag  >  env var  >  credentials file (secrets)  >  project config (./pragmata.yaml or pyproject.toml)  >  user config (~/.config/pragmata/config.yaml)  >  built-in defaults
+```
```
saschagobel: I’d make the explicit `config_path` precedence unambiguous. My expectation would be: 

```
explicit kwargs / CLI flags
> env vars
> explicit config_path, if provided
> auto-discovered project config
> auto-discovered user config
> defaults
```

explicit config path wins over any auto-discovered project/user config; project config only enters when no explicit config path is passed. Otherwise “explicit beats implicit” is violated.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+- follows [gcloud's `CLOUDSDK_SECTION_PROPERTY`](https://docs.cloud.google.com/sdk/docs/properties).
+- rejected: pydantic-settings' `PRAGMATA__ANNOTATION__URL` double-underscore - less shell-friendly, harder to type.
+
+> *Current state*: no systematic `PRAGMATA_*` prefix exists yet. Today the codebase reads a few hardcoded env vars directly (notably `ARGILLA_API_URL` in [`api/annotation_setup.py`](../../src/pragmata/api/annotation_setup.py)). The proposal is to add prefix-based resolution *alongside* canonical provider env vars (secrets), not to replace the latter.
+
+**Secrets are the exception - they use canonical provider env vars, not the `PRAGMATA_*` prefix.** See §1.1.5; already implemented in `API_KEY_ENV_VARS` / `resolve_api_key()`.
+
+Small set of shared top-level vars (outside per-tool scoping, all proposed):
+- `PRAGMATA_CONFIG_DIR` - override config location
+- `PRAGMATA_WORKSPACE_DIR` - override workspace root
```
saschagobel: PRAGMATA_CONFIG_DIR makes sense as an escape hatch overriding the platformdirs location. I’m less convinced by PRAGMATA_WORKSPACE_DIR; workspace/base dir is already an explicit kwarg/flag with cwd default, adding a global env override may create hidden state without much benefit.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+>- *Delegate model:* if user has already run `argilla login`, `pragmata annotation setup` detects that and skips the API key prompt. Keeps one source of truth and composes cleanly with the existing `ARGILLA_API_KEY` env var resolution. BUT couples us to Argilla's file format + location; breaks if they rename/relocate it; harder to reason about precedence when two stores exist.
+>- *Maintain our own model:* `~/.config/pragmata/credentials` holds Argilla + LLM provider keys in one place. Consistent resolution pattern across every secret pragmata needs. BUT duplicates Argilla's store if the user has also run `argilla login`; pragmata has to write/chmod/own the file; users need to update it in two places if they rotate keys (could setup a copy/symlink system, but brittle).
+>- *Recommendation:* delegate for Argilla specifically (it's the only provider with its own persistent credential store); maintain our own for LLM providers (no equivalent exists). Resolution order for Argilla becomes: `kwarg > ARGILLA_API_KEY env > ~/.cache/argilla/credentials > ~/.config/pragmata/credentials > MissingSecretError`.
```
saschagobel: For Argilla, I would prefer delegating to Argilla’s own credential store after ARGILLA_API_KEY, rather than also adding a pragmata-owned fallback (see my comment on lines R95 to R96). Maintaining two persistent credential stores for the same service seems likely to create rotation and precedence confusion.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+> *Current state*: no systematic `PRAGMATA_*` prefix exists yet. Today the codebase reads a few hardcoded env vars directly (notably `ARGILLA_API_URL` in [`api/annotation_setup.py`](../../src/pragmata/api/annotation_setup.py)). The proposal is to add prefix-based resolution *alongside* canonical provider env vars (secrets), not to replace the latter.
+
+**Secrets are the exception - they use canonical provider env vars, not the `PRAGMATA_*` prefix.** See §1.1.5; already implemented in `API_KEY_ENV_VARS` / `resolve_api_key()`.
+
+Small set of shared top-level vars (outside per-tool scoping, all proposed):
+- `PRAGMATA_CONFIG_DIR` - override config location
+- `PRAGMATA_WORKSPACE_DIR` - override workspace root
+- TODO: define here
+
+### 1.1.5 Secrets
```
saschagobel: See my comment on lines R95 to R96.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+- Internal resolution: `ToolSettings.resolve(config=..., overrides={...})` (§1.1)
+- Secrets via args or env only; never in config files (§1.1.5)
+
+**CLI ↔ API coupling.** CLI commands are thin wrappers over the API - flags land as kwargs with `UNSET` for un-passed; [ADR-0007](../decisions/0007-packaging-invocation-surface.md).
+
+**What this means for external-service wiring.** Users should be able to pass anything relevant for the client, including external service URLs (e.g. an external Postgres connection string) - this is the Infra UX side of the same contract (§2.2.3: `--external-postgres <url>` etc.). The API entrypoint accepts them as kwargs; CLI surfaces them as flags; env-var and config-file fallbacks work the same way as for every other setting.
+
+## 1.3 CLI surface
+
+- Each tool has its own config wizard verb
```
saschagobel: I agree that, if a config writer exists, it should be tool-scoped. But I would avoid reusing `setup` for config-file generation. It is already semantically overloaded and currently means Argilla provisioning. If we add a config writer, `configure` is clearer than `setup`.

That said, I’m not convinced an interactive config overwrite wizard is the right default UX for all tools or even at all. If the proposed wizard is only for overriding defaults, that is effectively an interactive config editor, whereas the simpler and more transparent UX is documented flags/env/config plus, at most, a command that writes a commented config template.

I think that i the more common pattern for developer-facing Python tools; interactive wizards are usually justified by auth, project scaffolding, cloud account selection, or infrastructure bootstrap.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+$ pragmata annotation import foo.json
+Error: Argilla stack is not running.
+  Run: pragmata annotation up
+
+NB: given OOTB works with opinionated defaults principle, we would NOT have following. (TODO: confirm this):
+
+$ pragmata annotation import foo.json      
+Error: annotation is not configured.
+  Run: pragmata annotation setup
```
saschagobel: Agreed: under zero-config defaults, “annotation is not configured; run setup” should not be a normal first-use error..

docs/design/install-and-bootstrap.md
saschagobel, line 1: I think this doc may be trying to cover two design problems that are related but separable enough to split.

§1 is mostly an API/config UX proposal: settings resolution, config discovery, env var conventions, secret lookup, entrypoint signatures, and CLI/API coupling across all tools. 

§2 is annotation infra UX: Docker stack lifecycle, local service bootstrap, provisioning, upgrade/uninstall behavior, etc. These have different audiences, implementation risks, and review questions.

I’d suggest splitting this into two design docs:
1. one for shared API/config UX across `annotation`, `querygen`, and `eval`; and
2. one for annotation-specific infra/bootstrap UX.

That would keep this doc shorter and make the open questions easier to resolve independently. It would also avoid config-resolution decisions getting bundled into the install/bootstrap discussion.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+## 2.1 Stack composition
+
+>TODO: need to decide where Docker Compose should live.
+> - Current is in `deploy/` separate from `annotation/` -> clean separation, but we might want to rethink this (see this 2.1 + 2.2.1)
+
+Bundled stack (mirrors the existing [`deploy/annotation/docker-compose.dev.yml`](../../deploy/annotation/docker-compose.dev.yml)):
+
+All bundled by default (zero-config principle). Each backing service (postgres/elasticsearch/redis) is opt-out-able via a Compose profile
+
+**Single shipped compose file, not a dev/prod pair.** Surveyed PyPI-distributed Docker-wrapping tools (Supabase, Airbyte `abctl`, Airflow, Dagster, Prefect, MLflow) all converge on **one shipped compose artefact** - either a single file or a programmatic equivalent. Nobody ships parallel `docker-compose.prod.yml` + `docker-compose.dev.yml` files side by side because it forces users to answer "which one do I run?" before they've done anything. Instead:
```
saschagobel: I agree with a single shipped compose artifact rather than parallel dev/prod files. For PyPI distribution, I’d prefer treating the runtime compose file as package data under `src/pragmata/annotation/` and resolving it with `importlib.resources`; `deploy/` can remain the contributor/dev location.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+| **Z. Eject** | Start with Y; `pragmata annotation eject` copies compose out and pragmata then uses the ejected copy, warning the user they own it from there | Zero-config | Explicit escape hatch, clean managed-vs-owned contract | None for non-ejected users; ejected users own drift | [create-react-app `eject`](https://create-react-app.dev/docs/available-scripts#npm-run-eject), [Expo eject-to-bare-workflow](https://docs.expo.dev/archive/customizing/), [Next.js manual config extraction](https://nextjs.org/docs) |
+
+**Recommendation (X).** User-editable default, preserves standard Docker contract (compose YAML is editable), keeps zero-config happy-path intact (packaged file IS the default), and researchers who want to change a port open the file like they would for any other Docker stack. Drift mitigation is deliberately minimal - we do not expect to ship compose updates often; over-engineering for rare upgrades is waste.
```
saschagobel: I currently lean toward the managed/locked compose model for v0.1. Users should not have to understand or edit Docker Compose YAML for the supported paths. The important customization surface is the CLI/env/config layer (`--external-postgres`, `--external-elastic`, ports, etc.); if that surface is sufficient, keeping the compose file package-owned avoids user drift and keeps upgrades simpler.

I’m less convinced by option X as the default. Copying an editable compose file into the user config dir creates an ownership/drift problem immediately and makes the happy path more Docker-centric than necessary. I’d reserve user-owned compose files for an explicit advanced escape hatch, not the default path.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+**Reading the divergence.** The two positions may be a framing mismatch rather than a real disagreement. Position A's "bash script" phrasing is less specific than it sounds - "*replaced likely by bash script*" reads as an aside; "*bash script helpers*" describes the wiring layer between user config and Compose, not a user-facing install script. If "bash script" means "whatever replaces the Makefile for unattended/prod installs" - that's satisfied by any of options A/B/C below. If it specifically means a shipped `scripts/bootstrap.sh` - that's option B. Worth confirming; the design doesn't hinge on the answer, but option A and option B have different maintenance implications.
+
+### 2.6.3 Options for the user-facing install surface
```
saschagobel: I agree with option A for v0.1: no shipped bootstrap script, just document `pip install "pragmata[annotation]"` plus `pragmata annotation up`. If demand appears, add option C on top.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+| **Compose file distribution** | Dev-only file in `deploy/`; **not shipped** in the installed wheel | Ship `src/pragmata/annotation/docker-compose.yml` as package data (single file, prod-first, dev override via separate file - see §2.2)* |
+| **Package data** | `importlib.resources` already used for [`core/annotation/collapsible_field.html`](../../src/pragmata/core/annotation/collapsible_field.html). | Same mechanism for the compose file |
+| **`questionary` / wizards** | No wizard today - `setup` is headless-flag-driven only | Add `questionary` under `[annotation]` (or other tools as needed)|
+
+
+## Open questions
+
+not comprehensive - several are inline thoughout above
+
+### §Q-prod-script: ship a per-tool bootstrap script for `annotation`?
```
saschagobel: No, not for v0.1.0.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+
+### §Q-wizard-lib: Questionary or minimal prompts?
```
saschagobel: Defer unless we decide to ship an actual interactive flow.

At the moment, I would not add a general config overwrite wizard. Defaults, flags, env vars, config files, and possibly a commented config-template command should be enough for normal configuration. If we later add an interactive annotation-specific flow that validates external services or handles provisioning/auth, then Questionary is a reasonable annotation-extra dependency. It should not be core, and it should not be added for querygen/eval without a concrete wizard workflow.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+Research strongly prefers [Questionary](https://questionary.readthedocs.io/); 2.1 addresses the older prompt_toolkit clash. Alternative: a lighter library for the handful of prompts we actually need.
+
+### §Q-auth-split: `pragmata annotation auth` / `login` separate command, or part of `setup`?
```
saschagobel: Neither for now.

I would not add a pragmata-managed auth/login command unless we decide to persist credentials ourselves, which I do not think we should do at this stage.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+Argilla API keys rotate independently of compose config. `gh` and `supabase` split auth from config; `aws` combines. Research leans split; current draft combines.
+
+
+### §Q-wizard-placement: where and how do wizards appear?
```
saschagobel: Nowhere for v0.1.

I would not add a general interactive config overwrite wizard at this stage.

docs/design/install-and-bootstrap.md
```diff
@@ -0,0 +1,604 @@
...
+
+Still open - noted in guiding principles but not resolved.
+
+### §Q-configured-check: include a "configured" pre-flight check?
```
saschagobel: No, not under the zero-config OOTB principle.

docs/design/config-and-settings.md
```diff
@@ -0,0 +1,331 @@
...
+
+## Install model
+
+### Optional extras
+
+```
+pip install pragmata                     # bare install - CLI skeleton only
+pip install pragmata[annotation]         
+pip install pragmata[querygen]           
+pip install pragmata[eval]
```
saschagobel, line 32: I think we should include a `pragmata[full]` extra in the install model, even if we keep the per-tool extras. Users often want the “just give me all runtime capabilities” path.

docs/design/config-and-settings.md
```diff
@@ -0,0 +1,331 @@
...
+Failure modes must be distinguished:
+
+| Failure | Cause | Error message hint |
+|---|---|---|
+| **Extra not installed** | `pip install pragmata` (bare) -> user calls `pragmata annotation ...` | "Install with: `pip install pragmata[annotation]`" |
+| **Docker missing (annotation only)** | Has extra -> no Docker daemon | "Docker is required for `pragmata annotation`. Install Docker" |
+
+Under the zero-config principle, "installed but unconfigured" is **not** a normal failure mode - defaults are sufficient and resolution is consistent on every run.
+
+> **New dependdencies proposed in this doc**
```
saschagobel: typo in "dependdencies"

docs/design/config-and-settings.md
saschagobel, line 1: I agree with the design, but the doc is somewhat repetitive around zero-config, no wizard, non-interactive commands, and secrets-not-in-config. Not blocking.

saschagobel (approved):
Thanks, that is directionally strong now. Just a few non-blocking tightening points.

docs/design/config-and-settings.md
henrycgbaker: Fixed in 3dc6b04.

docs/design/config-and-settings.md
henrycgbaker: Agreed - tracking as a separate follow-up PR so this one can merge as-is. Will add `pragmata[full]` = annotation + querygen + eval to the install model.

PR #163, opened 2026-04-24
test(integration): progressive preflight for annotation stack

henrycgbaker:
## Goal

`make test-all` produces errors when Docker isn't running -> now clean skips instead of 14 noisy fixture-setup errors. 
+ tell user up front which precondition failed (fail fast & loudly) so they know whether to start Docker, run `make docker-up`, or wait for warm-up.

## Scope

- Progressive preflight at session start: docker-cli -> docker-daemon -> argilla-tcp -> argilla-api. Cheapest checks first; first failure determines skip reason.
- `pytest_report_header` hook prints the status of each layer programmatically at session start.
- Session-scoped `annotation_stack_status` fixture exposes cached result; autouse skip fixture consults it for any `annotation`-marked test.
- Two integration files whose module-scoped `client` fixture was eagerly calling `rg.Argilla(...)` (which validates connection in `__init__` and caused error messages) now depend on `annotation_stack_status` and `pytest.skip` cleanly when the stack isn't ready.

## Implementation

- `tests/conftest.py` — `AnnotationStackStatus` dataclass holding the 4-layer check result, `_check_annotation_stack()` that runs the progressive probe once at `pytest_configure`, `pytest_report_header` banner, autouse skip fixture.
- `tests/integration/test_annotation_setup.py`, `tests/integration/test_annotation_import.py` - `client` fixture gains the `annotation_stack_status` dependency and skips with the preflight's reason if stack isn't ready.

## Testing

Before (Docker off): `634 passed, 3 skipped, 14 errors` (exit 1) — integration tests errored during fixture setup with raw `httpx.ConnectError`.

After (Docker off): `622 passed, 15 skipped, 0 errors` (exit 0) — preflight banner:

```
annotation stack: docker-cli [ok] daemon [--] argilla-tcp [--] argilla-api [--] — annotation integration tests will be skipped (Docker daemon not running (try: open -a Docker))
```

With Docker + Argilla up everything runs.

Lint + mypy green.

## References

- Follow-up in stacked PR: slim down `test_dev_stack_integration.py` (calls stale `make setup`/`make teardown` targets).

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,21 +2,66 @@
 
 import shutil
 import socket
+import subprocess
+import urllib.error
+import urllib.request
+from dataclasses import dataclass
 
 import pytest
 
 ARGILLA_DEFAULT_HOST = "localhost"
 ARGILLA_DEFAULT_PORT = 6900
+ARGILLA_DEFAULT_API_KEY = "argilla.apikey"
 _CONNECT_TIMEOUT_S = 2
+_HTTP_TIMEOUT_S = 3
 
 
-def _docker_available() -> bool:
-    """Check whether Docker CLI is on PATH."""
+@dataclass(frozen=True)
+class AnnotationStackStatus:
+    """Result of progressive preflight checks for the annotation stack."""
+
+    docker_cli: bool
+    docker_daemon: bool
+    argilla_tcp: bool
+    argilla_api: bool
+
+    @property
+    def ready(self) -> bool:
+        return self.docker_cli and self.docker_daemon and self.argilla_tcp and self.argilla_api
+
+    @property
+    def skip_reason(self) -> str | None:
+        if not self.docker_cli:
+            return "Docker CLI not on PATH"
+        if not self.docker_daemon:
+            return "Docker daemon not running (try: open -a Docker)"
+        if not self.argilla_tcp:
+            return f"Argilla not reachable at {ARGILLA_DEFAULT_HOST}:{ARGILLA_DEFAULT_PORT} (try: make docker-up)"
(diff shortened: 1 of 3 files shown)


henrycgbaker (commented):
PR 163 ready for review. Writing more detail here:
- potential areas of expected pushback is that this is over engineering for dev-only functionality - i.e. pytest_report_header() and pytest_terminal_summary() are pure convenience/diagnostics rather than functionality extending.
- could generalise them to be general infra diagnostics (rather than pytest-specific) but based on initial comments on 162 (that we don't plan bootstrap scripts) they likely won't be reused.
- can remove / narrow if this feels like creep

henrycgbaker:
edit: I have removed pytest_terminal_summary() as it's purely cosmetic.

tests/conftest.py
```diff
@@ -2,41 +2,135 @@
...
+    @property
+    def ready(self) -> bool:
+        return self.docker_cli and self.docker_daemon and self.argilla_tcp and self.argilla_api
+
+    @property
+    def skip_reason(self) -> str | None:
+        if not self.docker_cli:
+            return "Docker CLI not on PATH"
+        if not self.docker_daemon:
+            return "Docker daemon not running (try: open -a Docker)"
```
saschagobel: Non-blocking nit: "open -a Docker" is macOS-specific; I’d make that skip reason platform-neutral, e.g. “Docker daemon not running (start Docker, then retry)”.

saschagobel (approved):
Approved — this is a sound test-infra improvement. Only left one non-blocking nit.

PR #164, opened 2026-04-24
test(integration): slim dev-stack integration tests

henrycgbaker:
## Goal

1. the three tests in `tests/integration/test_dev_stack_integration.py` all shell out to `make setup` / `make teardown` = targets that don't exist in the current Makefile, so every test would fail even with Docker running. 
2. they duplicate work that the preflight in #163 already does, and test fundamental (well-tested) docker primitives (rather than our added functionality) -> prune

## Scope

Delete stale tests, keep the one unique assertion that genuinely exercises our stack (authenticated API returns the expected owner payload).

## Implementation

Before: 3 tests × ~50 lines, each spinning up and tearing down the stack (~120s each) inside pytest.

After: 1 test × ~15 lines, asserting `/api/v1/me` returns `{username: argilla, role: owner}`. The preflight from #163 gates collection -> stack orchestration belongs in the Makefile (`make docker-up`, `make test-stack`), not pytest.

Dropped tests and rationale:

| Test | Why dropped |
|---|---|
| `test_stack_boots_healthy` | Redundant with preflight's `argilla_api` check |
| `test_argilla_api_authenticated` | Kept (renamed to `test_argilla_api_authenticated_as_owner`) |
| `test_teardown_removes_volumes` | Orchestration concern - `docker compose down -v` is a well-trusted primitive; (could fold into `make test-stack` if needed) |

## References

- Stacked on #163 (`test/annotation-stack-preflight`). Needs that to merge first. After #163 lands, rebase this onto `main`.

### DIFF
--- a/tests/integration/test_dev_stack_integration.py
+++ b/tests/integration/test_dev_stack_integration.py
@@ -1,9 +1,4 @@
-"""Integration tests for the Argilla dev stack.
-
-Requires the stack to be running (make docker-up). The annotation-stack
-preflight in tests/conftest.py gates collection, so these tests skip
-cleanly when the stack is unavailable.
-"""
+"""Integration tests for the Argilla dev stack."""
 
 import json
 import urllib.request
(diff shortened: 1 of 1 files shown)


saschagobel (approved):
Thanks, this looks good.

PR #165, opened 2026-04-25
test(integration): purge orphan datasets in annotation cleanup fixtures

henrycgbaker:
## Goal

Integration tests that exercise full teardown (empty `dataset_id`) accumulate orphan datasets across runs - either when a dataset uses a non-canonical name, or when a test exits midway. Argilla refuses to delete a workspace while any dataset is linked, so the next `teardown_resources` call hits `409 ConflictError` and the suite fails.

Scope: test fixtures only. Production callers own their `dataset_id` and don't accumulate orphans, so `teardown_resources` keeps its current targeted behaviour.

## Implementation

New helper `tests/integration/_argilla_cleanup.purge_workspace_datasets(client, ws_base)` deletes every dataset in a workspace.

`clean_slate` (test_annotation_setup.py) and `clean_environment` (test_annotation_import.py) call the purge for each workspace in `AnnotationSettings.workspace_dataset_map` before `teardown_resources`. Both fixtures now share the same `purge → teardown` shape. Scoped to the declared workspaces only, not a global `client.datasets` sweep - keeps blast radius contained on shared/dev Argilla instances.

Earlier draft pushed the orphan-purge into `core/annotation/setup.py` so a full teardown would delete every dataset in each workspace before deleting the workspace. Rejected because:
- Multi-tenant blast radius: in shared Argilla deployments, that wipes datasets owned by other runs/users
- Silent contract change: same call signature, different (more destructive) behaviour
- Wrong layer: the symptom is test-fixture state, not a production-API gap

If a production-side purge is wanted later, the right shape is an explicit `purge_workspace(ws_base, *, confirm=True)` function, not an implicit branch in `teardown_resources`.

## Testing

- `pytest tests/integration/test_annotation_setup.py tests/integration/test_annotation_import.py -m "integration and annotation"` → 14 passed (with the test user pre-purged; the user-leak failure is the separate test-isolation bug being fixed in #166).
- Lint + mypy green.

## Follow-up

- #166 — function-scoped fixtures + user-purge, addressing the test-isolation bug referenced above.


henrycgbaker:
open question (@saschagobel let me know if you have thoughts otherwise I'll return to this this evening): 

better in tests just to remove ALL datasets cleanly (global, non-scoped)
```
for ds in client.datasets:
    ds.delete()
```
rather than scope to AnnotationSettings.workspace_dataset_map ?

NB pending this decision -> update #166 workflow accordingly

saschagobel:
> open question (@saschagobel let me know if you have thoughts otherwise I'll return to this this evening):
> 
> better in tests just to remove ALL datasets cleanly (global, non-scoped)
> 
> ```
> for ds in client.datasets:
>     ds.delete()
> ```
> 
> rather than scope to AnnotationSettings.workspace_dataset_map ?
> 
> NB pending this decision -> update #166 workflow accordingly

I’d avoid the global `for ds in client.datasets: ds.delete()` approach. Even though this is test code, it is too broad for shared/dev Argilla instances and undermines the blast-radius argument for keeping purge behavior out of production teardown.

My preference: purge all datasets, but only within the workspaces declared by the integration-test `AnnotationSettings.workspace_dataset_map`. That gives the fixtures clean ownership of their test topology without risking unrelated datasets.

henrycgbaker (commented):
> > open question (@saschagobel let me know if you have thoughts otherwise I'll return to this this evening):
> > 
> > better in tests just to remove ALL datasets cleanly (global, non-scoped)
> > 
> > ```
> > for ds in client.datasets:
> >     ds.delete()
> > ```
> > 
> > rather than scope to AnnotationSettings.workspace_dataset_map ?
> > 
> > NB pending this decision -> update #166 workflow accordingly
> 
> I’d avoid the global `for ds in client.datasets: ds.delete()` approach. Even though this is test code, it is too broad for shared/dev Argilla instances and undermines the blast-radius argument for keeping purge behavior out of production teardown.
> 
> My preference: purge all datasets, but only within the workspaces declared by the integration-test `AnnotationSettings.workspace_dataset_map`. That gives the fixtures clean ownership of their test topology without risking unrelated datasets. 

Great - that's already the shape here: helper purges per-workspace, and both clean_slate / clean_environment iterate workspace_dataset_map rather than global

saschagobel (approved):
Thanks, all good!

PR #167, opened 2026-04-25
refactor(makefile): consolidate docker-up variants behind PROFILE param

henrycgbaker:
## Goal

Replace four near-identical `docker-up*` Makefile targets with a single parameterised `make docker-up PROFILE=<name>`. Reduces duplication, makes the profile choice explicit at the call site, and aligns with the principle that the Makefile passes through to the underlying compose CLI rather than baking configuration into target names.

## Scope

`Makefile` only.

Before:
```makefile
docker-up
docker-up-external-pg
docker-up-external-es
docker-up
```

After:
```makefile
docker-up                                # defaults to PROFILE=all-bundled
make docker-up PROFILE=external-pg       # explicit override
make docker-up PROFILE=external-es
make docker-up PROFILE=external
```

## Implementation

- `PROFILE ?= all-bundled` makes `all-bundled` the default while allowing override on the command line.
- A single `case` statement in the recipe routes by `PROFILE`:
  - Sets the appropriate `--profile` flag (or empty for `external`).
  - Runs the env-var validation specific to that profile (`ARGILLA_DATABASE_URL` for `external-pg`, etc.).
- Unknown PROFILE values fail with a clear error listing the valid options.
- Top-of-file comment block documents the four valid values.
- `down/stop/logs/status` still use `$(ALL_PROFILES)` so they affect every profile's containers regardless of which one started the stack.

## Testing

- `make docker-up` (default) → starts stack as `all-bundled`. Argilla up at :6900 — verified.
- `make docker-up PROFILE=garbage` → `Error: unknown PROFILE 'garbage' (valid: all-bundled, external-pg, external-es, external)`, exit 1.
- `make docker-up PROFILE=external-pg` (without `ARGILLA_DATABASE_URL` in `.env`) → `Error: ARGILLA_DATABASE_URL must be set for PROFILE=external-pg`, exit 1.
- `make ci` green (598 passed, 39 deselected).
- `make test-all` green (637 passed).
- `tests/unit/test_dev_stack.py::test_makefile_defines_expected_targets` still passes — the unit test only pins canonical targets (`docker-up`, `docker-down`, etc.), not the variant targets that this PR removes.

## References

- Stacked on #166. Final PR in the integration-test cleanup chain.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,19 +2,13 @@ COMPOSE_FILE := deploy/annotation/docker-compose.dev.yml
 ENV_FILE     := deploy/annotation/.env
 ENV_EXAMPLE  := deploy/annotation/.env.dev.example
 
-.PHONY: docker-up docker-down docker-stop docker-logs docker-status ensure-env lint type-check test-stack test test-integration test-all ci
+MAKE .PHONY: docker-up docker-down docker-stop docker-logs docker-status ensure-env lint type-check test-stack test test-integration test-all ci
 
 # Compose profile options for the multi-target docker-* commands. Profiles
 # determine which backing services run as containers vs. expected externally.
 ALL_PROFILES := --profile all-bundled --profile external-pg --profile external-es
 
-# Default profile for `make docker-up`. Override on the command line:
-#   make docker-up PROFILE=external-pg
-# Valid values:
-#   all-bundled  — Argilla + Postgres + Elasticsearch + Redis all in containers (default)
-#   external-pg  — external Postgres (set ARGILLA_DATABASE_URL); ES + Redis bundled
-#   external-es  — external Elasticsearch (set ARGILLA_ELASTICSEARCH); Postgres + Redis bundled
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -2,11 +2,15 @@ COMPOSE_FILE := deploy/annotation/docker-compose.dev.yml
 ENV_FILE     := deploy/annotation/.env
 ENV_EXAMPLE  := deploy/annotation/.env.dev.example
 
-.PHONY: docker-up docker-up-external-pg docker-up-external-es docker-up-external docker-down docker-stop docker-logs docker-status ensure-env lint type-check test-stack test test-integration test-all ci
+MAKE .PHONY: docker-up docker-down docker-stop docker-logs docker-status ensure-env lint type-check test-stack test test-integration test-all ci
```
saschagobel: MAKE .PHONY: appears to be an accidental typo; because .PHONY is still present it may work, but it also defines an unintended MAKE target, so I’d change it back to plain .PHONY

saschagobel (approved):
Approved — the refactor is sound and simplifies the Makefile nicely; only one non-blocking cleanup comment.

PR #171, opened 2026-04-30
refactor(annotation): introduce TaskOverlap and nested workspace_dataset_map

henrycgbaker:
## Goal

Replace the dataset-wide ``min_submitted`` setting with a per-task ``TaskOverlap`` model, opening room for per-task overlap configuration without changing current runtime behaviour.

## What changes

- New `TaskOverlap` Pydantic model: `production_min_submitted: int = 1`, `calibration_min_submitted: int | None = 3`
- `AnnotationSettings.workspace_dataset_map`: `dict[str, list[Task]]` → `dict[str, dict[Task, TaskOverlap]]`
- Top-level `min_submitted: int = 1` removed (now lives on `TaskOverlap`)
- `fan_out_records()` reads `overlap.production_min_submitted` (semantically identical to the old `settings.min_submitted` for default config)
- `--min-submitted` CLI flag removed; `min_submitted` kwarg on `setup()` removed
- Existing tests updated; new tests for `TaskOverlap` defaults and validation

**Breaking change**: Configs that set top-level `min_submitted` need to migrate. 

## Testing

- 610 unit tests pass (new tests added for TaskOverlap)
- `mypy src/pragmata` clean

### DIFF
--- a/src/pragmata/api/annotation_setup.py
+++ b/src/pragmata/api/annotation_setup.py
@@ -20,7 +20,6 @@ def setup(
     *,
     api_url: str | Unset = UNSET,
     api_key: str | Unset = UNSET,
-    min_submitted: int | Unset = UNSET,
     base_dir: str | Path | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
 ) -> SetupResult:
@@ -31,7 +30,9 @@ def setup(
     workspaces. Existing resources are skipped.
 
     Datasets are not created here — they are auto-created on import,
-    scoped by dataset_id.
+    scoped by dataset_id. Per-task overlap (production and calibration
+    ``min_submitted``) is configured via ``workspace_dataset_map`` in the
+    YAML config or ``AnnotationSettings``.
 
     Settings are resolved from config file and/or keyword overrides. Omitted
     values fall through to config-file defaults, then built-in defaults.
@@ -44,7 +45,6 @@ def setup(
         users: User accounts to provision. Pass None to skip user setup.
         api_url: Argilla server URL.
         api_key: Argilla API key.
-        min_submitted: Minimum annotations required per record.
         base_dir: Workspace base directory. Defaults to cwd.
         config_path: Path to YAML config file for settings resolution.
 
@@ -56,7 +56,6 @@ def setup(
         env={"argilla": {"api_url": os.environ.get("ARGILLA_API_URL")}} if os.environ.get("ARGILLA_API_URL") else None,
         overrides={
             "argilla": {"api_url": api_url},
-            "min_submitted": min_submitted,
             "base_dir": base_dir,
         },
     )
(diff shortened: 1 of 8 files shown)


src/pragmata/api/annotation_setup.py
```diff
@@ -31,7 +30,9 @@ def setup(
     workspaces. Existing resources are skipped.
 
     Datasets are not created here — they are auto-created on import,
-    scoped by dataset_id.
+    scoped by dataset_id. Per-task overlap (production and calibration
+    ``min_submitted``) is configured via ``workspace_dataset_map`` in the
+    YAML config or ``AnnotationSettings``.
```
saschagobel: Non-blocking, but I would avoid pointing users at `AnnotationSettings` as a configuration surface here. Using YAML config for per-task overlap is fine, but `AnnotationSettings` should remain the internal/defaults model rather than something users are expected to instantiate or edit directly. If there are important overlap defaults there, we should expose them through the relevant public API/config documentation rather than documenting the settings model itself as user-facing.

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -18,25 +18,48 @@ class ArgillaSettings(BaseModel):
     api_url: str | None = None
 
 
+class TaskOverlap(BaseModel):
+    """Per-task overlap topology: production and (optional) calibration thresholds.
+
+    Calibration is declared here so future imports can route a subset of
+    records to a separate calibration dataset for IAA. This PR only reads
+    ``production_min_submitted``; calibration support arrives in subsequent
+    PRs.
```
saschagobel: Small docstring nit: I would remove “This PR only reads `production_min_submitted`; calibration support arrives in subsequent PRs.” from the model docstring. That is useful PR context, but it will become stale in the codebase.

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -18,25 +18,48 @@ class ArgillaSettings(BaseModel):
     api_url: str | None = None
 
 
+class TaskOverlap(BaseModel):
```
saschagobel, line 21: Non-blocking: I think the `TaskOverlap` model shape is sensible, but the overlap knobs should not be reachable only through the nested config. Since this PR removes the previous public `min_submitted` API/CLI override, the replacement levers for steering overlap (`production_min_submitted`, and eventually `calibration_min_submitted` / partitioning) should be exposed through the relevant public API and CLI arguments as well. Otherwise we make an important runtime behavior configurable internally and via config.yml but not conveniently steerable from the supported invocation surfaces.

saschagobel (approved):
Approved with comments. The refactor direction looks sound, but the new overlap knobs should be exposed through the supported API/CLI surfaces in line with our established settings wiring pattern, rather than being reachable only through the nested settings/config shape.

EDIT: Small moderation: I looked ahead in the follow-up PRs and saw that `calibration_fraction` is exposed there, which seems right — though its default should live in settings and use the same UNSET override pattern as elsewhere, not be hardcoded in the signature. For the production/calibration `min_submitted` knobs, I agree those may be advanced enough to stay config-only rather than API/CLI surface, but then I would still avoid pointing users directly at AnnotationSettings, since that remains strictly internal.

PR #172, opened 2026-04-30
refactor(annotation): always-suffixed dataset naming via dataset_name helper

henrycgbaker:
## Goal

Centralise Argilla dataset naming behind a single helper, and switch to always-suffixed names so production and calibration datasets coexist cleanly.

## What changes

- New `dataset_name(task, calibration, dataset_id)` helper in `argilla_task_definitions.py`. Always-suffixed: `task_<task>_production` or `task_<task>_calibration`. `dataset_id` is appended for run-scoping when present.
- Callsites updated to use the helper: `record_builder.fan_out_records`, `setup.teardown_resources`, `export_fetcher.fetch_task`. Old `apply_suffix(DATASET_NAMES[task], dataset_id)` calls removed.
- Integration tests updated for new dataset names (e.g. `retrieval_run1` → `retrieval_production_run1`).
- New unit tests cover the helper across all tasks, both purposes, and with/without `dataset_id`.

**Behaviour change for users**: default dataset name was `retrieval`; it is now `retrieval_production`. 

## Why

for IAA: centralising the naming convention now (before any calibration logic exists) keeps later changes additive

## Testing

- 618 unit tests pass (4 new for the helper)
- `mypy src/pragmata` clean

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -32,7 +32,7 @@ jobs:
         uses: actions/checkout@v6
 
       - name: Setup Node
-        uses: actions/setup-node@v6
+        uses: actions/setup-node@v6.3.0
         with:
           node-version: 20
           cache: npm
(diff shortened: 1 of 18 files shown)


saschagobel (approved):
This is in nice shape, thanks.

PR #173, opened 2026-04-30
feat(annotation): add PartitionManifest schema, calibration field, import paths

henrycgbaker:
## Goal

Add the schemas and path bundle that partial-overlap calibration support depends on. 

### Schemas

- New `PartitionManifestEntry` and `PartitionManifest` in `core/schemas/annotation_import.py`. The manifest is a persistent record_uuid → assignment map keyed by `dataset_id`. Each entry records the assignment (calibration vs production), the import that produced it, the fraction in force at that import, and a timestamp.
- `calibration: bool = False` field on `AnnotationBase` in `core/schemas/annotation_export.py`. Defaulted so existing construction sites continue to validate without modification.
- `calibration_enabled: dict[Task, bool]` field on `AnnotationExportMeta` (defaulted empty). The validator only checks key membership when the dict is non-empty, so current sidecars remain valid.

### Paths

- New `AnnotationImportPaths` bundle and `resolve_import_paths(workspace, dataset_id)` in `core/paths/annotation_paths.py`. The bundle is keyed by `dataset_id` (persistent topology scope, with empty-string mapping to a `default` directory) because imports accumulate state across calls into the same scope. The bundle exposes a `partition.meta.json` sidecar path.

## Why

These schemas and paths model the calibration vs production assignment that downstream IAA metrics based on partial overlap will need. 

## Testing

- 634 unit tests pass (16 new for `PartitionManifest`, `PartitionManifestEntry`, and `AnnotationImportPaths`)
- `mypy src/pragmata` clean

### DIFF
--- a/src/pragmata/core/paths/annotation_paths.py
+++ b/src/pragmata/core/paths/annotation_paths.py
@@ -36,6 +36,52 @@ def resolve_annotation_paths(*, workspace: WorkspacePaths) -> AnnotationPaths:
 
 
 @dataclass(frozen=True, slots=True)
+class AnnotationImportPaths:
+    """Path bundle for an annotation import scope (one per ``dataset_id``).
+
+    Imports accumulate state across calls into the same logical scope, so the
+    bundle is keyed by ``dataset_id`` (persistent topology scope) rather than
+    a per-call identifier. The partition manifest sidecar tracks calibration
+    vs production assignment for every record imported into this scope.
+
+    Attributes:
+        tool_root: Annotation tool root.
+        import_dir: Per-``dataset_id`` import directory.
+        partition_manifest: Partition manifest sidecar path.
+    """
+
+    tool_root: Path
+    import_dir: Path
+    partition_manifest: Path
+
+    def ensure_dirs(self) -> Self:
+        """Create the import directory scaffold."""
+        self.import_dir.mkdir(parents=True, exist_ok=True)
+        return self
+
+
+def resolve_import_paths(*, workspace: WorkspacePaths, dataset_id: str) -> AnnotationImportPaths:
+    """Build the path bundle for an annotation import scope.
+
+    Args:
+        workspace: Workspace path bundle.
+        dataset_id: Topology scope. Empty string maps to the ``"default"``
+            directory so the layout is uniform across scopes.
+
+    Returns:
+        Path bundle for the import scope.
+    """
+    tool_root = workspace.tool_root("annotation")
+    scope = dataset_id or "default"
+    import_dir = tool_root / "imports" / scope
+    return AnnotationImportPaths(
+        tool_root=tool_root,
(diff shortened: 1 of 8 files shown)


src/pragmata/core/schemas/annotation_import.py
```diff
@@ -44,3 +46,45 @@ class QueryResponsePair(BaseModel):
...
+        calibration: True if assigned to the calibration dataset, else production.
+        import_id: Identifier of the import call that produced this assignment.
+        calibration_fraction_at_import: The fraction in force at that import call.
+        assigned_at: When the assignment was made.
+    """
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    calibration: bool
+    import_id: str
```
saschagobel: Small schema-contract thought: should `import_id` be a `NonEmptyStr` here?

src/pragmata/core/schemas/annotation_import.py
```diff
@@ -44,3 +46,45 @@ class QueryResponsePair(BaseModel):
...
+        partition_seed: Hash seed used for new-record assignments. Stored at
+            manifest level - changing seed mid-scope only affects new records.
+        assignments: Map of record_uuid -> PartitionManifestEntry.
+    """
+
+    model_config = ConfigDict(extra="forbid")
+
+    dataset_id: str
+    created_at: datetime
+    updated_at: datetime
```
saschagobel, line 89: Non-blocking: consider adding a small manifest-level validator for `updated_at >= created_at`. Not essential for this PR, but it would make the persisted sidecar contract slightly more self-defensive.

src/pragmata/core/paths/annotation_paths.py
```diff
@@ -35,6 +35,48 @@ def resolve_annotation_paths(*, workspace: WorkspacePaths) -> AnnotationPaths:
...
+    Args:
+        workspace: Workspace path bundle.
+        dataset_id: Topology scope. Empty string maps to the ``"default"``
+            directory so the layout is uniform across scopes.
+
+    Returns:
+        Path bundle for the import scope.
+    """
+    scope = dataset_id or "default"
+    import_dir = workspace.tool_root("annotation") / "imports" / scope
```
saschagobel, line 73: Small path-safety concern: `dataset_id` is used directly as a filesystem path segment here, but the current contract only types it as `str`. Since `dataset_id` now defines artifact topology, we should validate it at the shared settings/schema boundary as a safe path segment, e.g. allow empty string for the default scope and otherwise only slug-like IDs without `/`, `..`, or surrounding whitespace.

saschagobel (approved):
Approved with comments. The schema/path additions look directionally sound and give the calibration split a clear persisted contract; my notes are small contract-hardening points.

saschagobel (approved):

PR #175, opened 2026-04-30
docs: annotation stack design doc

henrycgbaker:
## Goal

Add design doc for annotation-specific infrastructure: Docker stack lifecycle, compose distribution, and operational concerns for a self-hosted Argilla annotation campaign. Separated from shared config/settings (PR #162).

## Scope

- `annotation-stack-lifecycle.md` (new) - stack composition, compose distribution (locked/package-data model), first-run/upgrade/uninstall lifecycle, data persistence boundaries, prod bootstrap including multi-host deployment, cross-platform runtime, and infra error taxonomy for `pragmata annotation`
- `annotation-interface.md` (updated) - align `init`→`setup` naming with this doc, fix stale `/apps/` path, cross-link to lifecycle doc
- `README.md` (design-doc index) - new entry

## Implementation

Design-only PR. No code changes.

Key decisions documented:
- **Locked compose model** (option Y): compose file ships as package data under `src/pragmata/annotation/`, resolved via `importlib.resources`. End users never edit it directly.
- **Single shipped compose artefact** (not dev/prod pair); contributor dev override layered via Makefile using `docker compose -f ... -f ...`.
- **Lifecycle CLI**: `pragmata annotation up` / `down` as the single end-user stack entry point (no separate `init`; `setup` reserved for Argilla provisioning).
- **Image pulls**: `docker compose up --pull missing` for end users; Makefile keeps `--pull always` for contributors. Asymmetry intentional.
- **Network exposure**: shipped compose binds `127.0.0.1` only. Multi-annotator deployments documented as reverse-proxy / Cloudflare Tunnel pattern (operator-owned, not in pragmata). No `--bind` flag in v0.1.
- **Data persistence**: named volumes survive container recreation, host reboot, and VM stop-deallocate (on default OS-disk Docker setup); ephemeral-disk pitfall flagged in README; no backup verb in v0.1.
- **Credentials policy**: `up` consumes static operator-set credentials from env, prints none. Print-on-creation reserved for dynamically generated secrets (see #200 for `setup`).
- **Test boundary**: integration tests run (shipped + dev-override); one separate packaging smoke test exercises shipped-only from an installed wheel (file resolves + YAML parses).
- **Prod bootstrap**: no shipped script for v0.1 - two-line install documented in README, works identically across pip/pipx/uv.

## References

- ADR-0007 (packaging & invocation surface)
- ADR-0003 (infra: self-hosted only)
- Companion PR #162 (config & settings)
- Follow-up #200 (annotation setup user provisioning)


docs/design/annotation-stack-lifecycle.md
```diff
@@ -0,0 +1,228 @@
...
+- **Z** is a clean future escape hatch. Document the pattern, but do **not** build the `eject` verb until concrete demand materialises.
+
+>Related rejected patterns: generate compose from a template at install time (two sources of truth, drifts on upgrade); remote URL fetch (breaks offline installs, trust boundary). Supabase [issue #2435](https://github.com/supabase/cli/issues/2435) documents the specific pain of user-editable compose + CLI tight version coupling - option Y avoids the problem entirely.
+
+### 2.2 Distribution mechanism (how pkg'd YAML travels)
+
+Ships as package data inside the installed package at `src/pragmata/annotation/docker-compose.yml`, resolved at runtime via `importlib.resources.files("pragmata.annotation") / "docker-compose.yml"` + `as_file()`. Same mechanism already in use for [`core/annotation/collapsible_field.html`](../../src/pragmata/core/annotation/collapsible_field.html).
+
+Image tags are pinned in the shipped compose and treated as package-owned. `pip install -U pragmata` ships a new compose with new tags - users automatically pick it up on next `up` (no drift, no warning needed under option Y).
```
saschagobel: This should also state the packaging contract explicitly: the compose file must be included in both wheel and sdist artifacts, and packaging tests should verify that it is resolvable via importlib.resources from an installed package. Otherwise this can work in-tree but fail after PyPI installation.

docs/design/annotation-stack-lifecycle.md
```diff
@@ -0,0 +1,228 @@
...
+### 3.1 First-run UX
+
+**`pragmata annotation up` is the first-run command. No separate `init`. `annotation setup` stays as Argilla provisioning (see [`config-and-settings.md`](config-and-settings.md) §3), invoked after the stack is up.**
+
+`up` does:
+- Pre-flight in order: extra installed? → Docker daemon reachable? → required ports free?
+- Resolve the packaged compose via `importlib.resources` (no copy to disk - option Y, §2.1)
+- Pulls images on first invocation (slow; log clearly - make this prominent in docs)
+- Health-polls Argilla's health endpoint with a timeout
+- On success: prints URL, default API key, print creds (once) and next command
```
saschagobel: I’d avoid printing credentials/API keys by default. If stdout is captured by shell logs, CI logs, terminal recording, or support output, a warning printed alongside the secret comes too late. annotation up should print only non-sensitive connection info by default.

This also seems inconsistent with the shared config/secrets policy: “Secrets are not stored in any pragmata-owned file; LLM provider keys go through env vars; Argilla delegates to its own credential store.” If Argilla owns credential storage, annotation up should not print API keys/passwords by default either.

docs/design/annotation-stack-lifecycle.md
```diff
@@ -0,0 +1,228 @@
...
+| **B. `scripts/bootstrap-annotation.sh` in repo** | Static shell file checked into git, linked from docs | [Docker convenience script](https://github.com/docker/docker-install), [rustup](https://rustup.rs/), [nvm](https://github.com/nvm-sh/nvm#installing-and-updating) | Maintenance drift - every CLI flag change invalidates it. Docker's own convenience installer is [not recommended for production](https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script) for exactly this reason | **Skip** |
+
+> **Proposed `annotation up` sequence (to implement):**
+> 1. Pre-flight checks: `pragmata[annotation]` extra present → Docker daemon reachable → required ports free. Fail fast with actionable message on any failure.
+> 2. Resolve compose file via `importlib.resources` (in-memory path, no copy to disk).
+> 3. `docker compose pull` - only on first run or after `pip install -U` (detect by comparing pinned image digests vs locally cached). Log clearly; this is the slow step.
+> 4. `docker compose up -d` with resolved profile (§2.3).
+> 5. Poll `GET /api/v1/health` until 200 or timeout (default 120 s). Stream a progress indicator.
```
saschagobel: Is /api/v1/health compatible with the Argilla v2 server image we target via argilla>=2.0,<3.0?

saschagobel (approved):
Thanks, approved, I left a few non-blocking comments around making the package-data contract explicit, avoiding default credential/API-key display in annotation up, and confirming the Argilla v2 health endpoint before hard-coding it.

docs/design/annotation-stack-lifecycle.md
henrycgbaker: edit - coming back to this later

henrycgbaker (commented):
> I’d avoid printing credentials/API keys by default. If stdout is captured by shell logs, CI logs, terminal recording, or support output, a warning printed alongside the secret comes too late. annotation up should print only non-sensitive connection info by default.
> 
> This also seems inconsistent with the shared config/secrets policy: “Secrets are not stored in any pragmata-owned file; LLM provider keys go through env vars; Argilla delegates to its own credential store.” If Argilla owns credential storage, annotation up should not print API keys/passwords by default either. 

Agreed on not printing here in this case, but the reasoning matters i think because we have a documented exception (#200) that this rule would otherwise block.

Two cases, two rules:

1. `up` consumes static operator-set credentials (ARGILLA_USERNAME / _PASSWORD / _API_KEY in their env). The operator already knows these -> echoing them back adds zero info and only adds leakage surface mentionedd. So up prints URL + next-step hint only.

2. `setup` (#200) generates per-annotator pwds. Argilla hashes them at rest, so they're unrecoverable after creation -> print-once is the only way ops can get them. As these are scoped to within-app credentials only it's not same as leaking a genuine env var

docs/design/annotation-stack-design.md
```diff
@@ -0,0 +1,404 @@
...
+include = [
+  "src/pragmata/**/*.py",
+  "src/pragmata/**/*.yml",
+  "src/pragmata/**/*.html",
+  "pyproject.toml",
+  "README.md",
+  "LICENSE.md",
+]
+```
+
```
saschagobel: I agree with making the wheel package root explicit via `packages = ["src/pragmata"]`. I also buy the fail-fast rationale for the shipped compose file: it is a critical runtime artifact and the annotation stack contract depends on it being present in the installed distribution.

I would narrow this, though. Since `collapsible_field.html` lives under `src/pragmata`, it should already be covered by package-tree inclusion. I would avoid listing every non-Python package resource under `force-include`, otherwise `pyproject.toml` becomes a second manifest of internal resource paths. I’d reserve `force-include` for the compose file because it is the locked shipped infrastructure artifact.

I would also drop the explicit `sdist.include` whitelist. Hatch’s sdist default already includes all files not ignored by VCS, with `pyproject.toml`, the readme, and license files always included.

Suggested compromise:

```toml
[tool.hatch.build.targets.wheel]
packages = ["src/pragmata"]

[tool.hatch.build.targets.wheel.force-include]
"src/pragmata/annotation/docker-compose.yml" = "pragmata/annotation/docker-compose.yml"
```

The installed-artifact packaging smoke test should be the actual regression guard here. It should build/install the package artifact, resolve pragmata/annotation/docker-compose.yml via importlib.resources, parse it as YAML, and also resolve any runtime templates the annotation code loads, including collapsible_field.html. That gives us a stronger end-to-end packaging contract without duplicating every package resource path in pyproject.toml.

saschagobel (approved):
Great write-up, thanks. Left one non-blocking comment regarding the packaging contract.

PR #170, opened 2026-04-29
feat(annotation): partial-overlap calibration / production split for IAA

henrycgbaker:
## Goal

Support partial-overlap annotation: a small calibration subset annotated by all annotators (for inter-annotator agreement) while the bulk runs single-annotated. 

## Problem

Argilla v2's `TaskDistribution(min_submitted=N)` is dataset-wide and frozen once annotation begins ([docs](https://docs.argilla.io/latest/how_to_guides/distribution/), verified in `argilla-server/src/argilla_server/contexts/distribution.py`). There is no per-record overlap, no metadata-based completion threshold, etc. Argilla's own tracker confirms there is no
clean native fix:
- [argilla#3361](https://github.com/argilla-io/argilla/issues/3361) closed in favour of #4067
- [argilla#4067](https://github.com/argilla-io/argilla/issues/4067) is just a docs request to use metadata-based assignment
- [argilla#4792](https://github.com/argilla-io/argilla/pull/4792) closed unmerged; maintainer confirmed full overlap is just default behaviour

Metadata-based assignment only controls UI queue visibility, not completion thresholds, so it cannot solve mixed overlap within one dataset. The only viable workaround is **two datasets per task**: ``task_<task>_calibration``
(``min_submitted=N``) alongside ``task_<task>_production``(``min_submitted=1``).

## Use cases

The design supports these workflows naturally:

1. **Standard pilot (default)**: First import with default `calibration_fraction=0.1`. ~10% of records routed to calibration (`min_submitted=3`), 90% to production (`min_submitted=1`). Run IAA on calibration. Refine guidelines if needed.
2. **Production-only first batch**: First import with `calibration_fraction=0.0`. Only the production dataset is created in Argilla (lazy creation). Calibration dataset does not exist yet but topology supports it.
3. **Add calibration later**: A later import with `calibration_fraction=0.1` on a fresh batch creates the calibration dataset on the fly. Earlier production-only records stay where they are (manifest locks assignments).
4. **Re-calibrate after drift suspicion**: Run a fresh batch with `calibration_fraction=0.1`. New records partition normally; existing manifest entries preserved.
5. **Disable calibration entirely** (rare, opt-out): Set `calibration_min_submitted=None` in topology. All imports go production-only; `calibration_fraction > 0` errors hard at import time.

### Settings shape

[...]

PR #177, opened 2026-04-30
feat(annotation): partition records into calibration vs production for IAA

henrycgbaker:
## Goal

Add the partition + import-API layer of the calibration/production overlap feature: route a configurable fraction of records to a separate calibration dataset for IAA, while the bulk runs single-annotated in production.

Stacked on #173.

### Partition logic (`core/annotation/record_builder.py`)
- `_bucket_calibration(record_uuid, fraction, seed)` — deterministic per-record assignment via SHA-256 hash; `fraction=0.0` short-circuits to all-production, `fraction=1.0` to all-calibration.
- `assign_partitions()` — resolves each pair's calibration assignment, mutating the manifest in place. Existing manifest entries lock prior assignments (re-import safety); new records bucket via the manifest's stored seed.
- `load_partition_manifest()` / `write_partition_manifest()` — atomic IO (write-tmp-then-rename).
- `_build_batches()` — extends to key by `(task, calibration)` so fan-out routes to the right purpose dataset.
- `_ensure_dataset()` — extracted helper; resolves or creates the Argilla dataset for a `(task, purpose)` pair with the right `min_submitted` from the topology.
- `fan_out_records()` returns `(dataset_counts, calibration_count, production_count)`.

### Teardown (`core/annotation/setup.py`)
- `teardown_resources()` deletes both purpose datasets per topology, conditional on `calibration_min_submitted is not None`.

### Import API (`api/annotation_import.py`)
- New `calibration_fraction: float = 0.1` kwarg on `import_records()`.
- Hard error if `calibration_fraction > 0` but topology disables calibration for any task receiving records — silent rerouting hides user mistakes.
- Manifest IO with import-time provenance (`import_id`, `calibration_fraction_at_import`, `assigned_at`).
- Stored partition seed is logged-and-respected if it differs from current settings (existing assignments preserved).
- `ImportResult` extended with `calibration_count`, `production_count`, effective `calibration_fraction`.

### CLI (`cli/commands/annotation.py`)
- `--calibration-fraction` flag on `annotation import`.

## Implementation

Files modified:
- [src/pragmata/core/annotation/record_builder.py](src/pragmata/core/annotation/record_builder.py) — partition logic, manifest IO, per-purpose fan-out
- [src/pragmata/core/annotation/setup.py](src/pragmata/core/annotation/setup.py) — purpose-aware teardown
- [src/pragmata/api/annotation_import.py](src/pragmata/api/annotation_import.py) — `calibration_fraction` kwarg, manifest persistence, hard-error validation, extended `ImportResult`
- [src/pragmata/api/annotation_setup.py](src/pragmata/api/annotation_setup.py) — drop `min_submitted` kwarg
- [src/pragmata/cli/commands/annotation.py](src/pragmata/cli/commands/annotation.py) — `--calibration-fraction` CLI flag

## Testing

`uv run pytest -q -m "not integration"`: 673 passed.
`uv run mypy src/pragmata` clean. `ruff check` and `ruff format --check` clean.

New unit tests:
- [tests/unit/core/annotation/test_partition.py](tests/unit/core/annotation/test_partition.py) — `_bucket_calibration` determinism + bounds, `assign_partitions` honours existing manifest, manifest IO round-trip + atomic write
- [tests/unit/api/test_annotation.py](tests/unit/api/test_annotation.py) `TestImportRecordsCalibrationValidation` — fraction range validation, hard-error when topology disables calibration

New integration test:
- [tests/integration/test_annotation_calibration.py](tests/integration/test_annotation_calibration.py) — Argilla-backed E2E: default fraction, fraction=0 lazy skip, re-import locking, growing batches, dataset placement

Existing tests updated for the new `fan_out_records` 3-tuple shape and per-purpose dataset names.

## References

- Closes the partition + import-API portion of #170 (closed in favour of this 5-PR stack)
- Design docs: `docs/design/annotation-import-pipeline.md`
- ADR-0010 (one dataset per task, linked by `record_uuid`) — calibration dataset is a sibling of production, same shape


src/pragmata/api/annotation_import.py
```diff
@@ -45,18 +58,25 @@ def import_records(
     dataset_id: str | Unset = UNSET,
     base_dir: str | Path | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
+    calibration_fraction: float = 0.1,
```
saschagobel: `calibration_fraction` should follow the established settings wiring pattern: the default should live on AnnotationSettings, range check validation should be expressed on the Pydantic settings model, and the API kwarg should be float | Unset = UNSET threaded through as an override.

src/pragmata/api/annotation_import.py
```diff
@@ -76,32 +96,99 @@ def import_records(
...
+            > 0 when topology declares no calibration dataset for any task
+            receiving records.
     """
+    if not 0.0 <= calibration_fraction <= 1.0:
+        raise ValueError(f"calibration_fraction must be in [0.0, 1.0]; got {calibration_fraction}")
+
     raw = resolve_records(records, format=format)
     settings = AnnotationSettings.resolve(
-        config=load_config_file(config_path) if isinstance(config_path, (str, Path)) else None,
+        config=load_config_file(config_path) if isinstance(config_path, str | Path) else None,
```
saschagobel: This is equivalent on modern Python, but the codebase consistently uses isinstance(config_path, (str, Path)); I’d keep the existing tuple form here for consistency and readability.

src/pragmata/api/annotation_import.py
```diff
@@ -76,32 +96,99 @@ def import_records(
...
+        ImportResult with totals, per-dataset counts, partition counts, and
+        validation errors.
+
+    Raises:
+        ValueError: ``calibration_fraction`` is outside [0.0, 1.0], or is
+            > 0 when topology declares no calibration dataset for any task
+            receiving records.
     """
+    if not 0.0 <= calibration_fraction <= 1.0:
+        raise ValueError(f"calibration_fraction must be in [0.0, 1.0]; got {calibration_fraction}")
```
saschagobel: See comment on line R61

src/pragmata/api/annotation_import.py
```diff
@@ -76,32 +96,99 @@ def import_records(
...
+            for task_overlaps in settings.workspace_dataset_map.values()
+            for task, overlap in task_overlaps.items()
+            if overlap.calibration_min_submitted is None
+        ]
+        if missing:
+            raise ValueError(
+                f"calibration_fraction={calibration_fraction} > 0 but topology has no "
+                f"calibration dataset for tasks: {sorted(set(missing))}. Either set "
+                f"calibration_fraction=0.0 or enable calibration in workspace_dataset_map."
+            )
```
saschagobel: This topology/fraction compatibility check is an import invariant rather than API orchestration. I’d move it into a small core validation helper and call it from here, especially once calibration_fraction is resolved through AnnotationSettings.

src/pragmata/api/annotation_import.py
```diff
@@ -76,32 +96,99 @@ def import_records(
...
+            manifest=manifest,
+            fraction=calibration_fraction,
+            import_id=import_id,
+        )
+        write_partition_manifest(import_paths.partition_manifest, manifest)
+
+        calibration_count = sum(1 for is_cal in assignments.values() if is_cal)
+        production_count = len(assignments) - calibration_count
+
+        dataset_counts = fan_out_records(client, validation.valid, settings, assignments=assignments)
```
saschagobel: The manifest is persisted before `fan_out_records()` runs. If Argilla submission fails hard after this point, assignments are locked for records that were never imported, right? Should we prefer writing the manifest after successful fan-out?

src/pragmata/api/annotation_setup.py
```diff
@@ -31,8 +31,8 @@ def setup(
 
     Datasets are not created here — they are auto-created on import,
     scoped by dataset_id. Per-task overlap (production and calibration
-    ``min_submitted``) is configured via ``workspace_dataset_map`` in the
-    YAML config or ``AnnotationSettings``.
+    ``min_submitted``) is declared in
+    ``AnnotationSettings.workspace_dataset_map``.
```
saschagobel: This docstring should not point users at AnnotationSettings, which is internal. For the setup API surface, please phrase this in terms of the supported config path/YAML configuration.

src/pragmata/cli/commands/annotation.py
```diff
@@ -35,7 +35,7 @@ def setup_command(
 
     Datasets are created automatically on import, not here. Per-task overlap
     (production and calibration ``min_submitted``) is configured via
-    ``workspace_dataset_map`` in the YAML config.
+    ``workspace_dataset_map`` in the YAML config or ``AnnotationSettings``.
```
saschagobel: Same as in the API docstring: the CLI help/docstring should not point users at internal AnnotationSettings. Please keep this framed as supported user configuration.

src/pragmata/cli/commands/annotation.py
```diff
@@ -86,10 +86,18 @@ def import_command(
     format: str | None = typer.Option(
         None, "--format", help="File format override (json, jsonl, csv). Auto-detected by default."
     ),
+    calibration_fraction: float = typer.Option(
+        0.1,
+        "--calibration-fraction",
+        help="Fraction of records routed to the calibration dataset for this batch. "
+        "Set to 0.0 for production-only batches.",
+    ),
```
saschagobel: --calibration-fraction should follow the established UNSET/settings pattern. The default should live in the resolved annotation settings, not in the Typer option; the CLI option should default to None/omitted and pass UNSET when absent, matching the other config-backed options.

src/pragmata/cli/commands/annotation.py
```diff
@@ -101,8 +109,13 @@ def import_command(
         dataset_id=UNSET if dataset_id is None else dataset_id,
         base_dir=UNSET if base_dir is None else base_dir,
         config_path=UNSET if config is None else config,
+        calibration_fraction=calibration_fraction,
```
saschagobel: See comment on lines R89 to R94

src/pragmata/cli/commands/annotation.py
```diff
@@ -101,8 +109,13 @@ def import_command(
...
     )
     typer.echo(f"Total records: {result.total_records}")
+    typer.echo(
+        f"Calibration: {result.calibration_count}, "
+        f"production: {result.production_count} (fraction={result.calibration_fraction:.3f})"
```
saschagobel: With manifest locking, is result.calibration_fraction intended to mean the configured fraction used for new assignments only, rather than the realized calibration share in this import? On re-import, existing assignments are preserved, so the printed fraction=... could differ substantially from calibration_count / (calibration_count + production_count).

tests/integration/test_annotation_calibration.py
```diff
@@ -0,0 +1,198 @@
...
+    """Per-test workspace so each test owns its partition manifest."""
+    return tmp_path
+
+
+# ---------------------------------------------------------------------------
+# Tests
+# ---------------------------------------------------------------------------
+
+
+def test_default_calibration_fraction_creates_both_datasets(client: rg.Argilla, base_dir: Path) -> None:
```
saschagobel: test_default_calibration_fraction_creates_both_datasets() does not actually test the default: it explicitly passes calibration_fraction=0.5, while the name/docstring says default 0.1. Once calibration_fraction is moved into settings, these integration tests should exercise both paths: configured/default fraction via settings/config, and explicit CLI/API override. Right now they only exercise direct API kwarg overrides.

tests/integration/test_annotation_calibration.py
saschagobel, line 1: Several tests hardcode base_dir / "annotation" / "imports" / auto_id / "partition.meta.json". Since this path is owned by the path resolver, prefer using resolve_import_paths(...).partition_manifest in tests too; otherwise the test duplicates internal path layout and can drift from the canonical resolver.

tests/unit/core/annotation/test_partition.py
saschagobel, line 1: A couple of tests seem probabilistic: test_different_seed_different_bucket() and especially test_fraction_bounded_for_uniform_uuids(). They are very likely to pass, but still encode statistical expectations in a unit test. Since _bucket_calibration() is deterministic for a fixed seed/input set, should we prefer asserting exact expected outputs or an exact expected count for the fixed fixture set?

saschagobel (changes requested):
Requesting changes mainly to keep the calibration fraction on the established settings/config path.

Also, I don’t yet see focused unit coverage for the changed fan_out_records() contract since the API tests patch it out.

saschagobel (approved):
This looks good now, thanks.

PR #178, opened 2026-04-30
feat(annotation): tag exported rows with calibration and filter IAA inputs

henrycgbaker:
## Goal

Final part of the calibration/production overlap feature: surface partition information through the export and IAA pipelines so downstream consumers can distinguish calibration rows from production rows.

Stacked on #177.

### Export (`core/annotation/export_fetcher.py`)
- `fetch_task()` iterates the production dataset (always present) and the calibration dataset (when topology declares one for the task). Each row is tagged with `calibration: bool` so downstream consumers can filter.
- Calibration dataset may be missing when no records were ever routed to it (lazy creation); falls through silently to match the existing "skip empty CSV" pattern.

### Export sidecar (`core/annotation/export_runner.py`)
- `AnnotationExportMeta.calibration_enabled: dict[Task, bool]` is populated from the topology so consumers know which tasks have a calibration dataset.

### IAA (`core/annotation/iaa_runner.py`)
- `compute_iaa()` filters CSV rows to `calibration=True` by default — IAA is only meaningful on overlapped records. No new kwarg added; the column-based filter is the entire mechanism.
- Empty-result case returns an empty `IaaReport` with a warning, matching the existing "skip empty CSV" pattern.

## Implementation

Files modified:
- [src/pragmata/core/annotation/export_fetcher.py](src/pragmata/core/annotation/export_fetcher.py) — iterate both datasets, tag rows
- [src/pragmata/core/annotation/export_runner.py](src/pragmata/core/annotation/export_runner.py) — sidecar carries `calibration_enabled`
- [src/pragmata/core/annotation/iaa_runner.py](src/pragmata/core/annotation/iaa_runner.py) — filter rows by `calibration` column

## Testing

`uv run pytest -q -m "not integration"`: 673 passed.
`uv run mypy src/pragmata` clean. `ruff check` and `ruff format --check` clean.

Updated tests:
- [tests/unit/api/test_export_api.py](tests/unit/api/test_export_api.py) — both purpose datasets fetched per task
- [tests/unit/core/annotation/test_export_fetcher.py](tests/unit/core/annotation/test_export_fetcher.py) — `calibration` tagged on rows
- [tests/unit/core/annotation/test_export_runner.py](tests/unit/core/annotation/test_export_runner.py) — sidecar carries `calibration_enabled`
- [tests/unit/core/annotation/test_iaa_runner.py](tests/unit/core/annotation/test_iaa_runner.py) — filter to `calibration=True`
- [tests/unit/core/schemas/test_annotation_export.py](tests/unit/core/schemas/test_annotation_export.py) — `calibration` field validation

## References

- Closes the export + IAA portion of #170 (closed in favour of this 5-PR stack)
- Design docs: `docs/design/annotation-export-pipeline.md`
- ADR-0010 (one dataset per task, linked by `record_uuid`) — calibration vs production rows distinguished by column, single CSV per task

### DIFF
--- a/src/pragmata/core/annotation/export_runner.py
+++ b/src/pragmata/core/annotation/export_runner.py
@@ -109,6 +109,18 @@ def write_export_csv(
         raise
 
 
+def _resolve_calibration_enabled(settings: "AnnotationSettings", tasks: list[Task]) -> dict[Task, bool]:
+    """Per-task calibration topology lookup from settings."""
+    flags: dict[Task, bool] = {}
+    for task_overlaps in settings.workspace_dataset_map.values():
+        for task, overlap in task_overlaps.items():
+            if task in tasks:
+                flags[task] = overlap.calibration_min_submitted is not None
+    for task in tasks:
+        flags.setdefault(task, False)
+    return flags
+
+
 def assemble_export_meta(
     export_id: str,
     dataset_id: str | None,
@@ -116,6 +128,7 @@ def assemble_export_meta(
     include_discarded: bool,
     row_counts: dict[Task, int],
     n_annotators: dict[Task, int],
+    calibration_enabled: dict[Task, bool],
     constraint_summary: dict[str, int],
 ) -> AnnotationExportMeta:
     """Assemble run-level provenance metadata for an annotation export.
@@ -127,6 +140,8 @@ def assemble_export_meta(
         include_discarded: Whether discarded responses were included.
         row_counts: Rows written per task.
         n_annotators: Distinct annotator count per task.
+        calibration_enabled: Whether the topology declared a calibration
+            dataset for each task.
         constraint_summary: Constraint violation count per rule name.
 
     Returns:
@@ -140,6 +155,7 @@ def assemble_export_meta(
         include_discarded=include_discarded,
         row_counts=row_counts,
         n_annotators=n_annotators,
+        calibration_enabled=calibration_enabled,
         constraint_summary=constraint_summary,
     )
 
@@ -171,6 +187,7 @@ def run_export(
             include_discarded=include_discarded,
             row_counts={},
             n_annotators={},
+            calibration_enabled={},
             constraint_summary={},
         )
         write_export_meta(meta, paths.export_meta_json)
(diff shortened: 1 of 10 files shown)


src/pragmata/core/annotation/export_runner.py
```diff
@@ -109,13 +109,35 @@ def write_export_csv(
...
+
+    Raises if any requested task is absent from the topology, mirroring the
+    import-side validation. Silently defaulting missing tasks to False would
+    hide a config bug.
+    """
+    flags: dict[Task, bool] = {}
+    for task_overlaps in settings.workspace_dataset_map.values():
+        for task, overlap in task_overlaps.items():
+            if task in tasks:
+                flags[task] = overlap.calibration_min_submitted is not None
```
saschagobel, line 123: Small consistency question: can the same Task appear under multiple workspace_dataset_map entries? fetch_task() resolves the workspace/topology with first-match-wins semantics, while this helper keeps iterating and would let the last matching task determine calibration_enabled. If duplicate task placement is already rejected elsewhere, no issue; otherwise this should probably either mirror fetch_task()’s first-match behavior or, preferably, the settings schema should validate task uniqueness across workspaces.

tests/unit/core/annotation/test_export_fetcher.py
```diff
@@ -15,15 +15,25 @@
...
 # ---------------------------------------------------------------------------
 
 _UID1 = UUID("00000000-0000-0000-0000-000000000001")
 _UID2 = UUID("00000000-0000-0000-0000-000000000002")
-_SETTINGS = AnnotationSettings()
+
+# Tests below use a topology with calibration disabled so that fetch_task
+# iterates a single dataset per task. Calibration-aware fetching is
+# exercised in the integration suite.
+_SETTINGS = AnnotationSettings(
```
saschagobel: Test coverage thought: since fetch_task() now owns the production/calibration iteration and the calibration row tag, should we not have one focused mocked unit test here for “calibration-enabled topology fetches both datasets and tags rows correctly.”, relying on integration coverage for the persisted export contract.

saschagobel (approved):
This looks solid, thanks. Left one consistency and one test coverage question.

src/pragmata/core/annotation/export_runner.py
henrycgbaker: Added a model_validator to AnnotationSettings that rejects duplicate task entries at construction time (_resolve_calibration_enabled, fetch_task, and _invert_workspace_map in record_builder all had different behaviour so all are resolved by this)

PR #183, opened 2026-05-05
fix(deps): pinned to python=3.12

henrycgbaker:
Before Python >=3.12. This led to Pydantic v1 clash UserWarning when freshly installing python on new maching (VM)


henrycgbaker:
Branched off the wrong base — accumulated unrelated commits from #178's stack. Replaced by #184 (clean one-line change off main).

PR #184, opened 2026-05-05
fix(deps): raise Pydantic lower bound to >=2.13

henrycgbaker:
## Goal

Fix Python 3.14 compatibility. LangChain imports from \`pydantic.v1\` for backward compatibility; this is addressed in Pydantic 2.13 ([langchain-ai/langchain#33926](https://github.com/langchain-ai/langchain/issues/33926)). Raising the Pydantic lower bound removes the warning without pinning the repo to a specific Python version.

## Scope

- Bump \`pydantic>=2.0,<3.0\` → \`pydantic>=2.13,<3.0\` in \`pyproject.toml\`
- Update \`uv.lock\` (pydantic 2.12.5 → 2.13.3, pydantic-core bumped accordingly)
- Drop \`.python-version\` pin (now unnecessary — root cause fixed upstream)

## Implementation

Pydantic 2.13 resolves the \`pydantic.v1.fields.FieldInfo\` import that LangChain uses internally, which was incompatible with Python 3.14. Bumping the floor ensures all environments get a version where this is fixed.

## Testing

Tested under Python 3.14 with the Pydantic warning promoted to an error — no longer raised with \`pydantic>=2.13\`.

## References

- [langchain-ai/langchain#33926](https://github.com/langchain-ai/langchain/issues/33926)
- Discovered while deploying pragmata on the BSt VM (Python 3.14 resolved by uv on a clean machine)


saschagobel (changes requested):
Thanks for catching this — the underlying issue is real, but I’d prefer not to solve it by pinning the repo Python version. The Python 3.14 symptom is a Pydantic warning rather than a hard failure: LangChain imports `pydantic.v1` for backward compatibility (`langchain_core._api.deprecation -> pydantic.v1.fields.FieldInfo`; see https://github.com/langchain-ai/langchain/issues/33926), and the actual compatibility issue is addressed in Pydantic 2.13.

I just tested this under Python 3.14 with the warning promoted to an error, and raising our lower bound to `pydantic>=2.13,<3.0` fixes the LangChain import surface without a `.python-version` pin. So I’d suggest changing this PR to bump the Pydantic lower bound and update `uv.lock`.

saschagobel (approved):

PR #190, opened 2026-05-11
feat: split querygen model kwargs by planning and realization stage

saschagobel:
**Summary**

Replaces the shared querygen `model_kwargs` setting with stage-specific `planning_model_kwargs` and `realization_model_kwargs`. The change allows callers to pass provider-specific model options independently to the planning and realization models.

**Key changes**

- Replace `model_kwargs` with:
  - `planning_model_kwargs`
  - `realization_model_kwargs`
- Wire planning-stage execution to `planning_model_kwargs`
- Wire planning-summary execution to `planning_model_kwargs`
- Wire realization-stage execution to `realization_model_kwargs`
- Update the Python API and CLI surfaces
- Update unit and integration tests for the new settings/API/CLI contract

**Rationale**

The previous shared `model_kwargs` setting applied the same provider-specific options to both querygen stages. That became too coarse once planning and realization started using different model roles and, in practice, potentially different model families or reasoning settings.

Splitting this into `planning_model_kwargs` and `realization_model_kwargs` keeps provider-specific tuning stage-local: planning can receive options appropriate for structured blueprint generation and planning-summary updates, while realization can receive options appropriate for natural-language query writing.

**Compatibility**

This is a breaking pre-release API/CLI change:

- `model_kwargs` is removed
- `--model-kwargs` is removed
- config files must use `planning_model_kwargs` and/or `realization_model_kwargs` under `llm`

### DIFF
--- a/src/pragmata/api/querygen.py
+++ b/src/pragmata/api/querygen.py
@@ -67,7 +67,8 @@ def gen_queries(
     check_every_n_seconds: float | Unset = UNSET,
     max_bucket_size: int | Unset = UNSET,
     base_url: str | Unset = UNSET,
-    model_kwargs: dict[str, Any] | Unset = UNSET,
+    planning_model_kwargs: dict[str, Any] | Unset = UNSET,
+    realization_model_kwargs: dict[str, Any] | Unset = UNSET,
     batch_size: PositiveInt | Unset = UNSET,
     near_duplicate_tolerance: float | Unset = UNSET,
     enable_planning_memory: bool | Unset = UNSET,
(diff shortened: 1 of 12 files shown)


lramir14 (approved):
Thanks for your work.

The change looks clean and well-scoped. The splitting was implemented and fits the new querygen design. 

My only non-blocking comment would be perhaps documenting the change and why we decided to do so.

saschagobel:
> Thanks for your work.
> 
> The change looks clean and well-scoped. The splitting was implemented and fits the new querygen design.
> 
> My only non-blocking comment would be perhaps documenting the change and why we decided to do so.

Good point. I added a short rationale section in the PR body explaining that the shared `model_kwargs` became too coarse because planning/planning-summary and realization may need different provider-specific options, especially for reasoning/settings differences across model roles.

PR #191, opened 2026-05-11
fix: enforce batch-specific querygen structured output cardinality

saschagobel:
**Summary**

Tightens the structured-output contract for querygen planning and realization. Previously, `QueryBlueprintList` and `RealizedQueryList` validated the shape of returned items, but not the expected number of returned items. This allowed incomplete LLM responses to remain structurally valid and continue through the workflow instead of failing validation and triggering bounded retries.

**Key changes**

- Add a dynamic `QueryBlueprintList` schema factory constrained to the planning batch size
- Add a dynamic `RealizedQueryList` schema factory constrained to the realization batch size
- Wire planning-stage execution to use `len(batch_candidate_ids)` as the expected blueprint count
- Wire realization-stage execution to use `len(candidates)` as the expected realized-query count
- Preserve the existing item schemas and downstream `list[QueryBlueprint]` / `list[RealizedQuery]` contracts

The item schemas remain unchanged, and downstream code still receives plain `list[QueryBlueprint]` and `list[RealizedQuery]`.

## Testing

- Added focused schema tests for dynamic planning and realization output cardinality.
- Updated planning and realization stage tests to assert dynamic schema subclasses are wired into the LLM runnable.
- Unit and integration tests pass.

### DIFF
--- a/tests/unit/core/schemas/test_querygen_plan.py
+++ b/tests/unit/core/schemas/test_querygen_plan.py
@@ -3,7 +3,7 @@
 import pytest
 from pydantic import ValidationError
 
-from pragmata.core.schemas.querygen_plan import QueryBlueprint, QueryBlueprintList
+from pragmata.core.schemas.querygen_plan import QueryBlueprint, QueryBlueprintList, make_query_blueprint_list_schema
 
 
 @pytest.fixture()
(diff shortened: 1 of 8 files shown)


tests/unit/core/schemas/test_querygen_plan.py
```diff
@@ -113,6 +113,22 @@ def test_query_blueprint_list_accepts_valid_candidates(
...
+    schema = make_query_blueprint_list_schema(expected_length=2)
+    second_payload = {**base_payload, "candidate_id": "candidate-002"}
+
+    result = schema.model_validate({"candidates": [base_payload, second_payload]})
+
+    assert isinstance(result, QueryBlueprintList)
+    assert len(result.candidates) == 2
+
+    with pytest.raises(ValidationError):
+        schema.model_validate({"candidates": [base_payload]})
```
lramir14, line 130: My only concern here is if we are overlooking cases where there are more candidates than requested. We are only considering a less candidate case. (So far, OpenAI has done this).

tests/unit/core/schemas/test_querygen_realize.py
```diff
@@ -57,6 +57,21 @@ def test_realized_query_list_accepts_valid_payload(
...
+    schema = make_realized_query_list_schema(expected_length=2)
+
+    result = schema.model_validate(valid_realized_query_list_payload)
+
+    assert isinstance(result, RealizedQueryList)
+    assert len(result.queries) == 2
+
+    with pytest.raises(ValidationError):
+        schema.model_validate({"queries": valid_realized_query_list_payload["queries"][:1]})
+
```
lramir14, line 77: And it is the same concern here, as in the planning section. We are only evaluating when we have less candidate queries.

lramir14 (approved):
Thanks for the work, looks good to me. 

I only have the non-blocking suggestion I placed in the commented section when testing planning and realization. The new dynamic schema tests verify the valid case and the “too few items” case. Since the schema is intended to enforce exact cardinality, I think it would be useful to also add a “too many items” case for both planning and realization, e.g. three items when `expected_length=2`.

No blocker from my side.

PR #192, opened 2026-05-11
chore: tighten packaging and tooling configuration

saschagobel:
**Summary**

Tightens the project packaging and tooling configuration while preserving the existing extras-based developer workflow. The change adds clearer dependency bounds, scopes mypy to the package source tree, and relaxes Ruff docstring checks for test class initializers.

**Key changes**

- Add upper bounds for runtime dependencies:
- Add clearer bounds for optional extras:
- Keep `test`, `lint`, `type`, `annotation`, `querygen`, and `dev` as package extras to preserve the existing contributor workflow
- Keep `argilla` in the `test` extra to avoid breaking annotation-related tests
- Add `D107` to Ruff per-file ignores for tests
- Scope mypy to `src/pragmata`
- Refresh `uv.lock`


henrycgbaker (approved):
Approved with non-blocking (v minor) comment:

Pydantic floor from >=2.0 to >=2.13;not in description, as this bump from langchain 1.2 resolution in lock I think. Worth mentioning in commit?

saschagobel:
> Approved with non-blocking (v minor) comment:
> 
> Pydantic floor from >=2.0 to >=2.13;not in description, as this bump from langchain 1.2 resolution in lock I think. Worth mentioning in commit?

Thanks — I think this is already covered by the uv.lock refresh note in the PR body. The declared Pydantic floor was not changed in this PR; it was already >=2.13,<3.0 from #184. If the resolved Pydantic version changed in uv.lock, that is just part of the lockfile refresh here, not a new dependency-policy change.

PR #196, opened 2026-05-18
feat(annotation): open Locale type and YAML i18n catalogs for Argilla datasets

henrycgbaker:
**Stack (5 PRs):** #206 → #196 → #207 → #208 → #209

**Chain goal:** Adds locale-aware Argilla datasets (EN/DE) on top of a deployment/workspace/task settings inheritance hierarchy, with live widget locale switching and CLI flags for per-import overrides.

**This PR (2/5):** `Locale` type + EN catalog + Argilla wiring foundation.

---

## Goal

Extract Argilla dataset display strings into a per-locale catalog so future locales can be added without touching `argilla_task_definitions` or any other Python source. Establish the inheritance plumbing so workspaces and tasks can declare a locale override that the parent PR's `resolved_task()` accessor picks up automatically.

## Scope

- **`Locale` is an open `str` type alias** (`core/schemas/annotation_task.py`), not a closed `StrEnum`. Locale codes (`"en"`, `"de"`, `"EN"`, …) are validated at catalog lookup, not at the schema layer. A deployment adds a locale with zero Python edits.
- `locale` field added to `TaskSettings`, `WorkspaceSettings`, `AnnotationSettings`; `_INHERITED_FIELDS` extends to include `"locale"` so the parent PR's resolver walks it automatically.
- `ResolvedTaskSettings` (the frozen-dataclass accessor return type introduced in #206) extends to include `locale: Locale`. Consumers can read `settings.resolved_task(ws, task).locale` and get a concrete locale code.
- **Catalogs ship as YAML data**, not Python code. Translations in `core/annotation/locales/en.yaml`; the registry auto-discovers `*.yaml` files at import time, keyed by filename stem. Adding a locale = drop a YAML file.
- `loader.py` owns the locale-invariant *structure* (`_YES_NO_QUESTIONS_BY_TASK`) and fans YAML inputs over it into the flat tuple-keyed Catalog shape consumers expect. Reuses the project's `load_config_file` helper for YAML reading. `DiscardReason` is iterated against the YAML's `discard_reasons:` map so a missing or typo'd reason fails loud at load.
- `build_task_settings(locale)` parameterised on locale; field/question/guidelines titles and `LabelQuestion(labels=...)` dicts driven by the catalog. Identities (`name=`) and label values (`yes`/`no`/`DiscardReason.*.value`) remain stable across locales so exports merge cleanly across multi-language deployments.
- `Locale` re-exported via the `annotation` facade (`pragmata.annotation.Locale`).

## Why open string instead of closed enum

Closed `StrEnum` baked "which languages we support" into the schema layer. A downstream deployment that needs `EN` would have to fork pragmata to add the enum member. Matches how SOTA i18n tools (gettext, i18next, Rails) treat locale codes open BCP 47 strings, not closed enums.

Trade-off: loses compile-time autocomplete on `Locale.X` - typos surface at catalog lookup (`KeyError`) rather than type-check time. Acceptable given the win in extensibility.

## Why YAML instead of Python catalogs

Translations are pure strings - they belong in data, not code. Non-Python contributors (translators) can edit YAML without touching Python imports of `DiscardReason` and `Task` enums. Matches the project's existing pattern (settings configs are also YAML).

## Out of scope (later in the chain)

- DE catalog, shared structural metadata for the discard widget, discard widget i18n, catalog completeness test: PR #207.
- `--locale` CLI flag, `record_builder` reading inheritance-resolved locale, re-import conflict warning: PR #208.

## Test plan

- [x] `uv run python -m pytest tests/unit` (707 passed)
- [x] `uv run mypy src/pragmata` clean (65 source files)
- [x] `test_locale_resolves_from_deployment_when_unset` / `_workspace_when_task_unset` / `_task_override_wins_over_workspace_and_deployment` / `_default_resolves_to_en` (locale participates in `resolved_task()` walk)
- [x] `test_catalog_drives_field_title` (catalog swap changes rendered output)
- [x] `tests/test_facade_lazy.py` (`Locale` present in public surface)


henrycgbaker:
@saschagobel design question: include `locale` in env layer (commit 6379d2f -> 3487e91)?
Originally i had PRAGMATA_ANNOTATION_LOCALE read in the env layer alongside ARGILLA_API_URL, then I moved it out (3487e91), reasoning is in PR body, but nb this is slight deviation from other variable management in this project (but I think is justified)

henrycgbaker:
@saschagobel 

i made a change here last min from having locale as StrEnum -> str. Did this for extensibility for future contributors, but want to check as requires some additional machinery.

Before: class Locale(StrEnum): EN = "en"; DE = "de" -> adding a locale = editing the schema.

After: Locale = str, validated at catalog lookup (get_catalog(locale) raises KeyError if no YAML registered). Combined w/ catalog auto-discovery, a deployment now adds a locale by dropping <code>.yaml into core/annotation/locales/ -> zero Python edit

- decouples "which languages we support" from the schema layer
- matches how SOTA i18n tools (gettext, i18next, Rails) treat locale codes as open strings, not closed enums
- no deployment has to fork to add a custom locale
BUT 
- loses compile-time autocomplete on Locale.X 

i thought minor worthwhile tradeoff, but worth checking

src/pragmata/core/annotation/locales/en.yaml
```diff
@@ -0,0 +1,72 @@
...
+    context_set: Context set
+    query: Query
+    discard_flow: Discard this record
+  generation:
+    query: Query
+    answer: Answer
+    context_set: Context set
+    discard_flow: Discard this record
+
+questions:
```
saschagobel, line 24: Minor copy/methodology consistency point: since these YAML strings define the annotation instrument, I’d align the question wording more tightly with the displayed field names and label semantics.

For retrieval, I’d use “chunk” rather than “passage”; for grounding, “context set” rather than “context” / “retrieved context”; and in generation, “answer” consistently instead of “response”.

src/pragmata/core/annotation/locales/en.yaml
```diff
@@ -0,0 +1,72 @@
...
+    fabricated_source: >-
+      Does the answer cite a source not present in the retrieved context?
+    notes: Notes (optional)
+    discard_reason: Discard reason
+    discard_notes: Discard notes (optional)
+  generation:
+    proper_action: Did the system choose the appropriate action for this query?
+    response_on_topic: Does the response substantively address the user's query?
+    helpful: >-
+      Would this response enable a typical user to make progress on their task?
```
saschagobel: I’d prefer “Does…” over “Would…” for `helpful` so annotators judge the observed answer rather than a hypothetical effect.

src/pragmata/core/annotation/locales/en.yaml
```diff
@@ -0,0 +1,72 @@
...
+    incomplete: Does the response fail to cover required parts of the query?
+    unsafe_content: Does the response contain unsafe or policy-violating content?
```
saschagobel: “policy-violating” is only annotation-stable if reviewers have a policy/guideline. Otherwise “unsafe or harmful content” is more self-contained.

src/pragmata/core/annotation/locales/en.yaml
```diff
@@ -0,0 +1,72 @@
...
+    contradicted_claim_present: >-
+      Does the provided context contradict any claim in the answer?
+    source_cited: Does the answer contain a citation marker?
+    fabricated_source: >-
+      Does the answer cite a source not present in the retrieved context?
+    notes: Notes (optional)
+    discard_reason: Discard reason
+    discard_notes: Discard notes (optional)
+  generation:
+    proper_action: Did the system choose the appropriate action for this query?
```
saschagobel: The `proper_action` wording also feels a bit ambiguous. Suggestion:  "Does the answer reflect the appropriate response type for the query?"

src/pragmata/core/annotation/locales/loader.py
```diff
@@ -0,0 +1,59 @@
...
+    Task.RETRIEVAL: ["topically_relevant", "evidence_sufficient", "misleading"],
+    Task.GROUNDING: [
+        "support_present",
+        "unsupported_claim_present",
+        "contradicted_claim_present",
+        "source_cited",
+        "fabricated_source",
+    ],
+    Task.GENERATION: ["proper_action", "response_on_topic", "helpful", "incomplete", "unsafe_content"],
+}
```
saschagobel, line 27: Minor maintainability note: the yes/no question set is now encoded in two places: the actual `rg.LabelQuestion`s in `argilla_task_definitions.py`, and `_YES_NO_QUESTIONS_BY_TASK` in the locale loader, which generates the `.yes` / `.no` catalog labels. That is fine for now, but it is a potential drift point if a question is added, renamed, or removed later.

Longer term, the shared question structure could be moved into one small structural metadata constant consumed by both modules, but a regression test is probably enough here for now.

src/pragmata/core/annotation/locales/loader.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+def load_catalog(path: Path) -> Catalog:
+    """Build a Catalog from a YAML translation file.
+
+    Raises ``KeyError`` if a required top-level section, task, field, question,
```
saschagobel: Tiny docstring nit: I think this can also raise ValueError and others depending on whats missing/malformed. Maybe phrase this as “raises during catalog loading” rather than naming the specific error.

src/pragmata/core/annotation/locales/registry.py
```diff
@@ -0,0 +1,26 @@
+"""Catalog registry — auto-discovers locale YAML files at import.
+
+A deployment adds a locale by dropping ``<code>.yaml`` into this directory;
```
saschagobel: Design concern: `Locale` is intentionally open, but catalog discovery currently only scans YAML files bundled inside the package. That means downstream users cannot actually add a locale without modifying the installed package. Could we either document this as “bundled locales only” for now, or plan a user-provided catalog directory/path setting so deployments can register locale YAML files outside `site-packages`?

src/pragmata/core/schemas/annotation_task.py
```diff
@@ -2,6 +2,19 @@
 
 from enum import StrEnum
 
+Locale = str
```
saschagobel: Minor typing/API surface question: since `Locale` is intentionally an open string alias, should this be declared explicitly as a type alias,  `type Locale = str`? Plain `Locale = str` works, but because `Locale` is re-exported as public API, the explicit alias form would make the intent clearer: this is documentation/typing surface, not a runtime locale object or enum.

src/pragmata/annotation/__init__.py
```diff
@@ -17,6 +17,7 @@
     "ExportResult",
     "IaaReport",
     "ImportResult",
+    "Locale",
```
saschagobel: what user-facing role should `Locale` play as a public export? As an internal type alias it makes sense because it gives the locale string a named semantic role across settings and catalog lookup.

saschagobel (approved):
Approved, thanks.

I left a few non-blocking comments/questions around the public Locale export, EN wording consistency, potential drift between task definitions and catalog structure, a small docstring mismatch, and the design question of how externally supplied locale catalogs should be supported.

PR #199, opened 2026-05-18
chore(deploy): clarify .env.dev.example comments and external service templates

henrycgbaker:
## Summary

- Improved explanation & clarity
- Replaces example external service URLs (`user:pass@host:5432/argilla` etc.) with `<angle-bracket>` placeholders so they are unambiguously templates, not real defaults.

## Test plan

- [x] `docker compose --env-file deploy/annotation/.env.dev.example config` resolves without errors.


saschagobel (approved):
All good, thanks.

PR #200, opened 2026-05-18
fix(annotation): echo auto-generated passwords once after setup

henrycgbaker:
## Summary

When `pragmata annotation setup` provisions users w/o explicit passwords, the API returns them in `SetupResult.generated_passwords` but the CLI never surfaced them - operator had clear way to retrieve them afterwards.

Now prints them once immediately after setup, with a prompt to record them:

```
Users created: 3, skipped: 0

Generated passwords (shown once - record them now):
  alice: xK9mP2qR
  bob: nL4jW7cT
```

## Test plan

- [x] `pragmata annotation setup --users users.json` with password-less user entries prints generated passwords.
- [x] Re-running setup (idempotent path, users skipped) produces no password output.
- [x] Users with explicit passwords in the JSON produce no password output for those users.


saschagobel (approved):
Thanks for the fix.

PR #201, opened 2026-05-18
fix(infra): make stack teardown safe alongside live annotation data

henrycgbaker:
## Summary

2 related risks in the Makefile where a stack-management target could silently destroy a live annotation stack's data --> fixed by making the destructive operation explicit and isolating the test stack from the dev stack.

### 1. `make docker-down` silently deleted volumes

- Previously ran `down -v`, wiping all annotation data volumes
- Split into two explicit targets with unambiguous names and docstrings:
  - `docker-down` - removes containers, **preserves volumes** (safe default)
  - `docker-down-clean` - removes containers **and volumes** (irreversible, explicit)
  - NB: `make docker-down -v` is not a workaround, as `-v` is interpreted by `make` as verbose.
- Clarified `docker-stop` comment (pauses containers, nothing removed).

### 2. `make test-stack` could destroy a live dev stack's volumes

- `test-stack` ended with `down -v` against the default Compose project. If a dev stack was already running, `docker-up` attached to it rather than spinning up a fresh one, and the trailing `down -v` would then wipe that live stack's volumes - no warning, no opt-in.
- Fix: run the smoke test under an isolated Compose project (`-p pragmata-test`) so its teardown can't touch the dev stack's containers, network, or volumes.
- Also wrapped the recipe in a single shell with `set -e` and `trap ... EXIT` so the isolated project is torn down on health-check failure as well as success (previously a failing health check leaked containers).

### Trade-off noted

Both stacks publish host port 6900, so running `make test-stack` while a dev stack is up will fail at `docker compose up --wait` with a port conflict. This is the preferred failure mode (loud, immediate + we don't need concurrency w/ a running dev stack) 

## Test plan

- [x] `make docker-down` stops containers but annotation data survives (`make docker-up` brings it back intact)
- [x] `make docker-down-clean` removes volumes as before (equivalent to old `make docker-down`)
- [x] `make docker-stop` still pauses stack as before
- [x] `make test-stack` runs under isolated project and tears itself down on success
- [x] `make test-stack` tears down cleanly when the health check fails (trap fires)
- [x] `make test-stack` does not touch volumes of a separately running dev stack (verify with a populated dev stack on the side)


saschagobel (approved):
Approved, thanks.

PR #202, opened 2026-05-19
feat(packaging): add [full] and [everything] extras

henrycgbaker:
Implements the `[full]` and `[everything]` extras specified in the config-and-settings ADR (#162).

- `full = ["pragmata[annotation,querygen,eval]"]` - all per-tool extras
- `everything = ["pragmata[full,cohere,deepseek,openai,anthropic,google-genai]"]` - full + all provider extras
- `eval = []` placeholder so `[full]` is forward-compatible when the eval tool lands


saschagobel (approved):
Thanks, looks good!

PR #204, opened 2026-05-20
ci: add automatic PR labeling

saschagobel:
### Summary

Adds lightweight automatic pull request labeling.

This introduces a branch-based PR labeler for common contribution types and a size labeler that classifies pull requests by changed lines of code. The setup is intentionally scoped to low-noise labels and avoids automatic subsystem/topic labeling.

### Key Changes

- Add `.github/labeler.yml` with branch-prefix rules.
- Add `.github/workflows/pr-labeler.yml` to apply PR type and size labels.
- Use size thresholds.
- Exclude `uv.lock` from PR size calculation.
- Update Dependabot PRs to use the `dependencies` label.

### DIFF
--- a/.github/workflows/pr-labeler.yml
+++ b/.github/workflows/pr-labeler.yml
@@ -1,0 +1,46 @@
+name: Pull Request Labeler
+
+on:
+  pull_request_target:
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.run_id }}
+  cancel-in-progress: true
+
+jobs:
+  label:
+    name: Label pull request
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Apply type labels
+        uses: actions/labeler@v6
+        with:
+          sync-labels: true
+
+      - name: Apply size label
+        if: github.event.action != 'edited'
+        uses: codelytv/pr-size-labeler@v1
+        with:
+          xs_label: 'size: XS'
+          xs_max_size: '49'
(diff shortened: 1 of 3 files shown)


henrycgbaker (approved):

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,46 @@
...
+jobs:
+  label:
+    name: Label pull request
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Apply type labels
+        uses: actions/labeler@v6
+        with:
+          sync-labels: true
```
henrycgbaker: will this clobber manual labels not in this labeler.yml? [edit: ignore - I misunderstood this, this is fine]

.github/workflows/pr-labeler.yml
```diff
@@ -0,0 +1,46 @@
...
+    types:
+      - opened
+      - reopened
+      - synchronize
+      - edited
+
+permissions:
+  contents: read
+  pull-requests: write
+  issues: write
```
henrycgbaker: is this permission needed?

saschagobel: Good point. `pull-requests: write` is required for applying existing labels to PRs. `issues: write` is only needed if the action should create missing repository labels, so we can drop it here.

.github/workflows/pr-labeler.yml
saschagobel: It will not remove arbitrary manual labels. It will remove labels that are managed by `.github/labeler.yml` if the PR no longer matches the corresponding rule. So the trade-off is intentional: `sync-labels: true` keeps automated type labels in sync, but manually overriding those same managed labels may be reverted by the workflow. If we want manual type labels to always win, we should set `sync-labels: false`.

PR #206, opened 2026-05-20
refactor(annotation): introduce WorkspaceSettings + TaskSettings with inheritance propagation

henrycgbaker:
**Stack (5 PRs):** #206 > #196 > #207 > #208 > feat/annotation-cli-inheritance-flags (upcoming)

**Chain goal:** Adds locale-aware Argilla datasets (EN/DE) on top of a new deployment/workspace/task settings inheritance hierarchy, with live widget locale switching and CLI flags for per-import overrides.

**This PR (1/5):** Settings inheritance restructure. Hoists `production_min_submitted` and `calibration_min_submitted` into a `deployment > workspace > task` hierarchy. Inheritance resolves at read time (CSS-style **specified -> computed** projection), not at construction.

---

## Goal

Make per-workspace and per-task overrides expressible in the annotation settings model. Today `AnnotationSettings` is flat at the deployment level only; adding any new per-workspace setting (e.g. `locale` from #196) repeats a dead-end. This PR establishes a CSS-style inheritance pattern that future fields plug into without re-architecting.

## Design: read-time resolution, not init-time mutation

Settings hold **specified** values; inheritance is a *projection* exposed via a small accessor that walks the tree on read. This matches how mature hierarchical-settings systems work (CSS computed-style, OmegaConf interpolation, DRF per-view overrides, MUI/Chakra themes): the tree stores literal user intent and resolution happens at use.

```python
# Raw field values: 'INHERIT' or an explicit override at any scope.
settings.workspaces["retrieval"].tasks[Task.RETRIEVAL].production_min_submitted  # INHERIT

# Concrete resolved values via the accessor (task -> workspace -> deployment walk).
settings.resolved_task("retrieval", Task.RETRIEVAL).production_min_submitted     # int, e.g. 1
```

`ResolvedTaskSettings` is a small `@dataclass(frozen=True)` with concrete (non-`Inherit`) field types (`int`, `int | None`). Consumers (`record_builder`, `setup`, `export_fetcher`, `export_runner`) read through the resolver, removing the need for `cast(int, ...)` at call sites.

### Why a frozen dataclass and not a parallel Pydantic BaseModel?

The resolved-shape *could* have been declared as `class ResolvedTaskSettings(BaseModel)` with concrete types mirroring `TaskSettings`. Deliberately avoided that:

| | Frozen dataclass (chosen) | Parallel BaseModel (avoided) |
|---|---|---|
| Per-call cost | One immutable allocation | Full Pydantic validation pipeline per resolve |
| Duplication risk | One declaration site | Two near-identical class bodies that drift |
| Type narrowing for consumers | `rt.production_min_submitted: int` | Same, but re-validates already-validated values |
| Serialisation footprint | Inert (no `.model_dump()` on a resolved snapshot) | Tempting to dump alongside raw; serialisation footgun |

`dict[str, Any]` was the other shape considered. Rejected: every consumer would need `cast(int, rt["production_min_submitted"])` and lose mypy coverage, defeating the point.

What we have: a **typed accessor object** (the dataclass). What we do not have: a **second Pydantic model** that re-validates values just to read them. That is the line a typed-accessor pattern draws at this scale.

## Scope

- `workspace_dataset_map: dict[str, dict[Task, TaskOverlap]]` becomes `workspaces: dict[str, WorkspaceSettings]`, with `WorkspaceSettings.tasks: dict[Task, TaskSettings]` and per-workspace overrides. `TaskOverlap` becomes `TaskSettings`.
- `production_min_submitted` / `calibration_min_submitted` hoisted to all three scope levels (deployment / workspace / task). Typed `T | Inherit = INHERIT` at the nested scopes; opinionated defaults live on `AnnotationSettings` only.
- New `Inherit` sentinel in `core/settings/settings_base.py` (distinct from `Unset`: `Unset` is stripped pre-validation; `Inherit` persists through validation as a field value, identity-compared via `is INHERIT`).
- New `ResolvedTaskSettings` frozen dataclass and `AnnotationSettings.resolved_task(ws_name, task)` accessor. The accessor walks task -> workspace -> deployment, picking the first non-`INHERIT` value per field.
- `_check_calibration_topology` validates against resolved values via the same accessor; error format is `workspace/task` pairs.
- Consumers (`record_builder`, `setup`, `export_fetcher`, `export_runner`) and test fixtures migrated to read through the resolver. `cast(int, ...)` wrappers removed.

`calibration_fraction` stays deployment-only. Per-task fractions would require a partition-manifest refactor (partitioning is record-level today). Future work.

## Why this shape (vs init-time mutation)

A `model_validator(mode="after")` that mutates child instances via `setattr` has two problems:
1. **Shared-instance bug.** A `WorkspaceSettings` or `TaskSettings` instance reused across two `AnnotationSettings` constructions silently freezes resolved values from the first. The second construction reads stale state. Covered by regression test (`test_shared_workspace_settings_never_mutated`).
2. **Lossy serialisation.** `model_dump()` returns post-inheritance concrete values; round-trip through YAML loses user intent. Mature systems (CSS, OmegaConf) all preserve specified values.

Read-time resolution avoids both: nothing mutates, `model_dump()` round-trips losslessly, the `INHERIT` sentinel survives serialisation, and consumers get concrete-typed values via the accessor.

## Breaking changes

Hard YAML break. No compat shims (per project convention).

```yaml
# BEFORE
annotation:
  workspace_dataset_map:
    retrieval:
      retrieval: { production_min_submitted: 1, calibration_min_submitted: 3 }

# AFTER
annotation:
  production_min_submitted: 1        # NEW at deployment level (default 1)
  calibration_min_submitted: 3       # NEW at deployment level (default 3)
  workspaces:
    retrieval:
      tasks:
        retrieval: {}                # inherits from deployment
```

Other behaviour changes: `_check_calibration_topology` error format is `workspace/task` pairs; empty `tasks: {}` / `workspaces: {}` are silently permitted; raw field reads (`ts.production_min_submitted`) may return `INHERIT`. Consumers must use `settings.resolved_task(ws_name, task)`.

## Test plan

- [x] `uv run python -m pytest tests/unit` (702 passed)
- [x] `uv run mypy src/pragmata` clean (61 files)
- [x] Shared-instance mutation regression (`test_shared_workspace_settings_never_mutated`)
- [x] Lossless `model_dump()` round-trip (`test_model_dump_preserves_inherit_sentinel`)
- [x] Integration test `TestPerWorkspaceMinSubmitted` against live Argilla (requires Docker stack)


src/pragmata/core/settings/annotation_settings.py
```diff
@@ -19,82 +29,120 @@ class ArgillaSettings(BaseModel):
     api_url: str | None = None
 
 
-class TaskOverlap(BaseModel):
-    """Per-task overlap topology: production and (optional) calibration thresholds.
+class TaskSettings(BaseModel):
+    """Per-task overrides; inherited fields default to ``INHERIT`` (use workspace value)."""
```
saschagobel, line 33: Minor docs point: the old TaskOverlap docstring explained the operational meaning of production_min_submitted / calibration_min_submitted and why the calibration default is 3. With the rename to TaskSettings, that context largely disappears. I’d prefer to keep that explanation either on TaskSettings, on the inherited fields in AnnotationSettings, or in the module docstring, because these defaults encode annotation/IAA methodology rather than just plumbing.

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -19,82 +29,120 @@ class ArgillaSettings(BaseModel):
...
+        ts = ws.tasks[task]
+        kwargs: dict[str, object] = {}
+        for name in self._INHERITED_FIELDS:
+            value = getattr(ts, name)
+            if value is INHERIT:
+                value = getattr(ws, name)
+            if value is INHERIT:
+                value = getattr(self, name)
+            kwargs[name] = value
+        return ResolvedTaskSettings(**kwargs)  # type: ignore[arg-type]
```
saschagobel: Small typing/readability point: `resolved_task()` currently uses dynamic `getattr` / `kwargs` and then suppresses mypy with `type: ignore`. Since there are only two inherited fields, I think explicit field resolution would be easier to read and would keep the accessor fully type-checked.

saschagobel (approved):
Thanks, approved — I only left two very minor comments regarding docs and typing.

I really like the new deployment/workspace/task settings organization and the read-time inheritance model.

PR #207, opened 2026-05-20
feat(annotation): discard widget i18n with chrome-locale toggle

henrycgbaker:
**Stack (6 PRs):** #206 > #196 > #207 > #219 > #208 > #209

**Chain goal:** Adds locale-aware Argilla datasets (EN/DE) on top of a deployment/workspace/task settings inheritance hierarchy, with live widget locale switching and CLI flags for per-import overrides.

**This PR (3/6):** Discard widget i18n - template machinery + EN widget chrome strings so the widget renders a locale-aware JSON payload at dataset-creation time and switches at runtime via Argilla's chrome-locale toggle.

---

## Scope

- `discard_flow.html`: i18n JSON payload + supported-locales array injected at render time using `string.Template` with `@@` delimiter (avoids `$nuxt` / `$i18n` collisions in the JS body); JS picks active locale via `window.parent.$nuxt.$i18n.onLanguageSwitched` with a chrome > `<html lang>` > dataset-creation > EN fallback chain.
- `_render_discard_template` / `_discard_i18n_payload_for_locale` in `argilla_task_definitions.py` do the substitution per `(task, dataset_locale)`. Widget payload assembly uses a comprehension over `DISCARD_WIDGET_KEYS` for the chrome strings. `_DiscardTemplate` docstring spells out the `@@`-delimiter rationale and warns future contributors not to introduce `@@` tokens in the JS body unless they are intended as substitution placeholders.
- `locales/en.yaml`: adds widget chrome strings (panel summary, button label, etc.) under a `widget` kind, keyed for fan-out into per-task widget payloads.
- `locales/loader.py`: widget-key helper supporting payload assembly.
- The existing HTML-option enum-sync test is rewritten against the rendered i18n payload (`TestDiscardFlowI18nPayload`).

## Stable across locales (exports depend on this)

- Field and question `name=` identifiers
- Label values: `yes`, `no`, `DiscardReason.*.value`

## Test plan

- [x] `uv run python -m pytest tests/unit` (724 passed)
- [x] `uv run mypy src/pragmata` clean
- [x] `uv run ruff check src/pragmata tests` clean
- [x] `uv run ruff format --check src/pragmata tests` clean
- [x] `tests/unit/core/annotation/test_argilla_task_definitions.py::TestDiscardFlowI18nPayload`: rendered widget contains every `DiscardReason` value
- [x] Manual: chrome locale toggle in Argilla flips widget strings live; non-widget strings remain frozen at dataset creation

PR #208, opened 2026-05-20
feat(annotation): --locale flag, per-workspace locale, re-import conflict warning

henrycgbaker:
**Stack (6 PRs):** #206 > #196 > #207 > #219 > #208 > #209

**Chain goal:** Adds locale-aware Argilla datasets (EN/DE) on top of a deployment/workspace/task settings inheritance hierarchy, with live widget locale switching and CLI flags for per-import overrides.

**This PR (5/6):** `--locale` CLI flag + `record_builder` reads inheritance-resolved locale per dataset + re-import locale-conflict warning.

---

## Scope

- `--locale` flag on `import` only (nb not on setup or teardown; locale only matters at dataset creation, and import is the entry point that materialises a dataset).
- `api/annotation_import.import_records` accepts a `locale` kwarg that overrides the deployment default through the settings inheritance established in #206.
- `record_builder.fan_out_records` reads `settings.resolved_task(ws_base, task).locale` per (workspace, task) iteration, so per-workspace and per-task locale carve-outs in YAML produce datasets in the right language.
- `record_builder._ensure_dataset` detects locale mismatch on re-import by inspecting the existing dataset's `LabelQuestion` option keys against `CATALOGS`; if it differs from the resolved locale, logs a warning and proceeds (warn + append: data integrity preserved because label *values* are locale-invariant; only display strings are frozen at dataset creation).

## Polish (applied in the same PR)

- `parse_locale(raw: str | None) -> Locale | None` (was `(raw: str) -> Locale`), matching the `parse_tasks` / `parse_user_specs` sibling signatures. Caller in `cli/commands/annotation.py` simplified to `locale=parse_locale(locale) or UNSET`.
- New unhappy-path CLI tests in `tests/unit/cli/test_parsing.py::TestParseLocale`: `None`-returns-`None`, `en`/`de` parse, whitespace tolerance, invalid value raises `ValueError`.
- `test_relocale_warn_append` assertion softened from `"'en'"`/`"'de'"` literal substrings (brittle to `%r` formatter changes) to `Locale.EN.value` / `Locale.DE.value` substrings.
- `fix(cli): import Locale type from core schema after facade drop` - one-line `TYPE_CHECKING`-only import-source swap, added as an accretive commit after rebasing onto the new parent (#219). Necessary because main's commit `333b15d` dropped `Locale` from the `pragmata.annotation` facade re-exports.

## Test plan

- [x] `uv run python -m pytest tests/unit` (731 passed)
- [x] `uv run mypy src/pragmata` clean (67 source files)
- [x] `TestImportLocaleConflict::test_relocale_warn_append`: re-import with mismatched `--locale` logs warning + appends
- [x] `TestParseLocale::test_invalid_locale_raises`: invalid `--locale xx` surfaces as `ValueError`
- [x] Manual: workspace-level locale carve-out renders correct strings end-to-end (one EN workspace + one DE workspace under the same deployment)
- [x] Manual: re-import to EN-existing dataset with `--locale de`; verify warning logged, records appended

### DIFF
--- a/src/pragmata/annotation/__init__.py
+++ b/src/pragmata/annotation/__init__.py
@@ -17,6 +17,8 @@ __all__ = [
     "ExportResult",
     "IaaReport",
     "ImportResult",
+    "Locale",
+    "SUPPORTED_LOCALES",
     "SetupResult",
     "Task",
     "UserSpec",
@@ -31,6 +33,8 @@ _LAZY: dict[str, tuple[str, str]] = {
     "ExportResult": ("pragmata.core.annotation.export_runner", "ExportResult"),
     "IaaReport": ("pragmata.core.schemas.iaa_report", "IaaReport"),
     "ImportResult": ("pragmata.api.annotation_import", "ImportResult"),
+    "Locale": ("pragmata.core.schemas.annotation_task", "Locale"),
+    "SUPPORTED_LOCALES": ("pragmata.core.annotation.locales.registry", "SUPPORTED_LOCALES"),
     "SetupResult": ("pragmata.core.annotation.setup", "SetupResult"),
     "Task": ("pragmata.core.schemas.annotation_task", "Task"),
     "UserSpec": ("pragmata.core.settings.annotation_settings", "UserSpec"),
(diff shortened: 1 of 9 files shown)


saschagobel (approved):
Thanks, all good.

PR #209, opened 2026-05-21
feat(annotation): --no-calibration and --calibration-partition-seed import flags

henrycgbaker:
**Stack (7 PRs):** #206 → #196 → #207 → #219 → #208 → this PR → [edit, new: #220]

## Goal

Two additional runtime-modifier flags on `pragmata annotation import` to round out scriptable per-invocation overrides. Builds on the inheritance settings model (#206) and the locale work (#196 / #207 / #219 / #208).

## Scope

- `--no-calibration` (boolean) - composite flag: sets `calibration_min_submitted=None` AND `calibration_fraction=0.0`. The single-flag way to say "no calibration for this import."
- `--calibration-partition-seed INT` - overrides the deployment-level partition seed for new-record assignment determinism in scripted/test runs. Existing assignments remain locked by the manifest.
- CLI-layer conflict check: `--no-calibration` + `--calibration-fraction > 0` → `typer.Exit(2)` with clear stderr. `--calibration-fraction 0` is fine (no conflict; `--no-calibration` already sets it).
- `api.annotation_import.import_records` gains matching `calibration_min_submitted: PositiveInt | None | Unset = UNSET` and `calibration_partition_seed: NonNegativeInt | Unset = UNSET` kwargs, threaded into `AnnotationSettings.resolve()` overrides. `--no-calibration` is a CLI-layer composite translated to underlying field values - the API does NOT gain a `no_calibration: bool` kwarg (keeps API tied to settings fields).

NOT added to setup / teardown / export / iaa - strict consumer-only (only `import` creates datasets where these fields apply).

## Why `--no-calibration` (vs just `--calibration-fraction 0.0`)

The two flags operate on different downstream paths:

- `calibration_fraction=0.0` - calibration dataset is still *created* during setup (setup keys off `calibration_min_submitted is not None`), it just receives 0 records this batch. Export and IAA still treat the task as calibration-bearing. A later import without the flag resumes populating it.
- `calibration_min_submitted=None` - calibration is *structurally absent* for this run: setup skips creating the dataset, record assignment skips the bucket, export/IAA skip the task.

Because this PR deliberately does **not** expose `--calibration-min-submitted` as a flag (per the tiering - project shape stays config-only), `--no-calibration` is the only CLI path to flip `calibration_min_submitted` to `None`. So `--calibration-fraction 0.0` alone is "import nothing to calibration this time"; `--no-calibration` is "no calibration concept for this batch at all." They're not interchangeable.

--
## Why no `--calibration-fraction`

Review converged on tiered model: flags for connection/identity, flags for runtime selection / per-invocation modifiers, config-only for project shape / tuning constants. Both flags here are per-invocation runtime modifiers (analogue to dbt `--vars`, pytest `-k`, pytest-randomly `--randomly-seed`).

Not exposed as flags (per the review): `--production-min-submitted` and `--calibration-min-submitted`  are project shape (dataset distribution config set once per dataset) -> keep this kind of threshold config-only.

## Test plan

- [x] `tests/unit/cli/test_cli_annotation.py::TestImportCommandCascadeFlags` - flag wiring + conflict check tests
- [x] `tests/unit/api/test_annotation.py` - API kwarg threading tests
- [x] `uv run python -m pytest tests/unit` - 735 passed
- [x] `uv run mypy src/pragmata` clean
- [x] Manual: end-to-end import with each flag invoked; verify partition manifest reflects the values

## References

- Stacked on #208 (locale + per-workspace + re-import conflict)
- Inheritance settings model from #206

### DIFF
--- a/src/pragmata/api/__init__.py
+++ b/src/pragmata/api/__init__.py
@@ -1,7 +1,7 @@
 """Internal API layer."""
 
-from pragmata.core.settings.settings_base import UNSET, Unset
+from pragmata.core.settings.settings_base import UNSET
 
 from .version import get_version
 
-__all__ = ["get_version", "UNSET", "Unset"]
+__all__ = ["get_version", "UNSET"]
--- a/src/pragmata/api/__init__.py
+++ b/src/pragmata/api/__init__.py
@@ -1,7 +1,7 @@
 """Internal API layer."""
 
-from pragmata.core.settings.settings_base import UNSET
+from pragmata.core.settings.settings_base import UNSET, Unset
 
 from .version import get_version
 
-__all__ = ["get_version", "UNSET"]
+__all__ = ["get_version", "UNSET", "Unset"]
--- a/src/pragmata/cli/commands/annotation.py
+++ b/src/pragmata/cli/commands/annotation.py
@@ -97,11 +97,23 @@ def import_command(
         "Falls through to YAML config and built-in default (0.1) when omitted; "
         "set to 0.0 for production-only batches.",
     ),
+    no_calibration: bool = typer.Option(
+        False,
+        "--no-calibration",
+        help="Disable calibration entirely for this batch: sets calibration_min_submitted=None "
+        "and calibration_fraction=0.0. Cannot be combined with --calibration-fraction > 0.",
+    ),
+    calibration_partition_seed: int | None = typer.Option(
+        None,
+        "--calibration-partition-seed",
+        help="Deterministic seed for assigning new records to calibration vs production. "
+        "Existing assignments are locked by the partition manifest.",
+    ),
     locale: str | None = typer.Option(
         None,
         "--locale",
         help="Deployment-level UI locale for Argilla dataset titles/questions/guidelines "
-        "(en, de). Cascades to workspaces/tasks unless they carve out a value in YAML. "
+        "(en, de). Inherits to workspaces/tasks unless they carve out a value in YAML. "
         "Falls back to config, then 'en'.",
     ),
 ) -> None:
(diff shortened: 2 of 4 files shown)


src/pragmata/cli/commands/annotation.py
```diff
@@ -113,6 +125,22 @@ def import_command(
...
+        raise typer.Exit(code=2)
+
+    resolved_calibration_fraction: float | Unset
+    resolved_calibration_min_submitted: int | None | Unset
+    if no_calibration:
+        resolved_calibration_fraction = 0.0
+        resolved_calibration_min_submitted = None
+    else:
+        resolved_calibration_fraction = UNSET if calibration_fraction is None else calibration_fraction
+        resolved_calibration_min_submitted = UNSET
```
saschagobel: The CLI-level translation of `--no-calibration` into `calibration_fraction=0.0` and `calibration_min_submitted=None` is fine because `--no-calibration` is intentionally a CLI-only composite flag.

However, I’d avoid importing `Unset` just to annotate local intermediate variables; the command can pass conditional expressions directly into `annotation.import_records()` and keep the CLI surface coupled only to `UNSET`.


E.g.,

```python
if no_calibration and calibration_fraction is not None and calibration_fraction > 0:
    typer.echo(
        "Error: --no-calibration cannot be combined with --calibration-fraction > 0.",
        err=True,
    )
    raise typer.Exit(code=2)

result = annotation.import_records(
    records,
    api_url=UNSET if api_url is None else api_url,
    base_dir=UNSET if base_dir is None else base_dir,
    calibration_fraction=(
        0.0
        if no_calibration
        else UNSET
        if calibration_fraction is None
        else calibration_fraction
    ),
    calibration_min_submitted=None if no_calibration else UNSET,
    calibration_partition_seed=UNSET if calibration_partition_seed is None else calibration_partition_seed,
    locale=parse_locale(locale) or UNSET,
)
```

src/pragmata/api/__init__.py
saschagobel, line 1: If the CLI no longer imports `Unset`, I’d revert the `pragmata.api.__init__` change. `UNSET` is the runtime sentinel that CLI/API callers need; `Unset` is only a typing alias and does not need to be re-exported through the API namespace. Keeping it internal avoids widening the API surface unnecessarily.

saschagobel (approved):
Approved — the CLI-level --no-calibration translation looks appropriate, but I’d avoid importing/re-exporting Unset just for local annotations and keep the CLI coupled only to the runtime UNSET sentinel.

saschagobel (approved):

PR #210, opened 2026-05-21
docs(annotation): contributor README for adding locale YAML catalogs

henrycgbaker:
DEFERRED FOR FULL DOCS DISCUSSION

**Stack (6 PRs):** #206 → #196 → #207 → #208 → #209 → this PR

**Chain goal:** Locale-aware Argilla datasets (EN/DE) on top of a deployment/workspace/task settings inheritance hierarchy, with live widget locale switching and CLI flags for per-import overrides.

**This PR (6/6):** Contributor README for `core/annotation/locales/`.

---

## Goal

Document the locales package so a contributor (translator or dev) can add a new locale without reading the source. Reflects the YAML-as-data + open-`Locale`-string layout introduced in #196 and the discard-widget chrome strings introduced in #207.

## Scope

`src/pragmata/core/annotation/locales/README.md` — single file, sits next to the YAML catalogs it documents.

Covers:
- Package layout (`types.py` / `loader.py` / `registry.py` / `<code>.yaml`)
- Catalog shape (YAML nesting on disk → flat tuple-keyed dict in memory)
- What stays stable across locales (machine identifiers, label values, key set)
- Adding a new locale — 3 steps: copy YAML, translate values, run tests
- Correctness invariants (`_YES_NO_QUESTIONS_BY_TASK`, `DISCARD_WIDGET_KEYS`, `DiscardReason`)
- Automatic vs manual entries (what the loader fans out vs what the YAML must spell out)

## Why this beats the old version

The old README documented the Python-as-data flow: edit the `Locale` enum, write a `<locale>.py` module calling `build_programmatic_entries()`, register the catalog in `registry.py`. That's three Python edits plus a YAML-free workflow that excluded non-Python translators.

After #196 the flow is: drop a YAML file. Zero Python edits. The README now matches.

## Test plan

- [x] `uv run ruff check src tests` clean (markdown-only change; no code touched)
- [ ] Manual: a contributor following the README produces a valid catalog with no `KeyError` at load


henrycgbaker:
@saschagobel I've put this together for now as quick first pass, but it's low priority and i think should wait until we've had full discussion about documentation etc, so leaving in draft for now. Can also close & leave branch up if preferred

saschagobel:
Closing this as part of the alpha release documentation scope cut. The contributor guide would still be useful, but it is lower priority than the user-facing workflow documentation, and the current branch would need to be rebuilt on main before it is reviewable. We can bring it back later as a clean, docs-only PR if needed. Thanks for putting this together.

PR #211, opened 2026-05-21
feat(annotation): add --after, --before and --exclude-annotators IAA filters

henrycgbaker:
## Goal

Expose the existing `compute_iaa()` filter kwargs through the `annotation iaa` CLI so users can scope IAA computation by submission window and annotator set without writing Python.

## Scope

- `cli/commands/annotation.py` - three new options on `annotation iaa`:
  - `--after ISO_DATETIME` / `--before ISO_DATETIME` - keep only annotations submitted in the requested window
  - `--exclude-annotators ID1,ID2,...` - drop listed annotator IDs
- `cli/parsing.py` - two new helpers:
  - `parse_datetime` - ISO 8601 via `datetime.fromisoformat`, raises `typer.BadParameter` on invalid input
  - `parse_annotator_ids` - comma-separated, whitespace-stripped, empty entries dropped
- Both accept `str | None` and return `T | None`, matching the existing `parse_locale`/`parse_tasks` shape

## Implementation

- New parse helpers handle None upfront, eliminating `None if x is None else parse(x)` triples at the callsite
- Filters are runtime params (not settings-bearing), so they pass through `None` directly to `api/annotation_iaa.py` - no `UNSET` bridge needed (per the CLI conventions in `CLAUDE.md`)

## Testing

- 40 unit tests pass:
  - `tests/unit/cli/test_cli_annotation.py` - 5 new tests covering the three flags (happy path, invalid datetime, empty annotator list)
  - `tests/unit/cli/test_parsing.py` - 9 new tests for `parse_datetime` and `parse_annotator_ids`, including unhappy paths

## References

- Stacked on #209 (`feat/annotation-cli-inheritance-flags`)
- `docs/design/annotation-iaa.md` - IAA computation surface
- `core/annotation/iaa.py`- `compute_iaa()` filter kwargs being exposed

### DIFF
--- a/src/pragmata/cli/commands/annotation.py
+++ b/src/pragmata/cli/commands/annotation.py
@@ -225,12 +225,13 @@ def iaa_command(
     after: str | None = typer.Option(
         None,
         "--after",
-        help="Keep only annotations submitted on or after this ISO 8601 datetime (e.g. 2026-05-01T00:00:00).",
+        help="Keep only annotations submitted on or after this ISO 8601 date or datetime "
+        "(e.g. 2026-05-01 or 2026-05-01T00:00:00; a date is treated as midnight).",
     ),
     before: str | None = typer.Option(
         None,
         "--before",
-        help="Keep only annotations submitted before this ISO 8601 datetime.",
+        help="Keep only annotations submitted before this ISO 8601 date or datetime.",
     ),
     exclude_annotators: str | None = typer.Option(
         None,
--- a/src/pragmata/cli/commands/annotation.py
+++ b/src/pragmata/cli/commands/annotation.py
@@ -3,7 +3,13 @@
 import typer
 
 from pragmata.api import UNSET
-from pragmata.cli.parsing import parse_locale, parse_tasks, parse_user_specs
+from pragmata.cli.parsing import (
+    parse_annotator_ids,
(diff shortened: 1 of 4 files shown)


saschagobel:
Thanks, this PR needs a rebase and conflict resolution before review, so the diff only includes the intended files.

src/pragmata/cli/commands/annotation.py
```diff
@@ -216,6 +222,21 @@ def iaa_command(
...
+    after: str | None = typer.Option(
+        None,
+        "--after",
+        help="Keep only annotations submitted on or after this ISO 8601 datetime (e.g. 2026-05-01T00:00:00).",
+    ),
+    before: str | None = typer.Option(
+        None,
+        "--before",
+        help="Keep only annotations submitted before this ISO 8601 datetime.",
+    ),
```
saschagobel, line 235: Non-blocking: since datetime.fromisoformat() accepts date-only values and normalizes them to midnight datetimes, the CLI help should say ISO 8601 date/datetime rather than only datetime for both --after and --before. This avoids under-documenting a supported input form already covered by the CLI tests.

saschagobel (approved):
Thanks, approved, left one minor. non-blocking comment on cli help.

PR #212, opened 2026-05-21
refactor(cli): rename querygen --config to match annotation CLI

henrycgbaker:
## Goal

Cross-tool naming consistency: standardise on `--config` for config-file paths across all pRAGmata CLIs. `annotation` uses `--config` (5 subcommands); `querygen` was lone outlier using `--config-path`. `--config` is more concise.

## Scope

- `src/pragmata/cli/commands/querygen.py` - rename CLI flag from `--config-path` to `--config`
- Internal Python kwarg names unchanged (`config_path: str | Path` still flows through `api/querygen.py`)
- Only the user-facing CLI string changes

## Implementation

One-line option-string change. No behaviour change.

## Testing

- Manual: `pragmata querygen run --config <path>` resolves the same way as before
- No new tests needed - pure flag rename

## References

- Identified by a CLI parity audit across `cli/commands/`
- Aligns with the dominant convention: 5 of 6 config-accepting subcommands already use `--config`


saschagobel (approved):
All good, thanks!

lramir14 (approved):
Looks good. Approved.

PR #213, opened 2026-05-22
refactor(annotation): declarative LogicalConstraint SSOT for export checks

henrycgbaker:
**Stack (6 PRs):** #213 → #214 → #223 → #215 → #216 → #217

**This PR (1/6):** Replace hardcoded `if row.x and not row.y` violation checks with a declarative `LogicalConstraint` catalogue. Same export-time behaviour, now driven from one definition that the rest of the stack consumes.

---

## Scope

- `core/annotation/logical_constraints.py` (new): `LogicalConstraint` dataclass (binary implication `<when_question>=<when_value>` requires `<then_question>=<then_value>`, plus `constraint_id`, `message`), and `LOGICAL_CONSTRAINTS: dict[Task, list[LogicalConstraint]]`. `CONSTRAINT_BY_ID` flattened lookup used by the settings layer to validate override maps.
- `core/annotation/constraints.py`: task-keyed convenience checkers (`check_retrieval`/`check_grounding`/`check_generation`) over the new catalogue. `CONSTRAINT_CHECKERS: dict[Task, Callable]` is the public consumer surface.
- `violation_string()` preserves the existing export CSV/`constraint_summary` format byte-for-byte so downstream consumers don't break.

## Why

PRs 2-5 in the stack need to consume the same definitions (UI widget needs the messages; configurable-severity needs `constraint_id` keys; export summary keys by `constraint_id`). Keeping export-time and annotation-time off one declarative list guarantees they cannot drift.

Severity isn;t on the dataclass: it's a configurable runtime setting and lives on `AnnotationSettings.constraint_severity` (in PR 2).

## Test plan
- [x] `uv run python -m pytest tests/unit`


src/pragmata/core/annotation/export_constraint_checks.py
```diff
@@ -0,0 +1,53 @@
...
+    """Return constraint violation strings for a grounding annotation row."""
+    return _evaluate(Task.GROUNDING, row)
+
+
+def check_generation(row: GenerationAnnotation) -> list[str]:
+    """Return constraint violation strings for a generation annotation row."""
+    return _evaluate(Task.GENERATION, row)
+
+
+CONSTRAINT_CHECKERS: dict[Task, Callable] = {
```
saschagobel: Non-blocking typing suggestion: since `CONSTRAINT_CHECKERS` is the shared consumer surface, it might be worth making the return contract explicit instead of using bare `Callable`. For example:

```python
from collections.abc import Callable

type ConstraintChecker = Callable[..., list[str]]

CONSTRAINT_CHECKERS: dict[Task, ConstraintChecker] = {
    Task.RETRIEVAL: check_retrieval,
    Task.GROUNDING: check_grounding,
    Task.GENERATION: check_generation,
}
```

The `...` keeps the task-specific row argument flexible, while the alias documents that all checkers return violation strings.

tests/unit/core/annotation/test_logical_constraints.py
saschagobel, line 1: Optional/Non-blocking: I think the generic implication behavior is covered. The only possible gap is that the actual LOGICAL_CONSTRAINTS catalogue is not pinned against the annotation protocol, so a typo or accidental rule change could pass these tests. A single compact catalogue-shape/contract test may be enough here; I would not add a full per-constraint behavior matrix.

saschagobel (approved):
Thanks, this looks good. I left two small non-blocking suggestions: one around making the checker-map typing a bit more explicit, and one optional note on pinning the constraint catalogue against the annotation protocol in tests.

saschagobel (approved):

PR #214, opened 2026-05-22
feat(annotation): constraint severity in AnnotationSettings

henrycgbaker:
**Stack:** #213 (merged) → **#214** → #223 → #215 → #216 → #217

**Chain/feature goal:** Annotatotion-time logical-constraint feedback (warn/block annotator if/when annotation violates declared logical constraints). Sits on top of a shared constraint SSOT, with deployment- and workspace-level severity overrides.

Severity becomes a property of a deployment, not of a rule, so it lives in `AnnotationSettings` rather than on `LogicalConstraint`. 

**This PR (settings layer):** Add deployment-scope `constraint_severity` to `AnnotationSettings` - the SSOT the annotator-time widget (#223) consumes. Severity is a runtime setting, not a property of the rule.

---

## Scope

- `core/annotation/logical_constraints.py`: `Severity` type alias; `CONSTRAINT_BY_ID` by-id lookup over the flattened catalogue, used to validate that override maps reference known `constraint_id`s.
- `core/settings/annotation_settings.py`: `constraint_severity: dict[str, Severity]` with full defaults via `_DEFAULT_CONSTRAINT_SEVERITY` (alongside other deployment defaults like `production_min_submitted`, `calibration_min_submitted`). A `model_validator(mode="before")` overlays user-supplied entries on top (sparse overlay — configs specify only the `constraint_id`s they override; omitted keys fall through). Unknown `constraint_id`s are rejected at construction.

## Test plan
- [x] `uv run python -m pytest tests/unit` (978 passing)

### DIFF
--- a/src/pragmata/core/annotation/logical_constraints.py
+++ b/src/pragmata/core/annotation/logical_constraints.py
@@ -85,6 +85,24 @@ LOGICAL_CONSTRAINTS: dict[Task, list[LogicalConstraint]] = {
     Task.GENERATION: [],
 }
 
-# By-id lookup over the flattened catalogue, used by the settings layer to
-# validate that override maps reference known constraint_ids.
-CONSTRAINT_BY_ID: dict[str, LogicalConstraint] = {c.constraint_id: c for cs in LOGICAL_CONSTRAINTS.values() for c in cs}
+
+def _build_constraint_by_id(
+    catalogue: dict[Task, list[LogicalConstraint]] = LOGICAL_CONSTRAINTS,
+) -> dict[str, LogicalConstraint]:
+    """By-id lookup over the flattened catalogue.
+
+    Used by the settings layer to validate that override maps reference known
+    constraint_ids. Raises eagerly on a duplicate constraint_id rather than
+    silently letting the last one win, which would otherwise hide the earlier
+    constraint from validation and export checks.
+    """
+    by_id: dict[str, LogicalConstraint] = {}
+    for constraints in catalogue.values():
+        for constraint in constraints:
+            if constraint.constraint_id in by_id:
+                raise ValueError(f"duplicate constraint_id {constraint.constraint_id!r} in constraint catalogue")
+            by_id[constraint.constraint_id] = constraint
+    return by_id
+
+
+CONSTRAINT_BY_ID: dict[str, LogicalConstraint] = _build_constraint_by_id()
--- a/src/pragmata/core/annotation/logical_constraints.py
+++ b/src/pragmata/core/annotation/logical_constraints.py
@@ -7,11 +7,16 @@ the export code.
 """
 
 from dataclasses import dataclass
-from typing import Literal
+from enum import StrEnum
 
 from pragmata.core.schemas.annotation_task import Task
 
-Severity = Literal["warn", "block"]
+
+class Severity(StrEnum):
+    """Annotator-time feedback severity for a logical constraint violation."""
+
+    WARN = "warn"
+    BLOCK = "block"
 
 
 @dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


henrycgbaker:
@saschagobel this PR is big; I tried to divide it up, but semantically it makes sense as a cohesive whole

src/pragmata/core/annotation/logical_constraints.py
```diff
@@ -7,9 +7,12 @@
 """
 
 from dataclasses import dataclass
+from typing import Literal
 
 from pragmata.core.schemas.annotation_task import Task
 
+Severity = Literal["warn", "block"]
```
ddimmery: I think it's mainly aesthetic, but I would lean towards a `StrEnum` here. As far as I can tell, the repository already uses a combination of `Literal` and `StrEnum` (ADR-0005 suggests `StrEnum` on the contract layer, but we aren't there).

src/pragmata/core/annotation/logical_constraints.py
```diff
@@ -81,3 +84,7 @@ def violation_string(self) -> str:
     ],
     Task.GENERATION: [],
 }
+
+# By-id lookup over the flattened catalogue, used by the settings layer to
+# validate that override maps reference known constraint_ids.
+CONSTRAINT_BY_ID: dict[str, LogicalConstraint] = {c.constraint_id: c for cs in LOGICAL_CONSTRAINTS.values() for c in cs}
```
ddimmery: add check for uniqueness of `constraint_id` to avoid confusing errors where the last added constraint overwrites an earlier constraint with duplicate ID.

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -26,6 +33,13 @@
 # Inheritable calibration_fraction override: bounded [0, 1] or the INHERIT sentinel.
 CalibrationFractionOverride = Annotated[float, Field(ge=0.0, le=1.0)] | Inherit
 
+_DEFAULT_CONSTRAINT_SEVERITY: dict[str, Severity] = {
+    "evidence_requires_relevance": "block",
+    "evidence_excludes_misleading": "warn",
+    "contradiction_requires_unsupported": "block",
+    "fabricated_requires_cited": "block",
+}
+
```
ddimmery, line 45: I assume it makes sense later why this isn't just added to the definition of `LogicalConstraint`, which seems like it would be more straightforward with less coupling.

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -146,6 +161,29 @@ def resolved_task(self, workspace_name: str, task: Task) -> ResolvedTaskSettings
...
+    @model_validator(mode="before")
+    @classmethod
+    def _merge_constraint_severity_defaults(cls, data: object) -> object:
+        """Overlay user-provided deployment severities onto the built-in defaults.
+
+        Users supply only the constraint_ids they want to override; the rest
+        fall through to the field's default.
+        """
+        if isinstance(data, dict) and isinstance(data.get("constraint_severity"), dict):
+            data["constraint_severity"] = {**_DEFAULT_CONSTRAINT_SEVERITY, **data["constraint_severity"]}
```
ddimmery: This mutates the input object. Given how this is currently used (i.e. a freshly generated dict) that's not a big problem, but it might be safer to `data = {**data, "constraint_severity": {...}}`

ddimmery (approved):
Minor comments. The uniqueness check seems like the most useful one.

src/pragmata/core/settings/annotation_settings.py
henrycgbaker: because severity is a user-configurable deployment level concern -> same LogicalConstraint can be 'block' or a 'warn' in different deployment, I added a quick comment lines 36-8

PR #216, opened 2026-05-22
feat(annotation): configurable constraint severity at deployment + workspace scopes

henrycgbaker:
**Stack:** #213 (merged) → #214 (severity settings) → #223 (widget) → #215 → **#216** → #217

**This PR (5/6):** Add per-workspace severity overrides on top of the deployment-scope defaults introduced in PR 2. The annotator-time widget picks up the resolved severity per workspace.

---

## Scope

- `core/settings/annotation_settings.py`:
  - `WorkspaceSettings.constraint_severity: dict[str, Severity]` (sparse, empty default). Listed constraint_ids override the deployment value for that workspace only; omitted constraint_ids fall through.
  - `AnnotationSettings.resolved_severity(workspace_name, constraint_id)` walks workspace → deployment. Deployment defaults are guaranteed complete by the merge validator added in PR 2.
  - `AnnotationSettings.task_to_workspace()` inverts the topology to a `Task → workspace_name` map (validated 1:1 by the existing task-uniqueness validator). Consolidates the workspace lookup; the duplicate `_invert_workspace_map` helper is removed from `record_builder` and all its callers move to `settings.task_to_workspace()` — `fan_out_records` and `assign_partitions` (the latter being calibration/partition code that gained a use of the helper on `main` since this PR was first written), plus `argilla_task_definitions.build_task_settings`.
  - Validator rejects unknown `constraint_id`s at workspace scope (parallel to the deployment-scope check from PR 2).
- `core/annotation/argilla_task_definitions.py`: `_render_constraints_template` resolves severity per `(workspace_name, constraint_id)` so the widget renders the workspace-effective severity. `build_task_settings` skips tasks not present in the workspaces topology (rather than building unused `rg.Settings` that can't resolve severity).

## Override semantics

- `workspaces["foo"].constraint_severity = {"evidence_excludes_misleading": "block"}` upgrades that one constraint to blocking for workspace `foo`. Other constraints inherit the deployment value.
- Unknown `constraint_id` in an override map fails Pydantic validation at deployment-config load time (with the list of known ids in the error message).

## Test plan
- [x] `uv run python -m pytest tests/unit`

### DIFF
--- a/src/pragmata/core/annotation/argilla_task_definitions.py
+++ b/src/pragmata/core/annotation/argilla_task_definitions.py
@@ -18,6 +18,7 @@ dataset creation time.
 """
 
 import json
+from collections.abc import Callable
 from importlib.resources import files
 from string import Template
 from typing import Any
@@ -319,7 +320,7 @@ def build_task_settings(settings: AnnotationSettings, locale: Locale = "en") -> 
     # Build only for tasks present in the workspaces topology: the constraints
     # widget renders against the task's workspace, and there's no consumer for
     # the unused rg.Settings.
-    task_builders: dict[Task, callable] = {
+    task_builders: dict[Task, Callable[[], rg.Settings]] = {
         Task.RETRIEVAL: lambda: assemble(
             Task.RETRIEVAL,
             content_fields=[
--- a/src/pragmata/core/annotation/argilla_task_definitions.py
+++ b/src/pragmata/core/annotation/argilla_task_definitions.py
@@ -151,16 +151,18 @@ def _render_discard_template(template_text: str, task: Task, dataset_locale: Loc
     )
 
 
-def _render_constraints_template(task: Task, questions: list, template_text: str, settings: AnnotationSettings) -> str:
+def _render_constraints_template(
+    task: Task, questions: list, template_text: str, settings: AnnotationSettings, workspace_name: str
+) -> str:
     """Substitute the constraint + question-title payload into ``constraints_field.html``.
 
-    Each constraint's payload severity is the deployment-scope value from
-    ``settings.constraint_severity``.
+    Each constraint's payload severity is the resolved severity for
+    ``(workspace_name, constraint_id)``: workspace overrides win over deployment.
     """
     constraints = LOGICAL_CONSTRAINTS[task]
     referenced = {q for c in constraints for q in (c.when_question, c.then_question)}
     titles = {q.name: q.title for q in questions if isinstance(q, rg.LabelQuestion) and q.name in referenced}
-    payloads = [c.to_widget_payload(settings.constraint_severity[c.constraint_id]) for c in constraints]
+    payloads = [c.to_widget_payload(settings.resolved_severity(workspace_name, c.constraint_id)) for c in constraints]
     return Template(template_text).substitute(
         CONSTRAINTS_JSON=json.dumps(payloads, ensure_ascii=False),
         QUESTION_TITLES_JSON=json.dumps(titles, ensure_ascii=False),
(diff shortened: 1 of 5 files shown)


tests/unit/core/settings/test_annotation_settings.py
```diff
@@ -525,6 +525,116 @@ def test_yaml_subset_override(self):
...
+                "retrieval": WorkspaceSettings(
+                    constraint_severity={"evidence_excludes_misleading": "block"},
+                    tasks={Task.RETRIEVAL: TaskSettings()},
+                ),
+                "grounding": WorkspaceSettings(tasks={Task.GROUNDING: TaskSettings()}),
+                "generation": WorkspaceSettings(tasks={Task.GENERATION: TaskSettings()}),
+            },
+        )
+        # retrieval workspace has the override
+        assert s.resolved_severity("retrieval", "evidence_excludes_misleading") == "block"
```
ddimmery, line 591: It appears you're only testing _within_ workspace, not in separate workspaces.

Should add something like 

 ```assert s.resolved_severity("grounding", "evidence_excludes_misleading") == "warn"```

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -176,12 +196,15 @@ def _merge_constraint_severity_defaults(cls, data: object) -> object:
...
-            )
+        for scope_name, overrides in [("deployment", self.constraint_severity)] + [
+            (f"workspace {ws_name!r}", ws.constraint_severity) for ws_name, ws in self.workspaces.items()
+        ]:
+            unknown = set(overrides) - known
+            if unknown:
+                raise ValueError(
+                    f"{scope_name} constraint_severity references unknown constraint_id(s): "
+                    f"{sorted(unknown)}. Known constraint_ids: {sorted(known)}."
+                )
```
ddimmery, line 210: It appears that this is just checking the IDs exist, not that they're in the proper workspace. Consider a config like:

```yaml
workspaces:
  retrieval:
    constraint_severity: { fabricated_requires_cited: warn }   # a GROUNDING constraint
```

This would pass validation, but do nothing because `fabricated_requires_cited` is in `GROUNDING`, not `RETRIEVAL`.

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -140,6 +143,23 @@ class AnnotationSettings(ResolveSettings):
...
+
+        Severity is per-constraint (not per-task), so this walks only the two
+        scopes where it can be set. Deployment defaults are guaranteed complete
+        by the merge validator on ``constraint_severity``; unknown ``constraint_id``
+        raises ``KeyError``.
+        """
+        ws = self.workspaces[workspace_name]
+        if constraint_id in ws.constraint_severity:
+            return ws.constraint_severity[constraint_id]
+        return self.constraint_severity[constraint_id]
```
ddimmery, line 164: This assumes every constraint has a default, which I think is only enforced via unit tests (it at all). Would it make sense to have a global fallback default here?

src/pragmata/core/annotation/argilla_task_definitions.py
```diff
@@ -359,3 +367,4 @@ def assemble(task: Task, content_fields: list, questions: list, metadata: list,
             guidelines=t(Task.GENERATION, "guidelines", ""),
         ),
     }
+    return {task: build() for task, build in task_builders.items() if task in task_to_ws}
```
ddimmery, line 370: So if `settings.task_to_workspace()` doesn't include a task, we'll leave it out? This is new behavior, isn't it?

tests/unit/core/annotation/test_argilla_task_definitions.py
```diff
@@ -354,3 +354,40 @@ def test_field_is_below_content_above_discard(self):
...
+        import re
+
+        match = re.search(r"var CONSTRAINTS = (\[.*?\]);", template, re.DOTALL)
+        assert match, "CONSTRAINTS JSON not found in template"
+        return {c["constraint_id"]: c["severity"] for c in json.loads(match.group(1))}
+
+    def test_no_overrides_uses_deployment_defaults(self):
+        settings = AnnotationSettings()
+        task_settings = build_task_settings(settings)
+        for task in (Task.RETRIEVAL, Task.GROUNDING):
```
ddimmery: Any reason to only use the two?

ddimmery (approved):
No major problems, but a few minor concerns that would improve the diff.

tests/unit/core/annotation/test_argilla_task_definitions.py
henrycgbaker: Generation doesn't have any constraints (currently), but i'll add it in for completeness

src/pragmata/core/annotation/argilla_task_definitions.py
henrycgbaker: Yes, a task needs an owning workspace to resolve severity against / its rg.Settings would have no consumer - so it's skipped

PR #217, opened 2026-05-22
refactor(annotation): key constraint_summary by constraint_id

henrycgbaker:
**Stack:** #213 (merged) → #214 (severity settings) → #223 (widget) → #215 → #216 → **#217**

**This PR (6/6):** Finish the SSOT migration on the export side. `constraint_summary` aggregation in `export_meta.json` now keys by stable `constraint_id` (e.g. `evidence_requires_relevance`) instead of the verbose violation string. Closes the inconsistency where PR 5's (#216) severity overrides used `constraint_id` keys but the export summary used the formatted message string.

---

## Scope

- `core/annotation/export_constraint_checks.py`: `check_*` return `list[LogicalConstraint]` instead of `list[str]`; `CONSTRAINT_CHECKERS` and the `ConstraintChecker` type alias updated accordingly.
- `core/annotation/export_fetcher.py`: row tuples now carry `list[LogicalConstraint]` through to the runner.
- `core/annotation/export_runner.py`:
  - `write_export_csv`: per-row `constraint_details` CSV column unchanged - still verbose `";"`-joined violation strings (generated via `LogicalConstraint.violation_string()` at write time).
  - `run_export`: `constraint_summary[c.constraint_id] += 1` aggregation.
- Tests updated for the new key format.

## What stays the same

- Per-row CSV `constraint_details` column format (verbose strings - human-readable).
- `constraint_violated: bool` column (still derived from `bool(violations)`).
- `AnnotationExportMeta.constraint_summary` schema type (`dict[str, NonNegativeInt]`) - keys are arbitrary strings, only the runtime values change.

## Breaking change

`export_meta.json`'s `constraint_summary` keys change from verbose strings to short ids. If any downstream consumer reads those keys, it needs updating. Counts and per-row CSV data are unaffected.

## Test plan
- [x] `uv run python -m pytest tests/unit` (994 passing)

### DIFF
--- a/src/pragmata/core/annotation/logical_constraints.py
+++ b/src/pragmata/core/annotation/logical_constraints.py
@@ -4,7 +4,7 @@ A logical constraint is a binary implication on label values: when one question
 is answered with a specific yes/no value, another question is constrained to a
 specific yes/no value. The same definitions drive:
 
-1. Export-time validation (``constraints.check_*`` returns violation strings)
+1. Export-time validation (``export_constraint_checks.check_*`` returns violated constraints)
 2. Annotation-time UI warnings/blocks (constraints serialised via
    ``to_widget_payload()`` into ``constraints_field.html``)
 
--- a/src/pragmata/core/annotation/logical_constraints.py
+++ b/src/pragmata/core/annotation/logical_constraints.py
@@ -53,7 +53,7 @@ class LogicalConstraint:
         return getattr(row, self.then_question, None) != self.then_value
 
     def violation_string(self) -> str:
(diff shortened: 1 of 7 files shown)


src/pragmata/core/annotation/logical_constraints.py
ddimmery: This should be updated.

ddimmery (approved):
The design doc (`docs/design/annotation-export-pipeline.md:204`) still uses the old format, it would be worth updating to keep this all in sync, unless these docs are really meant to be immutable (then would an ADR be the right way to go?).

PR #220, opened 2026-05-25
feat(annotation): user-provided locale catalog directory

henrycgbaker:
**Stack (7 PRs):** #206 → #196 → #207 → #219 → #208 → #209→  this PR 

## Goal

Allow deployments to add or override Argilla dataset locales by dropping `*.yaml` files into a user-provided directory, without modifying the installed `pragmata` package or having a Python toolchain. Closes the gap raised in #196 review: `Locale` is intentionally an open string, but catalog discovery had been bundled-only.

## Scope

- **`AnnotationSettings.locale_catalog_dir: DirectoryPath | None`** (deployment scope, default `None`). When set, `*.yaml` files in this directory are layered over the bundled catalogs at registration time; user files override bundled entries on stem collision (matches gettext / i18next layering convention). Pydantic `DirectoryPath` enforces the path exists and is a directory at settings construction.
- **`register_catalog_dir(directory)`** in `core/annotation/locales/registry.py`. Idempotent. Clears the downstream `build_task_settings` cache so a second `import_records` call in the same process sees fresh catalogs.
- **`api.annotation_import.import_records`** calls `register_catalog_dir` once after settings resolve, before any catalog lookup. This is the only api entrypoint that transitively reads catalogs (`import → fan_out_records → build_task_settings`); setup/teardown/export/iaa don't touch catalogs.
- **`get_catalog` now validates against the live `CATALOGS` dict** and raises `ValueError` with the supported set on miss. Replaces the old `SUPPORTED_LOCALES` frozenset (which was a snapshot taken at import - would go stale the moment a user registered their catalog dir, so CLI input validation against it would reject any user-registered locale).
- **`parse_locale` becomes a pure normaliser** (no validation). Validation lives at the data source (`get_catalog`) so it always sees user-registered catalogs without per-caller plumbing.

## Architectural decisions

- **Validation at data source, not api boundary.** Matches SOTA i18n (Rails I18n, i18next, Django) which throw at lookup time. `get_catalog` is called ~3 times per import (cached at `build_task_settings`), not a hot path, so the "validate at edges" cost-saving doesn't materialise. Coverage is automatic for any future caller.
- **Module-level mutable `CATALOGS` dict** (gettext-style) rather than a `CatalogRegistry` class. Pragmata is single-tenant CLI: no multi-instance, no concurrency. Class refactor would be over-engineering for current needs; revisit if/when library-mode or threading enters scope.
- **Per-entrypoint `register_catalog_dir` call** rather than hoisting into `Settings.resolve()`. Settings should be pure data; only one api entrypoint transitively needs catalogs, so per-entrypoint discipline is trivial here.

## Test plan

- [x] `tests/unit/core/annotation/test_locale_registry.py` - 5 tests covering: add new locale, override bundled on stem collision, idempotent re-registration, empty-dir noop, cache invalidation regression
- [x] `tests/unit/core/settings/test_annotation_settings.py::TestLocaleCatalogDirValidator` - 4 tests covering pydantic DirectoryPath behaviour
- [x] `tests/unit/cli/test_parsing.py::TestParseLocale` - updated to assert pure normalisation (no validation)
- [x] `uv run python -m pytest tests/unit` - 741 passed
- [x] `uv run mypy src/pragmata` clean

## References

- Closes the catalog-discovery gap raised in #196 review
- Independent of #209 (both touch `api/annotation_import.py` in different parts of the function; auto-merge clean either order)

### DIFF
--- a/src/pragmata/api/annotation_import.py
+++ b/src/pragmata/api/annotation_import.py
@@ -9,6 +9,7 @@ from pathlib import Path
 from pragmata.api._error_log import error_log
 from pragmata.core.annotation.client import resolve_argilla_client
 from pragmata.core.annotation.loaders import RecordInput, resolve_records
+from pragmata.core.annotation.locales.registry import register_catalog_dir
 from pragmata.core.annotation.record_builder import (
     RecordError,
     assign_partitions,
@@ -132,6 +133,8 @@ def import_records(
             "locale": locale,
         },
     )
+    if settings.locale_catalog_dir is not None:
+        register_catalog_dir(settings.locale_catalog_dir)
 
     api_key = api_key if isinstance(api_key, str) else resolve_api_key("argilla")
     client = resolve_argilla_client(settings.argilla.api_url, api_key)
--- a/src/pragmata/api/annotation_import.py
+++ b/src/pragmata/api/annotation_import.py
@@ -70,6 +70,7 @@ def import_records(
     calibration_min_submitted: int | None | Unset = UNSET,
     calibration_partition_seed: int | Unset = UNSET,
     locale: Locale | Unset = UNSET,
+    locale_catalog_dir: str | Path | None | Unset = UNSET,
 ) -> ImportResult:
     """Validate and fan out records to per-purpose Argilla annotation datasets.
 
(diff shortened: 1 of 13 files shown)


src/pragmata/api/annotation_import.py
```diff
@@ -144,6 +145,8 @@ def import_records(
             "locale": locale,
         },
     )
+    if settings.locale_catalog_dir is not None:
+        register_catalog_dir(settings.locale_catalog_dir)
```
saschagobel, line 154: Non-blocking: locale_catalog_dir should be exposed consistently through the import API and CLI, not only via settings/config.

saschagobel (approved):
Thanks, nice work! Left one non-blocking comment on providing overrides for locale_catalog_dir.

saschagobel (approved):

PR #221, opened 2026-05-25
feat: add settings models for eval workflows

saschagobel:
### Summary

Adds the initial eval settings models for training evaluator models, predicting labels, and scoring labeled evaluation data. The settings define the Pragmata-owned configuration surface while keeping tlmtc-specific training and prediction options behind explicit pass-through kwargs.

### Key changes

- Add `EvalTrainSettings`
  - Selects `base_dir`, `labeled_data_path`, `export_id`, and `task`
  - Supports `train_kwargs` for `train_tlmtc`-specific options
- Add `EvalPredictSettings`
  - Selects `base_dir`, `unlabeled_data_path`, `evaluator_run_id`, and `task`
  - Supports `predict_kwargs` for `predict_tlmtc`-specific options
- Add `EvalScoreSettings`
  - Selects `base_dir`, `score_id`, `labeled_input_path`, `prediction_run_id`, and `task`
  - Keeps scoring independent of tlmtc
- Reuse the existing annotation `Task` enum for retrieval, grounding, and generation

**Tests**

- Add unit tests for eval settings construction and validation.
- Verify that undeclared top-level fields such as `run_id` and `label_source` are rejected.
- Verify that `train_kwargs` and `predict_kwargs` are accepted as pass-through dictionaries.
- Verify that default `score_id` values are distinct UUID-style hex strings.

### DIFF
--- a/tests/unit/core/settings/test_eval_settings.py
+++ b/tests/unit/core/settings/test_eval_settings.py
@@ -1,6 +1,7 @@
 """Unit tests for evaluation workflow settings."""
 
 from pathlib import Path
+from uuid import UUID
 
 import pytest
 from pydantic import BaseModel, ValidationError
@@ -12,7 +13,9 @@ from pragmata.core.settings.eval_settings import EvalPredictSettings, EvalScoreS
 def _assert_uuid_hex(value: str) -> None:
     """Assert that a value is a UUID4-style hex string."""
     assert len(value) == 32
-    assert int(value, 16) >= 0
+    parsed = UUID(hex=value)
+    assert parsed.version == 4
+    assert parsed.hex == value
 
 
 @pytest.mark.parametrize(
--- a/tests/unit/core/settings/test_eval_settings.py
+++ b/tests/unit/core/settings/test_eval_settings.py
@@ -1,0 +1,217 @@
+"""Unit tests for evaluation workflow settings."""
+
+from pathlib import Path
+
+import pytest
+from pydantic import BaseModel, ValidationError
+
+from pragmata.core.schemas.annotation_task import Task
+from pragmata.core.settings.eval_settings import EvalPredictSettings, EvalScoreSettings, EvalTrainSettings
+
+
+def _assert_uuid_hex(value: str) -> None:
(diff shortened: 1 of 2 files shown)


tests/unit/core/settings/test_eval_settings.py
```diff
@@ -0,0 +1,217 @@
...
+from pydantic import BaseModel, ValidationError
+
+from pragmata.core.schemas.annotation_task import Task
+from pragmata.core.settings.eval_settings import EvalPredictSettings, EvalScoreSettings, EvalTrainSettings
+
+
+def _assert_uuid_hex(value: str) -> None:
+    """Assert that a value is a UUID4-style hex string."""
+    assert len(value) == 32
+    assert int(value, 16) >= 0
```
lramir14: Here in the doc-string it is stated that the UUID has to be version 4 (UUID4) but it is assumed, not tested in the assertions. E.g: "parsed = UUID(hex=value) 
assert parsed.version == 4"

lramir14 (approved):
Thanks for the work. The implementation looks solid, and I only have one minor non-blocking comment: the docstrings define some precedence rules between labeled_data_path over export_id, and labeled_input_path over prediction_run_id; so I think it could be useful to add tests covering the case where both values are provided, to make the contract even more explicit.

saschagobel:
> Thanks for the work. The implementation looks solid, and I only have one minor non-blocking comment: the docstrings define some precedence rules between labeled_data_path over export_id, and labeled_input_path over prediction_run_id; so I think it could be useful to add tests covering the case where both values are provided, to make the contract even more explicit.

Good point on making the precedence contract explicit. I think I would keep that out of the settings tests, though: this module only validates and stores user-facing configuration values/general settings, and both fields are intentionally allowed to coexist at this layer. The precedence rule is applied by the downstream input-selection/orchestration logic, so I’d prefer to cover it there once that resolver/API path is implemented.

lramir14 (approved):
Thanks for the clarification, it makes sense to me. Also, thanks for tightening the UUID4 assertion. Looks good from my side.

PR #222, opened 2026-05-26
feat(cli): expose querygen runtime and throttle options

henrycgbaker:
## Summary

- wires six `QueryGenRunSettings` knobs to the `pragmata querygen gen-queries` Typer CLI that were previously dropped silently when passed via API kwargs.
- CLI users can now tune rate limiting, batch sizing, near-duplicate tolerance, and planning memory without hand-writing a config file / can script them.
- 3 clearly earn keep as per-run operator overrides (i.e. settings operators commonly want to override per run without editing YAML: batch size, near-duplicate tolerance, planning-memory feature flags); the other 3 are rate-limiting knobs that sit on the borderline scheduler knobs included for completeness.
- If we go with all of them this closes the CLI/Python-API parity gap: every keyword accepted by `pragmata.api.querygen.gen_queries` now has a matching CLI flag.

## Changes

- `src/pragmata/cli/commands/querygen.py`: adds six Typer options (`--requests-per-second`, `--check-every-n-seconds`, `--max-bucket-size`, `--batch-size`, `--near-duplicate-tolerance`, `--enable-planning-memory/--no-enable-planning-memory`) and forwards them to `querygen.gen_queries(...)` using the existing `UNSET if x is None else x` pattern.
- `tests/unit/cli/test_cli_querygen.py`: expands the omitted-options assertion from 19 to 25 forwarded kwargs; adds a test that round-trips each new flag through to the API call; adds a separate test asserting `--enable-planning-memory` toggles to `True`; updates the help-text test to cover the new options and pins `COLUMNS=200` so wider help output isn't truncated by the runner.

## Test plan

- [x] `uv run pytest tests/unit/cli/test_cli_querygen.py -x` — 7 passed
- [x] `uv run pytest tests/unit/cli/ -x` — 34 passed
- [x] `uv run ruff format` + `uv run ruff check` clean on both edited files
- [x] `uv run mypy src/pragmata/cli/commands/querygen.py` — no issues
- [x] CI: lint, type-check, full test suite

### DIFF
--- a/tests/unit/cli/test_cli_querygen.py
+++ b/tests/unit/cli/test_cli_querygen.py
@@ -1,5 +1,6 @@
 """Tests CLI command for synthetic query generation."""
 
+import re
 from pathlib import Path
 
 from typer.testing import CliRunner
@@ -8,6 +9,12 @@ from pragmata.api import UNSET
 from pragmata.cli.app import app
 
 runner = CliRunner()
+
+ANSI_ESCAPE_RE = re.compile(r"\x1b\[[0-9;]*m")
+
+
+def strip_ansi(text: str) -> str:
+    return ANSI_ESCAPE_RE.sub("", text)
 
 
 class _PreparedResult:
@@ -30,10 +37,24 @@ def test_querygen_command_registered() -> None:
     assert "querygen" in result.output
 
 
-def test_querygen_gen_queries_help_available() -> None:
-    result = runner.invoke(app, ["querygen", "gen-queries", "--help"])
+def test_querygen_gen_queries_help_available(monkeypatch) -> None:
+    monkeypatch.setenv("COLUMNS", "200")
+    result = runner.invoke(app, ["querygen", "gen-queries", "--help"], color=False)
+    output = strip_ansi(result.output)
 
     assert result.exit_code == 0
+    assert "Prepare a synthetic query generation run." in output
+    assert "--domains" in output
+    assert "--run-id" in output
+    assert "--planning-model-kwargs" in output
+    assert "--realization-model-kwargs" in output
+    assert "--requests-per-second" in output
+    assert "--check-every-n-seconds" in output
+    assert "--max-bucket-size" in output
+    assert "--batch-size" in output
+    assert "--near-duplicate-tolerance" in output
+    assert "--enable-planning-memory" in output
+    assert "--no-enable-planning-memory" in output
 
 
 def test_querygen_cli_delegates_to_public_api(monkeypatch) -> None:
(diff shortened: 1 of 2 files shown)


henrycgbaker:
The three rate-limiting flags - `--requests-per-second`, `--max-bucket-size`, `--check-every-n-seconds` = parameters of the LLM rate limiter. cf `--batch-size`, `--near-duplicate-tolerance`, and `--enable-planning-memory` = operator-facing tuning knobs we'd plausibly experiment with per run; rate-limiter settings are more typically pinned once per provider/account and rarely overridden per run.

Possible yardstick is that for cli flag to earn its keep if/when operator would want to override the field per run without editing YAML? If so first 3 flags clearly pass that bar; cf the 3 rate-limiter flags borderline. Still, they're cheap to add and harmless - but possible cleaner split could be:
- yes to `--batch-size`, `--near-duplicate-tolerance`, `--enable-planning-memory` as CLI flags
- Leave the three rate-limiter knobs YAML-only?

tests/unit/cli/test_cli_querygen.py
```diff
@@ -38,15 +31,9 @@ def test_querygen_command_registered() -> None:
 
 
 def test_querygen_gen_queries_help_available() -> None:
-    result = runner.invoke(app, ["querygen", "gen-queries", "--help"], color=False)
-    output = strip_ansi(result.output)
+    result = runner.invoke(app, ["querygen", "gen-queries", "--help"])
 
     assert result.exit_code == 0
```
lramir14, line 45: I think only assserting that result is 0 checks whether the command works but does not check whether the new options are in the help text. Shall we keep the small assertions for the new flags? e.g. --requests-per-second, --batch-size, --near-duplicate-tolerance, and --enable-planning-memory

tests/unit/cli/test_cli_querygen.py
```diff
@@ -10,12 +9,6 @@
 
 runner = CliRunner()
 
-ANSI_ESCAPE_RE = re.compile(r"\x1b\[[0-9;]*m")
-
-
-def strip_ansi(text: str) -> str:
```
lramir14, line 16: I see that this deletion goes together with the idea of not checking the help text output.

lramir14 (approved):
thanks for the work. I think the only non-blocking comments are the ones referred to the help text in the test section. Other than that all looks good.

lramir14 (approved):
thank you for your work! Looks good to me. I don't see any blocking issues here. Perhaps a suggestion, maybe clarify the help text for --near-duplicate-tolerance, since the direction of the parameter is easy to misunderstand. For example:
"Similarity threshold for blueprint deduplication. Candidates with similarity >= this value are removed; lower values are stricter, while 1.0 is the least aggressive setting". However, we can make this explicit in the documentation with more detail.

PR #223, opened 2026-05-27
feat(annotation): annotator-time constraint feedback widget

henrycgbaker:
**Stack:** #213 (merged) → #214 → **#223** → #215 → #216 → #217

**This PR (widget):** Render `LOGICAL_CONSTRAINTS` as an inline warn/block panel in the Argilla annotator UI, with severity sourced from `AnnotationSettings.constraint_severity` (#214). Constraints render right above the discard panel; `warn` shows an advisory, `block` disables submit until inputs are corrected.

> **Restructured:** this PR now carries the full widget (previously the bulk of the original #214), stacked on the settings layer #214. It also subsumes the locale fix that was this branch's original sole content, so the widget reads locale-invariant machine values from the start.

---

### Widget
- `core/annotation/constraints_field.html` (new): vanilla-JS Argilla CustomField that reads parent-DOM checkbox/label state per question, evaluates `LOGICAL_CONSTRAINTS` payload entries, and renders an inline panel above the form actions. MutationObserver + rAF-coalesced; signature-based change detection short-circuits no-op renders. Reads each label's machine value from its `for` attribute (`<question_name>_<machine_value>`, locale-invariant) rather than the localised display text, so constraints fire in non-EN locales (e.g. DE "Ja"/"Nein").
- `core/annotation/argilla_task_definitions.py`: `constraints_field()` builder, `_render_constraints_template()` substitutes the constraints/question-titles JSON; `WIDGET_FIELD_PLACEHOLDERS` consolidates the per-record placeholder pattern. `build_task_settings(settings, locale)` now takes settings (was no-arg, was `@functools.cache`d) so it can read severity from `settings.constraint_severity`.
- `core/annotation/logical_constraints.py`: `message` field on `LogicalConstraint` (annotator-facing text); `to_widget_payload(severity)` serialises a constraint for the JS widget (string `yes`/`no`, resolved severity).
- `core/annotation/record_builder.py`: writes both `discard_flow` and `constraints_panel` placeholders per record (Argilla skips a CustomField with no record value).

## UX
- `warn`: yellow panel, submit still allowed.
- `block`: red panel + disabled submit button (title attribute explains why); reverts when inputs change to resolve.
- Empty `LOGICAL_CONSTRAINTS[task]` ⇒ panel hidden automatically.

## Test plan
- [x] `uv run python -m pytest tests/unit` (984 passing)
- [ ] Live Argilla dataset render check — **re-verify post-merge.** The widget JS has no unit coverage and changed since the original check (reads the label `for`-attribute machine value; renders workspace-resolved severity once #216 lands), so the live render path is the real gate and was not re-run during the rebase.

### DIFF
--- a/src/pragmata/core/annotation/constraints_field.html
+++ b/src/pragmata/core/annotation/constraints_field.html
@@ -1,0 +1,293 @@
+<!DOCTYPE html>
+<html>
+<head></head>
+<body>
+<script>
+(function () {
+  // Substituted at dataset creation time by argilla_task_definitions.py.
+  // CONSTRAINTS shape: see LogicalConstraint.to_widget_payload in logical_constraints.py.
+  var CONSTRAINTS = $CONSTRAINTS_JSON;
+  var QUESTION_TITLES = $QUESTION_TITLES_JSON;
+
+  // Argilla DOM contract — verified against argilla-frontend v2.8.0 source.
+  // QuestionHeader stamps title onto `<div class="title-area" aria-label=...>`;
+  // SingleLabel wraps the whole component in `<div class="wrapper">`;
+  // LabelSelection renders chips as `<input type=checkbox>` + `<label class="label-text">`
+  // with `.label-active` added when selected; QuestionsForm puts the submit
+  // button (`<button type=submit class="button--submit">`) inside `.footer-form`.
+  // Any of these moving will need re-verification on Argilla upgrades.
+  var pdoc = window.parent.document;
+  var SUBMIT_SEL = '.button--submit, button[type="submit"]';
+  var FOOTER_SEL = '.footer-form';
+  var INJECTED_PANEL_ID = '__constraints_injected_panel';
+  var INJECTED_STYLE_ID = '__constraints_injected_style';
+
+  function ensureNode(id, parent, build) {
+    var existing = pdoc.getElementById(id);
+    if (existing && existing.isConnected) return existing;
+    if (!parent) return null;
+    var node = build();
+    node.id = id;
+    parent.appendChild(node);
+    return node;
+  }
+
+  function ensureInjectedStyle() {
+    return ensureNode(INJECTED_STYLE_ID, pdoc.head, function () {
+      var s = pdoc.createElement('style');
+      s.textContent =
+        '#' + INJECTED_PANEL_ID + '{border-radius:6px;padding:10px 12px;' +
+        'font-family:system-ui,sans-serif;font-size:13px;line-height:1.4;margin:0 0 12px 0}' +
+        '#' + INJECTED_PANEL_ID + '[hidden]{display:none}' +
+        '#' + INJECTED_PANEL_ID + '.warn{border:2px solid #d4a017;background:#fff8e1;color:#5d4400}' +
+        '#' + INJECTED_PANEL_ID + '.block{border:2px solid #d9534f;background:#fff5f5;color:#6e0a0a}' +
+        '#' + INJECTED_PANEL_ID + ' h4{margin:0 0 6px 0;font-size:13px}' +
+        '#' + INJECTED_PANEL_ID + ' ul{margin:0;padding-left:18px}' +
+        '#' + INJECTED_PANEL_ID + ' li{margin-top:4px}' +
+        // Trailing space in a CSS \HEX escape is the terminator (consumed),
+        // so add literal spaces after it for the actual gap.
+        '#' + INJECTED_PANEL_ID + ' li.warn::marker{content:"\\26A0   "}' +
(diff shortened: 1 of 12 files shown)


src/pragmata/core/annotation/constraints_field.html
```diff
@@ -0,0 +1,293 @@
...
+  // understands why submit isn't responding.
+  function applySubmitGate(hasBlocked) {
+    var btns = pdoc.querySelectorAll(SUBMIT_SEL);
+    for (var i = 0; i < btns.length; i++) {
+      var b = btns[i];
+      if (hasBlocked) {
+        if (!b.dataset.constraintBlocked) {
+          b.dataset.constraintBlocked = '1';
+          b.dataset.prevTitle = b.title || '';
+          b.disabled = true;
```
ddimmery, line 191: Does this also disable using keyboard shortcuts?

tests/unit/core/annotation/test_argilla_task_definitions.py
```diff
@@ -290,13 +291,66 @@ def test_catalog_drives_field_title(self, monkeypatch):
...
+        assert field.advanced_mode is True
+        assert field.required is False
+
+    def test_template_substitutions_are_resolved(self):
+        # The widget receives both placeholders — verify nothing was left
+        # unsubstituted (would be a hard-to-debug runtime JS error).
+        for settings in (_RETRIEVAL, _GROUNDING):
+            template = _get_field(settings, "constraints_panel").template
+            assert "@@CONSTRAINTS_JSON" not in template
+            assert "@@QUESTION_TITLES_JSON" not in template
```
ddimmery: Are these `@@...` or `$...`? Looks like the latter to me, but maybe there's more complexity here.

src/pragmata/core/annotation/record_builder.py
```diff
@@ -648,7 +648,7 @@ def fan_out_records(
         else:
             min_submitted = resolved.production_min_submitted
         locale = resolved.locale
-        task_settings_map = build_task_settings(locale)
+        task_settings_map = build_task_settings(settings, locale)
```
ddimmery: This is inside the `(task, calibration)` loop. The docstring for `build_task_settings` says this shouldn't be used like this? It seems like `locale` depends on `task`, so maybe this is necessary?

src/pragmata/core/annotation/constraints_field.html
```diff
@@ -0,0 +1,293 @@
...
+
+  // Argilla DOM contract — verified against argilla-frontend v2.8.0 source.
+  // QuestionHeader stamps title onto `<div class="title-area" aria-label=...>`;
+  // SingleLabel wraps the whole component in `<div class="wrapper">`;
+  // LabelSelection renders chips as `<input type=checkbox>` + `<label class="label-text">`
+  // with `.label-active` added when selected; QuestionsForm puts the submit
+  // button (`<button type=submit class="button--submit">`) inside `.footer-form`.
+  // Any of these moving will need re-verification on Argilla upgrades.
+  var pdoc = window.parent.document;
+  var SUBMIT_SEL = '.button--submit, button[type="submit"]';
```
ddimmery, line 20: This selects _all_ submit buttons docwide. Not a big problem (I assume there's only one), but worth noting.

src/pragmata/core/annotation/constraints_field.html
```diff
@@ -0,0 +1,293 @@
...
+<head></head>
+<body>
+<script>
+(function () {
+  // Substituted at dataset creation time by argilla_task_definitions.py.
+  // CONSTRAINTS shape: see LogicalConstraint.to_widget_payload in logical_constraints.py.
+  var CONSTRAINTS = $CONSTRAINTS_JSON;
+  var QUESTION_TITLES = $QUESTION_TITLES_JSON;
+
+  // Argilla DOM contract — verified against argilla-frontend v2.8.0 source.
```
ddimmery, line 12: v2.8.0 mismatches with the import in `record_builder.py` which says we're pinned at v2.6.0, but `pyproject.toml` just says `argilla>=2.0,<3.0`. Not sure what to believe!

ddimmery (approved):
No major issues, just flagging some things.

The text in these boxes are hardcoded to English. Given that the other UI is localized, this could be a little weird. Not a fix to be made now, but maybe something to flag somewhere in docstrings/comments.

src/pragmata/core/annotation/constraints_field.html
henrycgbaker: keeping button[type="submit"] as fallback so gate works if Argilla renames (vs alternatives potentially fragile); low risk as only record submit is type="submit", and we show one record at a time

PR #224, opened 2026-05-27
feat: add eval input dataframe schemas

saschagobel:
### Summary

Adds task-specific dataframe input schemas for the eval tool. The new contracts validate normalized eval input data for training, prediction, and scoring before downstream transformation for tlmtc execution.

## Key changes

- Add Pandera-based eval input schemas for:
  - retrieval train/score inputs
  - grounding train/score inputs
  - generation train/score inputs
  - retrieval prediction inputs
  - grounding prediction inputs
  - generation prediction inputs
- Add validation functions for:
  - `validate_eval_train_frame(...)`
  - `validate_eval_score_frame(...)`
  - `validate_eval_predict_frame(...)`
- Preserve additional identifier columns by keeping schemas non-strict
- Reject labeled prediction inputs, including task-specific label columns and generic `label_*` columns
- Add pandas and Pandera as core dataframe validation dependencies
- Refresh `uv.lock`

### Design notes

These schemas validate Pragmata-owned eval input contracts, not tlmtc artifacts. Train and score use the same labeled task-specific contracts, while predict uses unlabeled task-specific contracts and performs an explicit label-column guard in the validation function.

### Tests

- Added unit tests for `src/pragmata/core/schemas/eval_input.py`
- Added unit coverage for valid inputs, invalid structures, invalid values, label coercion, extra-column preservation, score reuse of train validation, and unlabeled prediction enforcement


src/pragmata/core/schemas/eval_input.py
```diff
@@ -0,0 +1,255 @@
...
+        task: Annotation task that determines the input contract to apply.
+
+    Returns:
+        Validated dataframe with all original columns preserved.
+
+    Raises:
+        EvalInputSchemaError: If the input is not a dataframe or violates the
+            task-specific scoring contract.
+    """
+    return validate_eval_train_frame(df, task=task)
```
lramir14: Minor comment: since validate_eval_score_frame() delegates to validate_eval_train_frame(), invalid scoring inputs currently raise an error mentioning the “training data contract.” The schema reuse makes sense, but a score-specific error message might be clearer for users who are validating score inputs.

lramir14 (approved):
Thank you for the nice work. I only have one comment. Currently, `validate_eval_predict_frame()` checks whether prediction inputs contain label columns, but it only rejects label columns for the selected task or columns starting with `label_`. Since the prediction schemas are `strict=False`, cross-task eval label columns could still be accepted as extra columns. For example, a retrieval prediction frame could contain generation/grounding labels such as `proper_action`, `support_present`, or `fabricated_source` and still pass validation.

If the intended contract is that prediction inputs must be fully unlabeled, should we reject the union of all known eval label columns across tasks, rather than only `LABEL_COLUMNS_BY_TASK[task]`?

saschagobel:
> Thank you for the nice work. I only have one comment. Currently, `validate_eval_predict_frame()` checks whether prediction inputs contain label columns, but it only rejects label columns for the selected task or columns starting with `label_`. Since the prediction schemas are `strict=False`, cross-task eval label columns could still be accepted as extra columns. For example, a retrieval prediction frame could contain generation/grounding labels such as `proper_action`, `support_present`, or `fabricated_source` and still pass validation.
> 
> If the intended contract is that prediction inputs must be fully unlabeled, should we reject the union of all known eval label columns across tasks, rather than only `LABEL_COLUMNS_BY_TASK[task]`?

Good catch on the ambiguity. The current behavior is intentional, though: prediction inputs must be unlabeled for the selected task, but they may carry arbitrary pass-through columns, including labels or predictions from other eval tasks. This supports workflows where the same example set is enriched across retrieval, grounding, and generation prediction steps and is explicitly supported in tlmtc. I clarified the docstring/error message.

lramir14 (approved):
Great enhancements. I think it is very clear right now. Approved.

PR #225, opened 2026-05-27
fix(annotation): IAA pivot keys retrieval rows by (record_uuid, chunk_id)

henrycgbaker:
Standalone bug fix that pairs naturally with the per-item-calibration stack (#226 → #227 → #228 → #229 → #230) but ships independently (no functional dependency).

---

## Scope

- `core/annotation/iaa_runner.py`: `_item_key` for retrieval rows now composes `(record_uuid, chunk_id)` instead of `record_uuid` alone. Grounding and generation keys are unchanged (one item per record).
- `tests/unit/core/annotation/test_iaa_runner.py`: pin the per-chunk pivot key behaviour and exercise the multi-chunk path.

## Why

Retrieval annotation items are one-per-chunk on the Argilla side (each chunk is its own record). Keying the IAA pivot by `record_uuid` alone conflated annotator agreement across all chunks of the same record into a single slot, inflating apparent agreement on records with multiple chunks. The fix realigns the pivot with the actual annotation unit.

## Test plan
- [x] `uv run python -m pytest tests/unit/core/annotation/test_iaa_runner.py`


lramir14 (approved):
Thanks for the work and nice implementation. I only have one comment or clarification. In the tests what happens when we have missing or empty `chunk_id` in retrieval rows. If the retrieval export/schema guarantees `chunk_id` is always present and non-empty, then this is fine. If not, `_item_key` could still collapse rows as `record_uuid:None` or `record_uuid:`. Not blocking from my side, but it may be worth either relying explicitly on the schema guarantee or adding a defensive check/error for missing `chunk_id`.

lramir14 (approved):
Thanks for the quick implementation. Looks good to me. I don't see any blocking issue. Approved.

PR #226, opened 2026-05-27
docs(annotation): ADR-0012 per-item calibration partition

henrycgbaker:
**Stack (4 PRs):** #226 → #227 → #228 → #229 (companion to standalone IAA fix #225)

**Chain goal:** Switch calibration vs production assignment from per-`record_uuid` to per annotation item (chunk for retrieval, record for grounding / generation), with inheritable per-task fraction and absolute cap.

**This PR (1/4):** Lands the ADR and the matching design-doc section before the implementation, so subsequent PRs can reference the decision and the schema shape.

---

## Scope

- `docs/decisions/0012-per-item-calibration-partition.md` (new): the decision, the option-set considered, Kish design-effect arithmetic justifying per-item over per-record under the existing naive Krippendorff bootstrap, strategic positioning vs LLM-judge mainstream + industry annotation services, consequences, out-of-scope future work, and references.
- `docs/decisions/README.md`: index entry for the new ADR.
- `docs/design/annotation-import-pipeline.md`: new "Per-record manifest schema" section describing the `PartitionManifestEntry` shape that #227 implements (`grounding_generation_calibration` + `retrieval_chunk_calibration`).

## Test plan
- [x] Docs only - no test changes.

PR #227, opened 2026-05-27
feat(annotation): per-item PartitionManifestEntry + per-item calibration assignment

henrycgbaker:
**Stack (4 PRs):** #226 → #227 → #228 → #229 (companion to standalone IAA fix #225)

**Chain goal:** Switch calibration vs production assignment from per-`record_uuid` to per annotation item (chunk for retrieval, record for grounding / generation), with inheritable per-task fraction and absolute cap.

**This PR (2/4):** Per-item partition logic + schema reshape + per-task fraction cascade. Ships the user-visible feature without the absolute cap (#229) or the API-hardening pass (#228).

---

### Schema
- `core/schemas/annotation_import.py`: `PartitionManifestEntry` replaces `calibration: bool` with `grounding_generation_calibration: dict[Task, bool]` (keyed by task) and `retrieval_chunk_calibration: dict[str, bool]` (keyed by `chunk_id`). Per-task `calibration_fraction_at_import` provenance.

### Partition logic
- `core/annotation/record_builder.py`: `assign_partitions` buckets units via per-(task, unit) digest `hash(seed || task || unit_id)` against `fraction * 2^32` so per-task draws are statistically independent. `_enumerate_units` resolves the per-task unit list, `_write_unit` routes the bool back into the right per-record dict.
- Returns `dict[str, PartitionManifestEntry]` as before; the `PartitionResult` bundle is extracted in #228.

### Settings cascade
- `core/settings/annotation_settings.py`: `calibration_fraction` is now inheritable across deployment / workspace / task scopes via the `Inherit` sentinel pattern established in #206. `_check_calibration_topology` walks per-(ws, task) using resolved values. The `_inherit` walk is hoisted into a single helper so `resolved_task` no longer repeats the chain four times.

### API / CLI
- `api/annotation_import.py`: `ImportResult` per-task dicts (`calibration_count`, `production_count`, `calibration_fraction`, `realised_calibration_fraction`) replace the scalar fields.
- `cli/commands/annotation.py`: CLI output prints one line per task.

### Tests
- Per-chunk routing independence, per-task seed independence, inheritance cascade, and per-(ws, task) topology validation. Manifest IO round-trip with the new shape.

## Breaking change

Pre-v0 on-disk manifests with `calibration: bool` no longer load. Pragmata is pre-1.0; affected workspaces re-bootstrap by deleting `partition.meta.json` and re-importing.

## Why

Per ADR-0012 (#226): per-item partition is the natural fit for the existing naive Krippendorff bootstrap and matches Argilla's per-item record creation. Per-task fraction cascade lets operators right-size each task independently (e.g. retrieval at 5%, grounding at 20%).

## Test plan
- [x] `uv run python -m pytest tests/unit` (770 passing in this branch's worktree)


src/pragmata/api/annotation_import.py
```diff
@@ -182,37 +193,73 @@ def import_records(
...
+    """
+    fraction: dict[Task, float] = {}
+    for ws_name, ws in settings.workspaces.items():
+        for task in ws.tasks:
+            if task in fraction:
+                continue
+            fraction[task] = settings.resolved_task(ws_name, task).calibration_fraction
+    for task in Task:
+        fraction.setdefault(task, 0.0)
+    return fraction
```
saschagobel: Blocking: this moves annotation partition/reporting business logic into the API layer.
`_total_units_per_task()` knows the task-specific unit model (retrieval counts chunks, grounding/generation count records), and `_resolve_per_task_fraction()` walks the workspace/task topology to resolve per-task calibration settings. Those are core annotation/settings semantics, not API orchestration. This also makes `api/annotation_import.py` inspect `PartitionManifestEntry` internals directly.

Per ADR-0007 and `docs/design/packaging-invocation-surface.md`, `api/` should orchestrate core functions, while implementation logic belongs in `core/`. Please move this summary/counting logic into core, ideally next to `count_calibration_per_task()` or as a small core partition summary helper, and have the API call that.

src/pragmata/api/annotation_import.py
```diff
@@ -182,37 +193,73 @@ def import_records(
...
-        production_count = len(assignments) - calibration_count
-        assigned_total = calibration_count + production_count
-        realised = calibration_count / assigned_total if assigned_total else 0.0
+        calibration_count = count_calibration_per_task(assignments.values())
+        total_count = _total_units_per_task(assignments.values())
+        production_count = {task: total_count[task] - calibration_count[task] for task in Task}
+        realised_fraction = {
+            task: (calibration_count[task] / total_count[task]) if total_count[task] else 0.0 for task in Task
+        }
+        configured_fraction = _resolve_per_task_fraction(settings)
```
saschagobel: Non-blocking: the new `ImportResult` shape makes sense, but this result assembly would be cleaner if the partition stats came from core as one summary object/helper.

Right now the API layer computes calibration counts, total units, production counts, realised fractions, and configured fractions directly. That spreads partition/reporting semantics into `import_records()`. A core helper like `summarize_partitions(assignments.values(), settings)` would let this function stay focused on orchestration and simply map the summary into `ImportResult`.

src/pragmata/core/annotation/record_builder.py
```diff
@@ -90,31 +92,31 @@ def derive_record_uuid(pair: QueryResponsePair) -> str:
...
+    metadata: dict = {
+        "record_uuid": record_uuid,
+        "chunk_id": chunk.chunk_id,
+        "doc_id": chunk.doc_id,
+        "chunk_rank": chunk.chunk_rank,
+    }
+    if pair.language is not None:
+        metadata["language"] = pair.language
+    return rg.Record(
+        id=f"ret-{record_uuid}-{index}",
```
saschagobel: Question: should retrieval `Record.id` now be based on `(record_uuid, chunk_id)` rather than `(record_uuid, enumerate_index)`?

This PR makes retrieval partition identity `(record_uuid, chunk_id)`: `derive_record_uuid()` is chunk-order-independent, the manifest routes retrieval by `chunk_id`, and duplicate chunk IDs are rejected. But the Argilla record ID still depends on the current list position.

If the same logical record is reimported with chunks in a different order, the manifest still treats it as the same record/chunks, but Argilla IDs change per chunk (`ret-<uuid>-0` can point to a different `chunk_id`). With per-chunk calibration, this can also leave stale records across calibration/production datasets when chunks route independently.

Is that expected, or should this ID use a stable chunk-derived component?

src/pragmata/core/annotation/record_builder.py
```diff
@@ -218,36 +223,117 @@ def assign_partitions(
...
+    new_pairs: dict[str, QueryResponsePair] = {}
+    result: dict[str, PartitionManifestEntry] = {}
     for pair in pairs:
         rid = derive_record_uuid(pair)
         existing = manifest.assignments.get(rid)
         if existing is not None:
-            assignments[rid] = existing.calibration
+            result[rid] = existing
+        else:
+            new_pairs[rid] = pair
```
saschagobel: Question: what is the intended behavior if the workspace/task topology changes for an existing `dataset_id`?

This branch reuses an existing manifest entry unchanged. If a record was first imported under a partial topology, e.g. retrieval-only, its manifest entry will not contain grounding/generation flags. If the same `dataset_id` is later imported with grounding/generation enabled, `_build_batches()` will default those missing flags to production via `.get(..., False)`, so the newly active tasks bypass the configured calibration fraction for existing records.

Is topology intended to be immutable per `dataset_id`? If so, should the manifest store/check a topology fingerprint and reject mismatches? If topology expansion is allowed, should missing task assignments be backfilled deterministically instead of defaulting to production?

src/pragmata/core/settings/annotation_settings.py
```diff
@@ -37,6 +40,7 @@ class TaskSettings(BaseModel):
     production_min_submitted: PositiveInt | Inherit = INHERIT
     calibration_min_submitted: PositiveInt | None | Inherit = INHERIT
     locale: Locale | Inherit = INHERIT
+    calibration_fraction: float | Inherit = INHERIT
```
saschagobel: Blocking: nested `calibration_fraction` overrides are not range-validated (also applies to line 59).

The deployment-level field uses `Field(0.1, ge=0.0, le=1.0)`, but the task/workspace fields are plain `float | Inherit`. A YAML override like `workspaces.retrieval.calibration_fraction: 1.5` or a task-level `-0.1` will validate, then `resolved_task()` returns that invalid value and `assign_partitions()` consumes it directly. Since this setting is now inheritable, all scopes should enforce the same fraction contract.

tests/integration/test_annotation_calibration.py
```diff
@@ -81,10 +81,10 @@ def test_explicit_fraction_creates_both_datasets(client: rg.Argilla, base_dir: P
     assert client.datasets(prod_name, workspace="retrieval") is not None
     assert client.datasets(cal_name, workspace="retrieval") is not None
 
-    assert result.calibration_count > 0
-    assert result.production_count > 0
-    assert result.calibration_count + result.production_count == len(records)
-    assert result.calibration_fraction == 0.5
+    assert all(result.calibration_count[t] > 0 for t in Task)
+    assert all(result.production_count[t] > 0 for t in Task)
+    assert all(result.calibration_count[t] + result.production_count[t] == len(records) for t in Task)
```
saschagobel: Non-blocking: these integration assertions still only exercise the one-chunk-per-record case.

Retrieval now counts annotation items as chunks, while grounding/generation count records. Because `_make_raw()` always emits exactly one chunk, `result.calibration_count[t] + result.production_count[t] == len(records)` holds for every task and would not catch retrieval accidentally still being counted per record. It would be useful to include at least one multi-chunk record here and assert retrieval totals against `sum(len(r["chunks"]) for r in records)` while grounding/generation remain `len(records)`.

tests/unit/core/annotation/test_fan_out_records.py
```diff
@@ -172,6 +200,39 @@ def test_skips_tasks_not_in_workspaces_topology(self, patched_create_dataset, ca
...
+                retrieval_chunk_calibration={"ca": True, "cb": False, "cc": True},
+                import_id="test",
+                calibration_fraction_at_import={t: 0.5 for t in Task},
+                assigned_at=now,
+            )
+        }
+
+        fan_out_records(client, records=[pair], settings=_settings(), assignments=assignments)
+
+        # Both retrieval datasets must exist (some chunks in cal, some in prod).
```
saschagobel, line 232: Non-blocking: this test should assert the actual chunk routing, not only that both retrieval datasets exist.

The important behavior is that `ca` and `cc` are logged to retrieval calibration and `cb` is logged to retrieval production. As written, the test would still pass if `_build_batches()` put the wrong chunk into each bucket as long as both datasets were touched.

tests/unit/core/annotation/test_import.py
```diff
@@ -68,25 +74,25 @@ def test_returns_hex_string(self) -> None:
...
 
 
 class TestBuildRetrievalRecords:
     def test_one_record_per_chunk(self) -> None:
         pair = _make_pair()
-        records = build_retrieval_records(pair, _UUID)
+        records = _build_retrieval_records(pair, _UUID)
         assert len(records) == len(pair.chunks)
 
     def test_record_ids_use_enumerate_index(self) -> None:
```
saschagobel: Question: should this test still assert enumerate-index-based retrieval IDs?

With this PR, retrieval annotation identity is `(record_uuid, chunk_id)`: `derive_record_uuid()` is chunk-order-independent, and partitioning/fan-out route retrieval by `chunk_id`. This test now locks in `ret-<uuid>-0` / `ret-<uuid>-1`, which is the unstable identity concern from `build_retrieval_record_for_chunk()`. If retrieval `Record.id` should be stable per chunk, this test should assert a chunk-derived ID instead.

saschagobel (changes requested):
Thanks for the substantial work here. I'm requesting changes for the API/core boundary issue and missing validation on inherited calibration_fraction, and I left the remaining comments as non-blocking questions/suggestions around partition summaries, retrieval ID stability, manifest reuse semantics, and test coverage.

src/pragmata/core/annotation/record_builder.py
henrycgbaker: this is a bug - switching to chunk-derived id so Argilla id matches partition id

src/pragmata/core/annotation/record_builder.py
henrycgbaker: I see 3(?) options here:
- 1. full task schema import - i.e even on missing tasks (not sure if there's a clean solution here, feels unncessarily cumbersome/goes against users' configs in settings.workspace) <- reject
- 2. backfill (ie run per-task assignments for missing tasks -> append to that)
- 3. fingerprint -> reject task expansion

I'd suggest to go with 2. Doesn't feel like much additional work, is nice-to-have (I can see a usecase when users want to make ad hoc changes to settings.workspace

saschagobel (approved):
Thanks for addressing the review feedback. Approving this from a code-review perspective, with the remaining note that the current mypy failure in `record_builder.py` needs to be fixed before merge.

saschagobel (approved):

PR #228, opened 2026-05-27
refactor(annotation): extract PartitionResult bundle + collapse count helpers

henrycgbaker:
**Stack (4 PRs):** #226 → #227 → #228 → #229 (companion to standalone IAA fix #225)

**Chain goal:** Switch calibration vs production assignment from per-`record_uuid` to per annotation item (chunk for retrieval, record for grounding / generation), with inheritable per-task fraction and absolute cap.

**This PR (3/4):** Pure refactor with no behaviour change. Tightens the api / core seam introduced in #227 so per-task fraction reporting and per-pair `record_uuid` derivation are each computed exactly once.

---

## Scope

- `core/annotation/record_builder.py`: introduce `PartitionResult(assignments, pairs_by_rid, calibration_fraction)`. `assign_partitions` now returns this bundle; `fan_out_records(client, settings, *, partition)` consumes it and `_build_batches(pairs_by_rid, assignments)` iterates the rid map instead of recomputing `derive_record_uuid(pair)` per pair.
- `core/annotation/record_builder.py`: replace `count_calibration_per_task` (returning `dict[Task, int]`) with `count_units_per_task` returning a `TaskUnitCounts(calibration, total)` frozen dataclass so the api layer gets both tallies in a single pass.
- `api/annotation_import.py`: drop `_total_units_per_task` and `_resolve_per_task_fraction`. The api now reads `partition.calibration_fraction` directly and uses the merged `count_units_per_task` helper.

## Why

After #227 lands, the api layer recomputes per-task fraction via its own workspace walk (`_resolve_per_task_fraction`), and `_build_batches` recomputes `record_uuid` for every pair even though `assign_partitions` already derived it. Both are wasted work; the natural fix is to surface the precomputed values via a bundle return type. This also unblocks the cap-related fields in #229 - `PartitionResult` is the obvious place to carry per-task cap resolution.

Reviewer sanity check: assertions in `test_partition.py` move from `result[rid]` to `partition.assignments[rid]`; the values they check are unchanged.

## Test plan
- [x] `uv run python -m pytest tests/unit` (770 passing)

### DIFF
--- a/src/pragmata/core/annotation/record_builder.py
+++ b/src/pragmata/core/annotation/record_builder.py
@@ -443,24 +443,6 @@ def count_units_per_task(entries: Iterable[PartitionManifestEntry]) -> TaskUnitC
     return TaskUnitCounts(calibration=calibration, total=total)
 
 
-def _configured_fraction_per_task(settings: AnnotationSettings) -> dict[Task, float]:
-    """Resolve per-task ``calibration_fraction`` for reporting.
-
-    Picks the first workspace that owns each task. Returns ``0.0`` for any task
-    missing from the workspaces topology, matching the assignment behaviour for
-    absent tasks.
-    """
-    fraction: dict[Task, float] = {}
-    for ws_name, ws in settings.workspaces.items():
-        for task in ws.tasks:
-            if task in fraction:
-                continue
-            fraction[task] = settings.resolved_task(ws_name, task).calibration_fraction
-    for task in Task:
-        fraction.setdefault(task, 0.0)
-    return fraction
-
-
 @dataclass(frozen=True)
 class PartitionSummary:
     """Per-task partition reporting derived from a set of manifest entries.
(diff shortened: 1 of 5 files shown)


src/pragmata/core/annotation/record_builder.py
```diff
@@ -233,13 +233,34 @@ def write_partition_manifest(path: Path, manifest: PartitionManifest) -> None:
...
+        assignments: record_uuid → PartitionManifestEntry, covering every
+            input pair (existing + newly assigned).
+        pairs_by_rid: record_uuid → QueryResponsePair for every input pair.
+        calibration_fraction: Per-task resolved fraction in force this run.
+            ``0.0`` for tasks absent from the workspaces topology.
+    """
+
+    assignments: dict[str, PartitionManifestEntry]
+    pairs_by_rid: dict[str, QueryResponsePair]
+    calibration_fraction: dict[Task, float]
```
saschagobel, line 254: Question: the PR body says the API now reads `partition.calibration_fraction` directly. The counts/realised fraction are derived from `partition.assignments`, but `ImportResult.calibration_fraction` still comes from `summary.configured_fraction`, which is recomputed from `settings` in `summarize_partitions()`. Am I missing where `PartitionResult.calibration_fraction` is consumed, or should `summarize_partitions()`/the API use the value already computed by `assign_partitions()`?

saschagobel (approved):
The refactor looks behavior-preserving overall; I left one minor question about PartitionResult.calibration_fraction being populated while the API still appears to use summary.configured_fraction.

src/pragmata/core/annotation/record_builder.py
henrycgbaker: PartitionResult.calibration_fraction wasn't consumed -> fixed in 6275fbd:
- summarize_partitions now takes the precomputed fraction from assign_partitions rather than re-resolving from settings, so the bundled value is the single source of truth
- no behaviour change - task-uniqueness enforced by a validator, so the two resolution paths always agree

saschagobel (approved):

PR #230, opened 2026-05-27
fix(annotation): topology-vs-manifest guard + bounded fraction at all scopes

henrycgbaker:
**Follow-up to stack:** #226 → #227 → #228 → #229 (companion to standalone IAA fix #225)

**This PR:** Two small hardening fixes uncovered by audit on the per-item partition stack, plus per-chunk integration coverage.

---

## Scope

- `core/annotation/record_builder.py`: `fan_out_records` now raises `RuntimeError` when the partition manifest assigns records to a task absent from the current workspaces topology. Previously it warned and continued, silently dropping data on disk if an operator reshaped their workspaces between imports.
- `core/settings/annotation_settings.py`: workspace- and task-scoped `calibration_fraction` now constrained to `[0.0, 1.0]` via a shared `CalibrationFraction = Annotated[float, Field(ge=0.0, le=1.0)]` alias. The deployment-scope guard already had this range check; the new alias closes the gap so a YAML setting of `workspaces.x.calibration_fraction: 1.5` is rejected at load.
- `api/annotation_import.py`: clarify the manifest-write-ordering comment - a retry under a binding cap is order-dependent (the cap-vs-existing logic is documented in `test_cap_under_split_imports_is_order_dependent_by_design`).
- Tests: `test_records_for_task_absent_from_topology_raises` replaces the prior warn-and-skip test. Locale-mismatch test gets a complete 3-task topology so the topology guard isn't tripped incidentally.
- Integration: `test_per_chunk_retrieval_routing_to_argilla` verifies per-chunk independence end-to-end against a live Argilla dataset (assignments in the manifest match the chunk_ids actually present in each retrieval dataset).

## Why

1. The topology-vs-manifest path was silent data loss. If a user removed a task from their workspace topology while old manifest entries for that task existed, `fan_out_records` quietly skipped them with a `logger.warning`. Promoting to `RuntimeError` forces the operator to address the inconsistency (either re-add the task or clear the affected partition manifest).

2. The deployment-level `calibration_fraction` field had `Field(ge=0.0, le=1.0)`, but the workspace and task variants were typed `float | Inherit` with no range check. A YAML setting of `workspaces.x.calibration_fraction: 1.5` validated, then resolved to a value `>= 1.0`, then routed 100% to calibration. The bounded alias closes that gap consistently across scopes.

## Test plan
- [x] `uv run python -m pytest tests/unit` (770 passing)
- [x] `uv run python -m pytest tests/integration/test_annotation_calibration.py -m "integration and annotation"` (8/9 passing; 1 pre-existing failure in `TestPerWorkspaceMinSubmitted::test_workspace_carve_out_creates_correct_dataset_min_submitted` exists on `main` already)

### DIFF
--- a/src/pragmata/api/annotation_import.py
+++ b/src/pragmata/api/annotation_import.py
@@ -209,7 +209,9 @@ def import_records(
 
         # Manifest is written only after fan-out succeeds. On failure, the
         # in-memory assignments are dropped; a retry with the same corpus + seed
-        # re-derives identical buckets (deterministic per-unit hashing).
+        # re-derives identical buckets under the fraction-only path. Under a
+        # binding cap, retries are order-dependent (see
+        # ``test_cap_under_split_imports_is_order_dependent_by_design``).
         dataset_counts = fan_out_records(client, settings, partition=partition)
         write_partition_manifest(import_paths.partition_manifest, manifest)
 
--- a/src/pragmata/core/annotation/record_builder.py
+++ b/src/pragmata/core/annotation/record_builder.py
@@ -684,8 +684,15 @@ def fan_out_records(
             continue
         ws_base = task_to_ws.get(task)
         if ws_base is None:
-            logger.warning("Task %r not in workspaces topology - skipping", task)
-            continue
+            # Manifest entries exist for a task absent from the current topology -
+            # silently skipping would drop data on disk. Force the operator to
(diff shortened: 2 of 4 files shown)


henrycgbaker:
Folded into #229 — the audit-fix commit is now part of that PR's history (cherry-picked cleanly onto the same tip, verified via diff audit + full test suite).

PR #233, opened 2026-05-27
feat(annotation): render collapsible widget content as markdown + HTML

henrycgbaker:
## Summary

Brings the collapsible widget's record-content rendering in line with the main `rg.TextField` rendering (see #232). Pipeline content (e.g. publikationsbot output) is markdown with embedded HTML/XML structure tags; we want the widget to render it the same way Argilla's `use_markdown=True` renderer renders the main TextField.

Two commits on this branch (will squash on merge):

1. `textContent` → `innerHTML` on the `#content` slot.
2. Add `marked.js` via CDN, run record content through `marked.parse(text)` before assigning to `innerHTML`.

Together: widget renders headings, bold, numbered lists, line breaks, *and* embedded HTML — same as the main TextField.

**Trust assumption:** record content is first-party pipeline output; no XSS sanitisation needed.

## Files changed

- `src/pragmata/core/annotation/collapsible_field.html`
  - `marked.min.js` `<script src=...>` in `<head>`
  - `marked.parse(text)` then `innerHTML` (with graceful fallback to plain passthrough if the CDN load fails)
  - Drops `white-space: pre-wrap` on `#content` (markdown rendering handles its own whitespace) and tightens first/last child margins so rendered HTML sits flush in the collapsed area.

## Relationship to #232

Independent of #232 (no code/test dependency). #232 configures `use_markdown` on the main `rg.TextField` chrome; this PR is about the embedded collapsible widget's own iframe. They touch disjoint code and can land in either order.

## Note on `marked.js` via CDN

Loaded from jsdelivr (`https://cdn.jsdelivr.net/npm/marked@12.0.0/marked.min.js`). Annotator browsers fetch it per record. Works through normal corporate proxies; deployments behind strict firewalls that block CDNs fall back to plain passthrough (graceful degradation: text still shows, just unrendered). If that's a concern, a follow-up could inline a smaller parser (e.g. snarkdown, ~1KB) — flagged but not done here since the demo deployment confirmed CDN works.

## Test plan

- [x] Manual: verified on a demo deployment with publikationsbot-emitted markdown-with-embedded-HTML content. Widget renders identically to the main `use_markdown=True` TextField for the same content shape.
- [ ] Reviewer: confirm `marked@12.0.0` pin is acceptable; or request inline parser instead.


saschagobel:
Closing this as part of the alpha-release scope cutoff. Rich rendering in the collapsible fields would be useful, but the current plain-text rendering remains functional and sufficient for the alpha. The external CDN dependency and unsanitized HTML rendering need proper consideration and validation rather than a quick merge. We can revisit this later if the requirement remains. Thanks for the work on this.

PR #234, opened 2026-05-27
fix(annotation): decouple partition_scope from dataset_id

henrycgbaker:
## Summary

- Introduces `partition_scope: SafePathSegment` on `AnnotationSettings`, defaulting to `dataset_id` (for backwards compatibility, but also as a sensible default)
- `resolve_import_paths()` and `load_partition_manifest()` now key off `partition_scope`; `dataset_name()` (Argilla naming) continues using `dataset_id`
- `PartitionManifest.dataset_id` renamed to `partition_scope` in Python, serialised as `"dataset_id"` via Pydantic alias — existing on-disk manifests load without migration
- `import_records()` API gains an optional `partition_scope` kwarg

## Problem fixed

Two annotation configs sharing `dataset_id: ""` would share one partition manifest, so the first config to hit `calibration_max_records` would exhaust the cap for all subsequent imports in the same run — even across unrelated domains with different workspaces.

## Backwards compat / future palns

- No changes required for existing callers. When `partition_scope` is not set, it defaults to `dataset_id`, preserving the exact previous behaviour. Existing `partition.meta.json` files on disk are unaffected (JSON key stays `"dataset_id"`).
- Plan to refactor how datasets can be named with dataset_id (currently, has to be a suffix, but this is needless constraint now we have partition_scope)

## Test plan

- [x] 807 unit tests pass (`uv run pytest tests/unit/`)
- [x] New integration regression test `TestPartitionScopeDecoupling::test_independent_scopes_each_get_calibration` — two configs with `dataset_id: ""`, distinct `partition_scope`, `calibration_max_records: 30`; asserts scope B gets > 0 calibration records after scope A hits the cap
- [x] Existing integration tests (`tests/integration/test_annotation_calibration.py`) unchanged in behaviour

### DIFF
--- a/src/pragmata/api/annotation_import.py
+++ b/src/pragmata/api/annotation_import.py
@@ -75,6 +75,7 @@ def import_records(
     api_key: str | Unset = UNSET,
     format: str = "auto",
     dataset_id: str | Unset = UNSET,
+    partition_scope: str | Unset = UNSET,
     base_dir: str | Path | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
     calibration_fraction: float | Unset = UNSET,
@@ -120,7 +121,11 @@ def import_records(
         api_key: Argilla API key.
         format: File format override — 'auto' (default), 'json', 'jsonl',
             or 'csv'. Only used for str/Path inputs.
-        dataset_id: Suffix appended to dataset names for run scoping.
+        dataset_id: Suffix appended to dataset names in Argilla (cosmetic).
+        partition_scope: Calibration-budget scope key. Defaults to
+            ``dataset_id`` when omitted. Set independently to share clean
+            Argilla dataset names across domains while keeping separate
+            calibration caps.
         base_dir: Workspace base directory. Defaults to cwd.
         config_path: Path to YAML config file for settings resolution.
         calibration_fraction: Deployment-level fraction of annotation items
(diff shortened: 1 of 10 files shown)


henrycgbaker:
Three things deliberately not fixed; capturing for future reference.

1. partition_scope: "" is indistinguishable from "not set" if not self.partition_scope treats an explicit empty string the same as an omitted field. A user who writes partition_scope: "" in YAML while having dataset_id: "DEM" will silently get partition_scope="DEM". Fixing this properly requires a None sentinel (SafePathSegment | None = None), which is a wider type change. Left as-is since an explicitly empty partition_scope with a non-empty dataset_id has no meaningful use case today.

2. Defaulting is one-way: partition_scope ← dataset_id, not vice versa:
If a caller sets partition_scope="proj" without setting dataset_id, Argilla dataset names are unscoped (retrieval_production, no suffix). Intentional — the whole point of the decoupling — but a sharp edge. API docstring on import_records warns about this.

3. export_annotations() and teardown() don't expose partition_scope:
Neither touches the partition manifest today, so no operational need. If either ever needs to correlate with manifest state, the kwarg will need to be added at that point.

saschagobel:
Closing this as part of the alpha-release scope cutoff. This addressed a real issue in the demo branch, but it targets demo-2026-05-26 and is tied to the calibration-cap work that is not part of the current alpha scope. Bringing it into the release would require rebuilding and reviewing the configuration, path, and persisted-manifest changes against current main. Thanks for the work on this.

PR #236, opened 2026-05-29
fix: set default request timeout on LLM client

henrycgbaker:
## Why
Part of the **querygen resumable-runs** feature stack (crash-resilient checkpoint/resume for the two-stage planning → dedup → realization pipeline). This PR ships first: it is the actual fix for the upstream incident (langchain → httpx with `timeout=None` hangs).

## What
- Default `timeout=600` (int) on `build_llm_runnable` — applies to every provider.
- Caller-overridable via `model_kwargs["timeout"]`.
- `int` not `float`: `ChatMistralAI.timeout: int` rejects floats; OpenAI/Azure accept both.

## Stack (7 PRs)
1. **this PR** — `01-llm-default-timeout`
2. `02-atomic-write-helper`
3. `03-paths-and-fingerprints`
4. `04-planning-batch-artifact`
5. `05-selected-blueprints-artifact`
6. `06-realization-batch-artifact`
7. `07-api-orchestration`

Final stack tip is byte-identical to feat/querygen-resumable-stack on querygen paths (modulo the deliberately-excluded annotation atomic-write adoption, deferred to a separate change).

## Test plan
- [x] `tests/unit/core/querygen/test_llm.py` — 13 tests covering Azure / Mistral / caller-override / underlying-client-timeout.
- [x] Full unit suite green.


src/pragmata/core/querygen/llm.py
saschagobel, line 1: The added default is clear and the docstring already captures the important behavior, including caller override semantics. I’d drop the explanatory inline comments around _DEFAULT_REQUEST_TIMEOUT_SECONDS and init_kwargs.update(model_kwargs) to keep this module lean.

tests/unit/core/querygen/test_llm.py
saschagobel, line 1: Non-blocking suggestion: the first two tests exercise the same default path, so I’d collapse them into one leaner test. Also, `azure_openai` is not a supported `model_provider`.

```python
def test_build_llm_runnable_applies_int_default_timeout(
    mock_llm_setup: dict[str, Any],
) -> None:
    """A default int request timeout is forwarded when the caller supplies none."""
    build_llm_runnable(
        system_text="s",
        user_text="u",
        model_provider="mistralai",
        model="magistral-medium-latest",
        api_key="k",
        output_schema=_DummyOutputSchema,
        requests_per_second=1.0,
        check_every_n_seconds=1.0,
        max_bucket_size=1,
        base_url=None,
        model_kwargs={},
    )

    init_kwargs = mock_llm_setup["init"].call_args.kwargs
    assert init_kwargs["timeout"] == 600
    assert type(init_kwargs["timeout"]) is int
```

The override test is worth keeping separately because it guards the caller contract, but I’d switch it to a supported provider, e.g., `openai`.

saschagobel (approved):
Thanks, this is a good narrow robustness fix for the upstream LLM timeout hang.

I adjusted the title/labels to treat this as a bugfix rather than resumability architecture; the two comments are non-blocking cleanup around comment verbosity and test lean-ness/provider validity.

PR #237, opened 2026-05-29
refactor(core): add shared atomic_write_text helper, adopt in querygen planning-summary export

henrycgbaker:
**Stack (6 open PRs):** #237 → #238 → #239 → #240 → #241 → #242 

**This PR (1/6):** Extracts the tmpfile + atomic-rename idiom into a shared `core/atomic_io.py` helper and adopts it for querygen's planning-summary write — the atomic-write primitive every later checkpoint artifact (#239–242) writes through.

---

## What
- `core/atomic_io.py` — `atomic_write_text` context manager: PID+uuid-uniquified tempfile, `Path.replace` for the atomic rename, no fsync (matches house style).
- Querygen `export.py` adopts it for `export_planning_summary` (indented JSON for human-scannable persisted artifacts).
- **Note:** annotation `record_builder`/`export_runner` should adopt `atomic_write_text` in a follow-up — kept out here to keep this PR's dependency surface to querygen only.

## Why
Extracts the tmpfile + atomic-rename idiom (previously duplicated inline in two annotation sites — `record_builder.write_partition_manifest` and `export_runner.write_export_csv`) into one shared core helper, then adopts it for querygen's planning-summary write. Reaches the rule-of-3 with the upcoming querygen artifact writes in #239–242.

## Test plan
- [x] `tests/unit/core/test_atomic_io.py` — round-trip + temp-cleanup-on-exception.

### DIFF
--- a/src/pragmata/core/atomic_io.py
+++ b/src/pragmata/core/atomic_io.py
@@ -10,22 +10,21 @@ from uuid import uuid4
 
 @contextmanager
 def atomic_write_text(path: Path) -> Iterator[TextIO]:
-    """Yield a text handle whose contents replace ``path`` atomically on success.
+    """Atomically replace a text file after a successful write.
 
-    Writes to a uniquified temp file in the same directory and ``Path.replace``s
-    it into place when the ``with`` block exits cleanly; on any exception the
-    temp file is removed and ``path`` is left untouched. The temp name carries
-    the PID and a random suffix so concurrent writers to the same target cannot
-    clobber each other's temp file.
+    Writes UTF-8 text to a unique temporary file in the target directory and
+    renames it into place with ``Path.replace`` when the context exits cleanly.
+    If the write fails, the temporary file is removed and the existing target is
+    left untouched. The parent directory must already exist.
 
-    No fsync (matches the repo's existing atomic-write idiom): a torn or
-    unflushed file is never the result of a successful rename, and callers that
-    validate-on-read treat a corrupt final file as drift and recompute.
+    The handle is opened with ``newline=""`` so callers writing CSV via the
+    :mod:`csv` module get correct line endings (harmless for JSON/text); the
+    temp name carries the PID and a random suffix so concurrent writers to the
+    same target cannot clobber each other.
 
-    The handle is opened in text mode with ``encoding="utf-8"`` and
-    ``newline=""`` (the latter so callers writing CSV via :mod:`csv` get correct
-    line endings; harmless for JSON/text). The parent directory must already
-    exist.
+    Atomic replacement is provided for normal readers, but the file and parent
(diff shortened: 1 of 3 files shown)


src/pragmata/core/atomic_io.py
```diff
@@ -0,0 +1,43 @@
...
+    ``newline=""`` (the latter so callers writing CSV via :mod:`csv` get correct
+    line endings; harmless for JSON/text). The parent directory must already
+    exist.
+
+    Args:
+        path: Final destination path; replaced atomically on clean exit.
+
+    Yields:
+        A writable text file handle for the temp file.
+    """
```
saschagobel, line 35: Non-blocking: the helper itself looks good, but I’d simplify this docstring. Right now it mixes the contract, implementation details, CSV-specific newline rationale, concurrency rationale, and durability caveat into one fairly dense block. For a small core utility, I think this would be easier to maintain if the docstring stated the contract briefly and moved the fsync/newline nuance into 1–2 compact notes.

For example, something along these lines would be enough:

"""Atomically replace a text file after a successful write.

Writes UTF-8 text to a unique temporary file in the target directory and
renames it into place with ``Path.replace`` when the context exits cleanly.
If the write fails, the temporary file is removed and the existing target is
left untouched.

This provides atomic replacement for normal readers, but intentionally does
not fsync the file or parent directory. The parent directory must already
exist.

Args:
        path: Final destination path; replaced atomically on clean exit.

Yields:
        A writable text file handle for the temp file.
"""

src/pragmata/core/querygen/export.py
saschagobel, line 1: I’d avoid adding a private `_atomic_write_json()` helper in `querygen/export.py`. As currently used, it adds an extra layer for one call site; if this stays local, I think the write is clearer inline in `export_planning_summary()`.

More importantly, this creates an inconsistency in the same module: `export_planning_summary()` uses atomic JSON writing, while `export_queries()` still writes `meta_path` directly with `Path.write_text()`. Since eval and later querygen artifacts will likely need the same behavior, I’d prefer to either:

1. inline the `atomic_write_text()` block here and keep the PR narrowly scoped, or
2. promote this to a shared `atomic_write_json()` helper in `core/atomic_io.py` and use it for both `export_planning_summary()` and the query metadata JSON.

My preference is option 2, with a small helper that wraps `json.dump(..., indent=2)` plus a trailing newline. That avoids a local one-off abstraction and gives the upcoming eval/querygen artifact writes one canonical JSON path.

saschagobel (approved):
Thanks, this is a helpful refactor and tests cover the core behavior well. I left two non-blocking comments: one on simplifying the dense atomic_write_text docstring, and one on avoiding a one-off local JSON helper / applying atomic JSON writes consistently.

saschagobel (approved):

PR #238, opened 2026-05-29
feat(querygen): add checkpoint run-dir paths and per-stage fingerprints

henrycgbaker:

**This PR (2/6):** Foundation. Adds the checkpoint path bundle, the per-stage fingerprint helpers used to detect config drift, and the embedding-model constant — all dormant at merge time (consumed by the later artifacts and orchestration).

---

## What
- `querygen_paths.py`: nests resumable state under `<run_dir>/checkpoints/` (`checkpoints_dir`, `planning_batches_dir`, `selected_blueprints_json`, `realization_batches_dir`). Separates the resumable working state from the deliverable CSV.
- `planning_summary.py`: `_sha256_json` shared helper; `fingerprint_planning_llm_settings` + `fingerprint_realization_llm_settings` (per-stage). A realization-model change re-runs **only Stage 2**; the frozen Stage 1 result and planning checkpoints stay valid.
- `deduplication.py`: `EMBEDDING_MODEL_CHECKPOINT` constant exposed for the dedup artifact's provenance field.

## Why
Foundation for the checkpoint/resume work that lands in #239–242. The path bundle, fingerprint helpers, and embedding-model constant land first so the later artifact and orchestration PRs are pure consumers — nothing here is wired into a live code path at merge.

## Test plan
- [x] `tests/unit/core/paths/test_querygen_paths.py` — checkpoints/ nesting expectations.
- [x] Existing planning-summary + deduplication tests still pass.

### DIFF
--- a/src/pragmata/core/atomic_io.py
+++ b/src/pragmata/core/atomic_io.py
@@ -1,0 +1,43 @@
+"""Atomic file-write helper shared across pragmata subsystems."""
+
+import os
+from collections.abc import Iterator
+from contextlib import contextmanager
+from pathlib import Path
+from typing import TextIO
+from uuid import uuid4
+
+
+@contextmanager
+def atomic_write_text(path: Path) -> Iterator[TextIO]:
+    """Yield a text handle whose contents replace ``path`` atomically on success.
+
+    Writes to a uniquified temp file in the same directory and ``Path.replace``s
+    it into place when the ``with`` block exits cleanly; on any exception the
+    temp file is removed and ``path`` is left untouched. The temp name carries
+    the PID and a random suffix so concurrent writers to the same target cannot
+    clobber each other's temp file.
+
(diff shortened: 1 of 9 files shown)


saschagobel (commented):
I think the checkpoint path additions here are directionally fine and seem like the right filesystem shape for crash recovery.

My concern is with the per-stage fingerprint helpers, especially `fingerprint_planning_llm_settings()` and `fingerprint_realization_llm_settings()`. Those push the design toward partial reuse under changed settings, e.g. reusing Stage 1 when only the realization model changes. For the first version, I think we should scope this to **resumability** only: same `run_id`, same effective settings, recover after crash/abort. If settings that affect the run differ, I’d prefer one strict compatibility check that fails clearly and asks the user to start a new run or delete/recreate the existing one.

So I’d suggest removing the per-stage fingerprints in this stack and replacing them with a single same-run compatibility checkpoint identity.

saschagobel:

PR #239, opened 2026-05-29
feat(querygen): add planning-batch checkpoint artifact and read engine

henrycgbaker:

**This PR (3/6):** First of three vertical artifact slices. Adds per-batch checkpointing for Stage 1 planning, plus the **generic read engine** all three artifact readers (this PR's, #240's, #241's) delegate to.

---

## What
- `PlanningBatchArtifact` schema (`querygen_output.py`) with `extra="forbid"` + `1:1 candidate_ids/blueprints` model_validator.
- `assemble_planning_batch_artifact` stamps `pragmata_version` + `created_at` internally.
- `export_planning_batch_artifact` — atomic JSON write via the helper from #237.
- `read_checkpoint_artifact` (engine): returns `(artifact, list[DriftedField])` — `(artifact, [])` reusable, `(None, [])` absent, `(None, drifted)` config-drifted; raises on damaged file (caller self-heals).
- `DriftedField` — `@dataclass(frozen=True, slots=True)` with `field`/`stored`/`expected`, in line with core's frozen-dataclass container convention.
- `read_planning_batch_artifact` — thin typed wrapper over the engine (matches the existing `read_planning_summary_artifact` per-artifact typed-reader house pattern).

## Why
The read engine is the reusable core of the whole stack: every checkpoint read (#240's frozen result, #241's realization batches, #242's orchestration) goes through `read_checkpoint_artifact`, so the drift/absent/damaged trichotomy is defined once here and the later PRs only add typed wrappers.

## Test plan
- [x] `tests/unit/core/querygen/test_planning_batches.py` — schema, round-trip, header-mismatch (drift) per field, version-mismatch, malformed/extra-field, length-validator.

### DIFF
--- a/src/pragmata/core/querygen/assembly.py
+++ b/src/pragmata/core/querygen/assembly.py
@@ -1,10 +1,16 @@
 """Assembly of synthetic query generation output."""
 
+import importlib.metadata
 from datetime import UTC, datetime
 
 from pragmata.core.querygen.planning_summary import fingerprint_querygen_spec
 from pragmata.core.schemas.querygen_input import QueryGenSpec
-from pragmata.core.schemas.querygen_output import PlanningSummaryArtifact, SyntheticQueriesMeta, SyntheticQueryRow
+from pragmata.core.schemas.querygen_output import (
+    PlanningBatchArtifact,
+    PlanningSummaryArtifact,
+    SyntheticQueriesMeta,
+    SyntheticQueryRow,
+)
 from pragmata.core.schemas.querygen_plan import QueryBlueprint
 from pragmata.core.schemas.querygen_realize import RealizedQuery
 from pragmata.core.schemas.querygen_summary import PlanningSummaryState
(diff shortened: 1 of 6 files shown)


saschagobel (commented):
The `PlanningBatchArtifact` boundary itself makes sense to me for Stage 1 crash recovery, and the assemble/export/read split is broadly in line with the existing artifact pattern.

My concern is with the generic read/drift engine introduced here. `DriftedField` and `read_checkpoint_artifact()` feel heavier than needed for a first crash-resumability pass, especially because the broader stack currently moves toward config-drift handling and partial reuse under changed settings. For v1, I think we can keep this much simpler: checkpoint absent → run; checkpoint present and compatible with the same `run_id` / same effective settings → reuse; checkpoint incompatible → fail clearly. That likely only needs a small internal checkpoint read helper, not a reusable drift framework returning `(artifact, list[DriftedField])`.

So I’d suggest simplifying here and keeping the planning-batch reader focused on strict same-run compatibility.

saschagobel:

PR #246, opened 2026-06-01
feat(annotation): persist n_retrieved_chunks on retrieval records

henrycgbaker:

**Chain goal:** Persist K (number of retrieved chunks per query) on retrieval records, surface it as panel-completeness columns + sidecar at export time, ship a live `annotation status` CLI on top, and add an opt-in `--tag-partial-panels` advisory write so annotators can filter partially-done panels in the Argilla UI.

**This PR (1/6):** Foundation. Stamps `n_retrieved_chunks` on retrieval record metadata at import and surfaces it on `RetrievalAnnotation`. Ships nothing user-visible on its own — the export columns (#247/#268) and live status CLI (#248/#269) consume it.

---

### Record building
- `core/annotation/record_builder.py`: add `"n_retrieved_chunks": len(pair.chunks)` to every chunk-record's metadata dict.

### Task definition
- `core/annotation/argilla_task_definitions.py`: declare `IntegerMetadataProperty("n_retrieved_chunks", min=1, visible_for_annotators=False)` on retrieval task settings.

### Schema
- `core/schemas/annotation_export.py`: `RetrievalAnnotation.n_retrieved_chunks: int` field (mirrors the existing `chunk_rank: int`).

### Export readback
- `core/annotation/export_fetcher.py`: `_build_row` retrieval branch reads `metadata.get("n_retrieved_chunks", 0)` (defensive default for records imported before this PR).

### Tests
- New `test_n_retrieved_chunks_stamped_on_every_chunk_record` in `test_import.py` locks the persist contract.
- Metadata-name list in `test_argilla_task_definitions.py` updated.
- Fixture refresh across `test_annotation_export.py`, `test_export_runner.py`, `test_export_constraint_checks.py`, `test_iaa_runner.py`, `test_export_api.py`.

## Why

K varies per query (~15% K<5, 61% K=5, 24% K>5; the retriever is threshold-based, not top-K), so the export side can't infer K from `chunk_rank` or from record-counting alone — it needs a persisted SSOT to drive true top-K metrics in the consumer pipeline (precision@K, recall@K, NDCG@K, MRR@K).

End-to-end mirror of the existing `chunk_rank` field: stamped at import, declared on the task, modelled on `RetrievalAnnotation`, read back by the fetcher.

A one-off backfill script for records imported before this PR (in-flight batches) lives outside the supported CLI/API surface — retrospective ops migration, not packaged functionality.

## Base

`feat/querygen-resumable-stack` — kitchen-sink integration branch carrying the full dep surface this stack assumes (`Locale`, `atomic_write_text`, `LogicalConstraint`, `WIDGET_FIELD_PLACEHOLDERS`, the `export_constraint_checks` rename). Rebases cleanly to `main` once those upstream stacks land.

## Test plan
- [x] `uv run python -m pytest tests/unit` (863 passing)

---

### DIFF
--- a/src/pragmata/core/annotation/export_fetcher.py
+++ b/src/pragmata/core/annotation/export_fetcher.py
@@ -65,7 +65,11 @@ def _build_row(
             chunk_id=metadata.get("chunk_id", ""),
             doc_id=metadata.get("doc_id", ""),
             chunk_rank=metadata.get("chunk_rank", 0),
-            n_retrieved_chunks=metadata.get("n_retrieved_chunks", 0),
+            # -1 is a sentinel for "metadata absent" (pre-backfill/pre-this-PR records) —
+            # real K is always >=1 (enforced by the min=1 metadata constraint), and -1
+            # can't be confused with a real count. Downstream consumers must guard with
+            # `k > 0` before using it as a divisor.
+            n_retrieved_chunks=metadata.get("n_retrieved_chunks", -1),
             topically_relevant=_to_bool(answers.get("topically_relevant")),
             evidence_sufficient=_to_bool(answers.get("evidence_sufficient")),
             misleading=_to_bool(answers.get("misleading")),
--- a/src/pragmata/core/annotation/export_fetcher.py
+++ b/src/pragmata/core/annotation/export_fetcher.py
@@ -64,6 +64,7 @@ def _build_row(
             chunk_id=metadata.get("chunk_id", ""),
             doc_id=metadata.get("doc_id", ""),
             chunk_rank=metadata.get("chunk_rank", 0),
+            n_retrieved_chunks=metadata.get("n_retrieved_chunks", 0),
             topically_relevant=_to_bool(answers.get("topically_relevant")),
             evidence_sufficient=_to_bool(answers.get("evidence_sufficient")),
             misleading=_to_bool(answers.get("misleading")),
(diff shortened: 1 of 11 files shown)


src/pragmata/core/annotation/export_fetcher.py
```diff
@@ -64,6 +64,7 @@ def _build_row(
             chunk_id=metadata.get("chunk_id", ""),
             doc_id=metadata.get("doc_id", ""),
             chunk_rank=metadata.get("chunk_rank", 0),
+            n_retrieved_chunks=metadata.get("n_retrieved_chunks", 0),
```
ddimmery: This zero is a sentinel, right? I ask because I assume downstream we're assuming `k > 0` (e.g. in the task definition) and then use it as a denominator in metrics, so we'll need to watch it there.

ddimmery (approved):
This is already live, right? Otherwise there would need to be some kind of migration to deal with adding columns mid-run.

henrycgbaker:
> This is already live, right? Otherwise there would need to be some kind of migration to deal with adding columns mid-run.

yes this is already live (I had a non-git tracked backfill script)

src/pragmata/core/annotation/export_fetcher.py
henrycgbaker: yes, 0 is sentinel - real values guaranteed ≥1 by Argilla schema (IntegerMetadataProperty(..., min=1)). I could change to `None`, only (non-major) issue is that this is close in the pipeline to csv/df export - so this would be adding non-int datatype which i think pandas coerces into `NaN`. In reality this is a non-issue as it's not reachable/edge case due to IntegerMetadataProperty

ddimmery: maybe `-1` to keep it an integer but make it very clear it's different?

PR #247, opened 2026-06-01
feat(annotation): panel completeness calculator

henrycgbaker:
**Stack (6 PRs):** #246 (merged) → #247 → #268 → #248 → #269 → #249

**Chain goal:** Persist K (number of retrieved chunks per query) on retrieval records, surface it as panel-completeness columns + sidecar at export time, ship a live `annotation status` CLI on top, and add an opt-in `--tag-partial-panels` advisory write so annotators can filter partially-done panels in the Argilla UI.

**This PR (2/6):** The standalone, export-agnostic panel-completeness calculator, plus the shared retrieval-walk primitive it and #268 both consume. Split out of what was originally a single ~1000-line PR; wiring completeness into the retrieval export path ships separately in #268.

---

### Completeness aggregator
- New `core/annotation/completeness.py`: groups retrieval snapshots by `record_uuid`, distinct-by-`chunk_id`, derives per-panel facts, emits `CompletenessReport` (`by_uuid` + `summary`). Aggregation is split into a typed `_PanelAccumulator` plus `_aggregate_snapshots` (grouping) / `_summarize_groups` (transform) passes.

### Shared retrieval walk
- `core/annotation/export_fetcher.py`: `RetrievalRecordSnapshot` + `walk_retrieval_records` — one Argilla scroll, no status filter (so records lacking terminal responses are still seen for the integrity check), feeding both this module's completeness pass and #268's row emission. Plus `resolve_task_purposes`, a topology lookup extracted from `fetch_task` (no behaviour change).

### Schema
- `core/schemas/annotation_export.py`: `CompletenessSummary` + `KBucketStat` models for the export sidecar (populated by #268).

### Tests
- New `test_completeness.py`: happy path, distinct-by-chunk_id, orphan exclusion, integrity warnings (records-vs-K + mixed-backfill within panel), all-unknown-K warning, K-bucket cross-tab + per-K histogram, STRICT `panel_complete` (discards do NOT count), `n_discarded_chunks` subset of `n_annotated_chunks`, calibration walked when topology declares it.

## Why STRICT `panel_complete`

The computed `panel_complete: bool` is STRICT: True iff every K chunk in the panel has at least one **submitted** response. Discarded responses are abstentions, not judgements (pragmata's `DiscardReason` enum is refusal-only: `INVALID_OR_UNREALISTIC` / `UNCLEAR` / `OUTSIDE_REVIEWER_EXPERTISE`), so treating them as covered would feed unjudged chunks into NDCG@K / precision@K denominators as 0-relevance labels. The bool is derivable from the count columns, so a consumer wanting the permissive policy computes `n_annotated_chunks == n_retrieved_chunks` in one line.

The retriever is threshold-based, so dropping partial panels biases metrics toward small K (MNAR); this module is the shared calculation both the export path (#268) and the live status path (#248) build on — one aggregator, two consumers.

## Test plan
- [x] `uv run python -m pytest tests/unit` (1040 passing)

### DIFF
--- a/src/pragmata/core/annotation/completeness.py
+++ b/src/pragmata/core/annotation/completeness.py
@@ -1,0 +1,219 @@
+"""Per-panel retrieval completeness: how many chunks per query have a submitted response.
+
+A "panel" is the set of retrieval chunk-records sharing one ``record_uuid``
+(one query, K chunks).
+
+The exported ``panel_complete`` flag uses the STRICT (submitted-only)
+definition: a panel is complete iff every one of its K chunks has at least
+one SUBMITTED response. This matches what downstream eval scorers actually
+need: pragmata's ``DiscardReason`` enum is refusal-only
+(``INVALID_OR_UNREALISTIC`` / ``UNCLEAR`` / ``OUTSIDE_REVIEWER_EXPERTISE``),
+so a discarded chunk is an abstention — not a judgement — and treating it
+as covered would feed unjudged chunks into NDCG@K / precision@K denominators
+as if they carried a 0-relevance label.
+
+Three count columns let consumers derive any policy they need:
+
+- ``n_annotated_chunks`` = distinct chunks with >=1 terminal response (submitted OR discarded)
+- ``n_submitted_chunks`` = distinct chunks with >=1 submitted response
+- ``n_discarded_chunks`` = distinct chunks with >=1 discarded response
+
+Sets aren't disjoint — a chunk where annotator A submitted and annotator B
+discarded is in all three. So ``n_submitted + n_discarded >= n_annotated``
+(equality when no mixed chunks). A consumer wanting the PERMISSIVE
+("terminal counts as covered") policy derives it in one line::
+
+    panel_complete_permissive = (n_annotated_chunks == n_retrieved_chunks)
+
+Computation is INDEPENDENT of the export's ``include_discarded`` flag (which
+gates only row emission to the CSV). When invoked from ``run_export`` the
+underlying walk is shared with the export's row-emission fetch so we don't
+scroll the retrieval datasets twice; standalone callers use the
+``compute_completeness(client, settings)`` convenience that does its own walk.
+"""
+
+import logging
+from dataclasses import dataclass
+
+import argilla as rg
+
+from pragmata.core.annotation.export_fetcher import RetrievalRecordSnapshot, walk_retrieval_records
+from pragmata.core.schemas.annotation_export import CompletenessSummary, KBucketStat
+from pragmata.core.settings.annotation_settings import AnnotationSettings
+
+logger = logging.getLogger(__name__)
+
+
+@dataclass(frozen=True)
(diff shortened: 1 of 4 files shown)


henrycgbaker:
open question: the export ships `panel_complete: bool` per retrieval row, computed as (`n_submitted_chunks == n_retrieved_chunks) AND (n_retrieved_chunks > 0`) -> = STRICT, since pragmata's `DiscardReason` enum is refusal-only (`INVALID_OR_UNREALISTIC / UNCLEAR / OUTSIDE_REVIEWER_EXPERTISE`) so a discarded chunk is an absence, not judgement. The bool is derivable from columns already on the row (`n_submitted_chunks, n_retrieved_chunks`) -> technically redundant column (pure derivation).

I picked (i)

- (i) (current) ship panel_complete per row + sidecar n_complete aggregate. 1-column filter for df consumers; sidecar aggregate matches the row predicate 
-  (ii) Drop row bool, keep sidecar n_complete. Saves a column; sidecar still expresses pragmata's STRICT policy in aggregate form; consumer derives the row predicate if they want to verify.
- (iii) Drop both -> policy-neutral data contract: ship only the counts (n_annotated_chunks, n_submitted_chunks, n_discarded_chunks, n_records_seen) + bucketed n_panels. Consumer applies their own predicate AND their own aggregation.

NB:  annotator-facing needs_completion tag (live status path, PR #249) is unaffected by this — it's computed in-memory during the status walk, not from the CSV.

src/pragmata/core/annotation/completeness.py
```diff
@@ -0,0 +1,219 @@
...
+_BUCKET_KEYS = ("k_lt_5", "k_eq_5", "k_gt_5")
+
+
+def compute_completeness_from_records(snapshots: list[RetrievalRecordSnapshot]) -> CompletenessReport:
+    """Pure aggregator over pre-walked retrieval snapshots.
+
+    Shared by ``compute_completeness`` (standalone) and ``run_export`` (which
+    walks once and feeds both the row-emission and the completeness passes).
+    """
+    groups: dict[str, dict] = {}
```
ddimmery: Could use a `dataclass` or `TypedDict` for clearer type safety.

src/pragmata/core/annotation/completeness.py
```diff
@@ -0,0 +1,219 @@
...
+        return "k_lt_5"
+    if k == 5:
+        return "k_eq_5"
+    return "k_gt_5"
+
+
+_BUCKET_KEYS = ("k_lt_5", "k_eq_5", "k_gt_5")
+
+
+def compute_completeness_from_records(snapshots: list[RetrievalRecordSnapshot]) -> CompletenessReport:
```
ddimmery: This function is pretty complicated. To simplify you could extract the aggregation part (iterating through snapshots to populate the dict) and the transformation part (iterating through groups to construct the summary). Not critical, but could be useful (could also allow testing individual components more cleanly.

ddimmery (approved):

PR #248, opened 2026-06-01
feat(annotation): panel-status core engine

henrycgbaker:
**Stack (6 PRs):** #246 → #247 → #268 → #248 → #269 → #249

**This PR (4/6):** Core engine for live per-panel completeness + all-task progress, computed directly against Argilla (no export round-trip). Exposed via CLI/API in #269. Stacked on #268.

- **All tasks**: walks every dataset and computes record progress (total / completed) across `retrieval` / `grounding` / `generation`.
- **Retrieval panels**: the retrieval row also carries panel-completeness (`panel_complete` = all K chunks have ≥1 submitted, `overlap_satisfied` = each chunk meets its `min_submitted`, integrity vs `n_retrieved_chunks`).
- **Config-free + multi-domain**: selects datasets by name-prefix (no `AnnotationSettings`), covering every workspace in one pass. Panels keyed by `(workspace, record_uuid)`. `min_submitted` read live from each dataset's Argilla settings.

The CLI/API surface (`pragmata annotation status`) ships in #269. The `--tag-partial-panels` write ships in #249.

PR #249, opened 2026-06-01
feat(annotation): tag partial panels via status --tag-partial-panels

henrycgbaker:

**This PR (6/6):** Adds the advisory `needs_completion` tag write onto `annotation status`.

- `status --tag-partial-panels`: stamps `needs_completion` on the **unresolved chunks of PARTIAL panels** (`0 < submitted < K`) and clears stale tags, sharing the single status walk. Annotators filter `needs_completion=true` in the Argilla UI to focus on the records that will complete a partially-done panel.
- **Partial-only gate** (not "any incomplete panel"): a fully-unstarted panel is never tagged, so the tag stays a selective mop-up signal instead of ≈ everything pending.
- **Multi-domain + prod/cal-split aware**: a panel whose chunks are split across a workspace's production + calibration datasets (per-item calibration) is evaluated as one unit; writes route to each chunk's owning dataset.

Reuses `metadata_ops` for the replace-safe upsert. Stacked on #269.

Also re-exports `TagResult` from the `pragmata.annotation` facade: `StatusReport.tag_result` is a public field of this type, but the facade export was pre-declared two PRs ago (before `panel_status.py` existed), dropped once that PR's version of the module still didn't define the class, and never re-added once this PR added it for real.

### DIFF
--- a/src/pragmata/core/annotation/metadata_ops.py
+++ b/src/pragmata/core/annotation/metadata_ops.py
@@ -1,0 +1,76 @@
+"""Shared safe metadata operations for live Argilla mutations.
+
+Used by both the ``--tag-partial-panels`` write path in ``panel_status`` and the
+one-off backfill script under ``scripts/``. Centralises the two safety
+invariants that every metadata write must respect on Argilla v2.8.0:
+
+1. **Argilla metadata is REPLACE, not merge.** Every ``dataset.records.log``
+   call replaces the record's metadata wholesale. To avoid clobbering
+   existing keys, always fetch the current dict, merge in the update, and
+   send the FULL resulting dict.
+2. **Property declaration is additive and idempotent.** Adding a metadata
+   property to an existing dataset is non-destructive, but the SDK raises
+   if the property already exists (with override warning). Skip the add
+   when the property is already present.
+
+Writes go via ``rg.Record(id=..., metadata={...})`` rather than a raw dict
+payload: the SDK's ``IngestedRecordMapper`` flattens dict keys against the
+dataset schema, so a ``{"id": ..., "metadata": {...}}`` shape would treat
+"metadata" as an unknown top-level attribute and silently send an empty
+metadata dict (wiping the record). Passing an ``rg.Record`` bypasses the
+mapper.
+"""
+
+import logging
+from collections.abc import Iterable, Mapping
+
+import argilla as rg
+
+logger = logging.getLogger(__name__)
+
+
+def ensure_metadata_property(dataset: rg.Dataset, prop: rg.MetadataType) -> bool:
+    """Idempotently declare ``prop`` on ``dataset``.
+
+    Returns True if the property was newly added (and the dataset settings
+    pushed to the server), False if it was already present.
+    """
+    existing = dataset.settings.metadata[prop.name]
+    if existing is not None:
+        return False
+    dataset.settings.add(prop)
+    dataset.settings.update()
+    logger.info("Declared metadata property %r on dataset %s", prop.name, dataset.name)
+    return True
+
+
+def build_metadata_upsert(
+    record: rg.Record,
+    updates: Mapping[str, object],
+    *,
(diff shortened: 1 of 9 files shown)


src/pragmata/core/annotation/metadata_ops.py
```diff
@@ -0,0 +1,76 @@
...
+    logger.info("Declared metadata property %r on dataset %s", prop.name, dataset.name)
+    return True
+
+
+def build_metadata_upsert(
+    record: rg.Record,
+    updates: Mapping[str, object],
+    *,
+    remove_keys: Iterable[str] = (),
+) -> rg.Record | None:
```
ddimmery, line 52: Hm. This is all non-obvious behavior, but as best I can tell without testing, this is also recommended for v2.8.0. We're still just saying `"argilla>=2.0,<3.0"` -- maybe it would be better to pin to 2.8 and force intentional version changes to reduce the risk of something changing about upsert behavior and clobbering information we want to retain?

src/pragmata/core/annotation/panel_status.py
```diff
@@ -415,3 +451,81 @@ def _row(label: str, task: str, b: dict[str, int]) -> ProgressRow:
...
+    """
+    datasets_by_key: dict[tuple[str, str], rg.Dataset] = {}
+    for rec in collected.records:
+        datasets_by_key.setdefault((rec.workspace, rec.dataset.name), rec.dataset)
+    for dataset in datasets_by_key.values():
+        ensure_metadata_property(dataset, rg.TermsMetadataProperty(NEEDS_COMPLETION_KEY, visible_for_annotators=True))
+
+    batched: dict[tuple[str, str], list[rg.Record]] = {}
+    n_tagged = n_cleared = n_already = 0
+    for (_ws, uuid), group in _group_by_panel(collected.records).items():
```
ddimmery, line 488: So we're completely separating calibration from production here, and "partial" means "within the context of this particular dataset"? This is ok, but it should be documented.

src/pragmata/core/annotation/metadata_ops.py
```diff
@@ -0,0 +1,76 @@
+"""Shared safe metadata operations for live Argilla mutations.
+
+Used by both the ``--tag-partial-panels`` write path in ``panel_status`` and the
+one-off backfill script under ``scripts/``. Centralises the two safety
```
ddimmery: I don't see this script, is it meant to be here?

src/pragmata/core/annotation/panel_status.py
```diff
@@ -415,3 +451,81 @@ def _row(label: str, task: str, b: dict[str, int]) -> ProgressRow:
...
+            if should_have_tag and already:
+                n_already += 1
+                continue
+            if should_have_tag:
+                upsert = build_metadata_upsert(rec.record, {NEEDS_COMPLETION_KEY: NEEDS_COMPLETION_VALUE})
+            elif already:
+                upsert = build_metadata_upsert(rec.record, {}, remove_keys=[NEEDS_COMPLETION_KEY])
+            else:
+                continue
+            if upsert is None:
```
ddimmery, line 505: This should never be the case, should it?

src/pragmata/core/annotation/panel_status.py
```diff
@@ -55,13 +58,31 @@
 
 RETRIEVAL_TASK = "retrieval"
 _TASK_ORDER = {"retrieval": 0, "grounding": 1, "generation": 2}
+NEEDS_COMPLETION_KEY = "needs_completion"
+NEEDS_COMPLETION_VALUE = "true"
```
ddimmery, line 65: I think there should be a comment indicating that if these are changed, there may need to be manual modifications to the Argilla data to remove old, stale metadata (the tools here will not get rid of it afaict). I can't imagine there will ever be a reason to change this mid-run, though.

ddimmery (approved):

PR #253, opened 2026-06-15
feat: resolve train input paths from annotation exports

saschagobel:
### Summary

Adds eval train input path resolution to the eval path layer.

The new resolver determines the concrete labeled CSV consumed by `eval train` from one of three sources:

- a direct user-provided labeled CSV path
- an explicit annotation `export_id`
- the latest valid annotation export for the requested task

It also preserves annotation export provenance by returning the resolved `annotation_export_id` when the training input comes from the annotation tool.

### Key changes

- Rename the existing score path bundle from `EvalPaths` to `EvalScorePaths`.
- Add `EvalTrainPaths` for resolved eval train input paths.
- Add `resolve_eval_train_paths(...)` with selector precedence:
  - direct `labeled_data_path`
  - explicit `export_id`
  - latest valid annotation export for the task
- Add `find_latest_annotation_export_id(...)` to select the newest annotation export with:
  - valid `annotation_export.meta.json`
  - task membership in export metadata
  - matching task CSV presence
- Reuse annotation export path resolution via `resolve_export_paths(...)` instead of duplicating annotation export layout logic.

### Tests

- Update score path tests for the `EvalScorePaths` rename.
- Add unit tests covering:
  - direct labeled CSV path resolution
  - missing direct CSV rejection
  - explicit annotation export ID resolution
  - missing annotation export metadata rejection
  - missing requested task CSV rejection
  - latest-export fallback when no selector is provided
  - latest-export filtering by task
  - latest-export rejection when no exports or no matching exports exist
  - deterministic tie-breaking for equal export timestamps, if covered

### DIFF
--- a/src/pragmata/core/paths/eval_paths.py
+++ b/src/pragmata/core/paths/eval_paths.py
@@ -1,14 +1,155 @@
-"""Path bundles for eval score runs."""
+"""Path bundles for eval workflows."""
 
 from dataclasses import dataclass
+from datetime import datetime
 from pathlib import Path
 from typing import Self
 
+from pragmata.core.paths.annotation_paths import resolve_export_paths
 from pragmata.core.paths.paths import WorkspacePaths
+from pragmata.core.schemas.annotation_export import AnnotationExportMeta
+from pragmata.core.schemas.annotation_task import Task
 
 
 @dataclass(frozen=True, slots=True)
-class EvalPaths:
+class EvalTrainPaths:
+    """Resolved labeled input path and source provenance for eval train.
+
+    Attributes:
+        training_input_csv: Concrete labeled CSV path consumed by eval train.
+        annotation_export_id: Resolved annotation export ID when the input comes
+            from the annotation tool. ``None`` for direct user-provided CSV input.
+    """
+
+    training_input_csv: Path
+    annotation_export_id: str | None = None
+
+
+def find_latest_annotation_export_id(
+    *,
+    workspace: WorkspacePaths,
+    task: Task,
+) -> str:
+    """Find the most recent annotation export containing the requested task CSV."""
+    exports_dir = workspace.tool_root("annotation") / "exports"
+    if not exports_dir.is_dir():
+        raise FileNotFoundError(
+            f"No annotation exports found for eval training. Expected annotation exports directory at {exports_dir}."
+        )
+    completed_exports: list[tuple[datetime, str]] = []
+
+    for meta_path in exports_dir.glob("*/annotation_export.meta.json"):
+        meta = AnnotationExportMeta.model_validate_json(meta_path.read_text(encoding="utf-8"))
+
+        if task not in meta.tasks:
+            continue
+
+        export_paths = resolve_export_paths(
+            workspace=workspace,
+            export_id=meta.export_id,
+        )
+
+        match task:
+            case Task.RETRIEVAL:
+                task_csv = export_paths.retrieval_annotation_csv
+            case Task.GROUNDING:
+                task_csv = export_paths.grounding_annotation_csv
+            case Task.GENERATION:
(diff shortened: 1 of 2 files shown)


src/pragmata/core/paths/eval_paths.py
```diff
@@ -1,14 +1,155 @@
-"""Path bundles for eval score runs."""
+"""Path bundles for eval workflows."""
 
 from dataclasses import dataclass
+from datetime import datetime
 from pathlib import Path
 from typing import Self
```
lramir14, line 6: I think this import is not used.

src/pragmata/core/paths/eval_paths.py
```diff
@@ -1,14 +1,155 @@
...
+        )
+
+    resolved_export_id = export_id or find_latest_annotation_export_id(
+        workspace=workspace,
+        task=task,
+    )
+    export_paths = resolve_export_paths(
+        workspace=workspace,
+        export_id=resolved_export_id,
+    )
```
lramir14, line 123: Isn't this function (resolve_eval_train_paths) revalidating the validation done already by find_latest_annotation_export_id? the function already guarantees metadata, and csvs.

lramir14 (approved):
thank you for your nice work. I approve but have two non-blocking comments. The first one is that the Self import seems unused, so can be removed.  The second  is about a possible redundant validation in resolve_eval_train_paths for the auto-select branch, when export_id is None and find_latest_annotation_export_id is called, it already guarantees the metadata and CSV exist, so the checks below are unreachable in that branch. Everything else looks good.

src/pragmata/core/paths/eval_paths.py
saschagobel: `datetime` is used in the local type annotation for completed_exports:

`completed_exports: list[tuple[datetime, str]] = []`

That said, it is only used for that type annotation, so I can remove the explicit annotation and let the type be inferred if you prefer avoiding the import.

src/pragmata/core/paths/eval_paths.py
saschagobel: I think this validation is still needed. `find_latest_annotation_export_id()` only runs when no explicit `export_id` is provided. If the caller passes `export_id`, that lookup and its validation are skipped entirely.

`resolve_export_paths()` also only constructs the path bundle; it does not validate that metadata or CSV files exist. So `resolve_eval_train_paths()` still needs to validate the explicitly selected export before returning the train input path.

saschagobel:
> thank you for your nice work. I approve but have two non-blocking comments. The first one is that the Self import seems unused, so can be removed.

`Self` is used as the return type of `EvalScorePaths.ensure_dirs()`:

`def ensure_dirs(self) -> Self:`

So I think the import is still needed unless we change that annotation.

> The second is about a possible redundant validation in resolve_eval_train_paths for the auto-select branch, when export_id is None and find_latest_annotation_export_id is called, it already guarantees the metadata and CSV exist, so the checks below are unreachable in that branch. Everything else looks good.

Yes, for the auto-select branch specifically, `find_latest_annotation_export_id()` has already filtered to an export with metadata and the requested task CSV. The checks below are still needed for the explicit `export_id` path, because passing `export_id` skips `find_latest_annotation_export_id()` entirely.
I kept the validation in `resolve_eval_train_paths()` as a shared post-selection guard for both paths, but I can add a short comment or split the explicit/auto-select branches if you think that would make the intent clearer.

src/pragmata/core/paths/eval_paths.py
lramir14: All good, I agree is better to have them explicit.

src/pragmata/core/paths/eval_paths.py
lramir14: Ok, it's clear now. All good.

PR #255, opened 2026-06-16
feat: add predict dataframe import

saschagobel:
### Summary

Adds eval prediction dataframe import support.

`import_eval_predict_frame(...)` reads an explicit unlabeled prediction CSV and validates it against the task-specific eval prediction contract. This mirrors the existing train import while keeping prediction input explicit and free of annotation export lookup logic.

### Key changes

- Add `import_eval_predict_frame(...)` to `core/eval/imports.py`.
- Read prediction CSVs via pandas using UTF-8 encoding.
- Validate imported frames with the existing task-specific eval prediction schema.

### Tests

- Add unit tests covering:
  - successful unlabeled prediction CSV import and validation
  - preservation of additional provenance columns
  - rejection of labeled prediction inputs
  - rejection of prediction CSVs that violate the eval prediction data contract
- Group train and predict import tests by workflow-specific test classes.

### DIFF
--- a/src/pragmata/core/eval/imports.py
+++ b/src/pragmata/core/eval/imports.py
@@ -5,7 +5,7 @@ from pathlib import Path
 import pandas as pd
 
 from pragmata.core.schemas.annotation_task import Task
-from pragmata.core.schemas.eval_input import validate_eval_train_frame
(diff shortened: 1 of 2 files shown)


lramir14 (approved):
Thanks for the work. I think it is a nice implementation, and satisfies the PR specification. Approved without comments.

PR #257, opened 2026-06-22
feat: add eval training split groups and row consolidation

saschagobel:
### Summary

Extends the eval transform layer with training-specific preparation before handing data to `tlmtc`.

Retrieval training rows now get a `split_group` column derived from `record_uuid`, so rows from the same retrieval example stay together across train/validation/test splits. Training data also consolidates duplicate annotation rows deterministically before column renaming.

### Key changes

- Map retrieval train `record_uuid` to `tlmtc`'s `split_group` column.
- Leave grounding and generation ungrouped by default.
- Add duplicate training-row consolidation before `tlmtc` column mapping.
- Group duplicate annotation units by task:
  - retrieval: `record_uuid`, `chunk_id`
  - grounding/generation: `record_uuid`
- For duplicate groups:
  - keep single-row groups unchanged
  - select the earliest row matching the strict per-label majority vector
  - fall back to the earliest row when there is a tie or no exact consensus match
- Preserve behavior for standalone training CSVs that do not include annotation export identifiers.

### Tests

- Update transform tests for retrieval `split_group` mapping.
- Add unit tests covering:
  - retrieval train split-group output
  - retrieval training without `record_uuid`
  - duplicate consolidation by strict consensus match
  - duplicate consolidation fallback on ties
  - retrieval duplicate grouping by `record_uuid` and `chunk_id`
  - reserved `split_group` collision rejection

### DIFF
--- a/src/pragmata/core/eval/transforms.py
+++ b/src/pragmata/core/eval/transforms.py
@@ -1,11 +1,14 @@
 """Transforms from pragmata eval frames to tlmtc-compatible frames."""
 
+import logging
 from typing import Literal
 
 import pandas as pd
 
 from pragmata.core.schemas.annotation_task import Task
 from pragmata.core.schemas.eval_input import LABEL_COLUMNS_BY_TASK, TEXT_COLUMNS_BY_TASK
+
+logger = logging.getLogger(__name__)
 
 _DUPLICATE_KEY_COLUMNS_BY_TASK: dict[Task, tuple[str, ...]] = {
     Task.RETRIEVAL: ("record_uuid", "chunk_id"),
@@ -145,5 +148,16 @@ def _consolidate_training_rows(
 
     for target_position, label_override in label_overrides_by_target.items():
         consolidated.loc[target_position, label_override.index] = label_override
+
+    logger.info(
+        "Consolidated duplicate eval training rows for %s: input_rows=%d output_rows=%d "
+        "collapsed_rows=%d duplicate_units=%d key_columns=%s",
+        task.value,
+        len(frame),
+        len(consolidated),
+        len(frame) - len(consolidated),
+        duplicate_groups.ngroups,
+        key_columns,
+    )
 
     return consolidated.reset_index(drop=True)
--- a/src/pragmata/core/eval/transforms.py
+++ b/src/pragmata/core/eval/transforms.py
@@ -7,6 +7,12 @@ import pandas as pd
 from pragmata.core.schemas.annotation_task import Task
 from pragmata.core.schemas.eval_input import LABEL_COLUMNS_BY_TASK, TEXT_COLUMNS_BY_TASK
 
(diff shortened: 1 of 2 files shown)


src/pragmata/core/eval/transforms.py
```diff
@@ -44,11 +55,54 @@ def build_tlmtc_frame(
...
+
+    for _, group in working.groupby(list(key_columns), sort=False, dropna=False):
+        selected_position = group.index[0]
+
+        if len(group) > 1:
+            labels = working_labels.loc[group.index]
+            positive_counts = labels.sum(axis=0)
+            majority_threshold = len(group) / 2
+
+            has_tie = positive_counts.eq(majority_threshold).any()
```
henrycgbaker: computed across whole label vector -> tie on any single label discards consensus selection for all labels (eg 3-1 majority on `proper_action` thrown away because `helpful` happens to be 2-2). Consider taking the majority independently per-label, rather than across whole label vector?

src/pragmata/core/eval/transforms.py
```diff
@@ -44,11 +55,54 @@ def build_tlmtc_frame(
...
+            if not has_tie:
+                majority_vector = positive_counts.gt(majority_threshold).astype("int64")
+                matching_rows = labels.eq(majority_vector, axis=1).all(axis=1)
+
+                if matching_rows.any():
+                    selected_position = matching_rows.idxmax()
+
+        kept_positions.append(selected_position)
+
+    return frame.iloc[kept_positions].copy()
```
henrycgbaker: silently drops rows (n annotator rows -> 1 per unit) w/ no record of how many were collapsed. Add a log/logger.info of input vs output row count? surfaces upstream duplication anomalies instead of hiding them

src/pragmata/core/eval/transforms.py
```diff
@@ -7,6 +7,12 @@
 from pragmata.core.schemas.annotation_task import Task
 from pragmata.core.schemas.eval_input import LABEL_COLUMNS_BY_TASK, TEXT_COLUMNS_BY_TASK
 
+_DUPLICATE_KEY_COLUMNS_BY_TASK: dict[Task, tuple[str, ...]] = {
```
henrycgbaker, line 13: re-encodes the canonical annotation-unit key  (`record_uuid, chunk_id` for retrieval, `record_uuid` otherwise) = already lives in `iaa_runner._item_key` (string form) and in `record_builder`. Worth abstracting into shared constant?

henrycgbaker (approved):
Some minor comments/suggestions

src/pragmata/core/eval/transforms.py
saschagobel: Good point. I agree this should probably be a shared annotation-unit abstraction, but I'd prefer to keep it out of this PR because it would touch eval, IAA, and record-building code. I'll leave the local mapping here for now. The shared constant/helper should be handled in a follow-up PR.

PR #258, opened 2026-06-23
feat: add curated tlmtc train settings

saschagobel:
### Summary

Adds `pragmata`-level defaults for the small set of `tlmtc` training settings that should differ from plain `tlmtc` defaults for eval evaluator training.

These settings are exposed directly because they define the intended `pragmata` evaluator setup: multilingual long paired-text classification with EuroBERT checkpoints. The broader `tlmtc` training surface remains available through `train_kwargs`, so `pragmata` does not mirror every `tlmtc` argument.

### Key changes

- Add task-dependent default `target_name` values:
  - retrieval: `Retrieval evaluation`
  - grounding: `Grounding evaluation`
  - generation: `Generation evaluation`
- Set `pragmata` evaluator checkpoint defaults:
  - `checkpoint`: `EuroBERT/EuroBERT-610m`
  - `proxy_checkpoint`: `EuroBERT/EuroBERT-210m`
- Enable `scale_learning_rate` by default because the proxy and target checkpoints differ.
- Set `sequence_length` to `1024` for long paired RAG inputs.
- Enable `trust_remote_code` by default for EuroBERT compatibility.
- Keep all remaining `tlmtc` train options in `train_kwargs` rather than exposing them as first-class `pragmata` settings.

The reserved-kwargs merge and collision handling for these curated settings will be implemented in a follow-up adapter PR.

### Tests

- Update eval settings tests for the new train defaults.
- Cover task-derived `target_name` behavior.
- Cover explicit overrides for the newly exposed train settings.

### DIFF
--- a/src/pragmata/core/settings/eval_settings.py
+++ b/src/pragmata/core/settings/eval_settings.py
@@ -63,7 +63,7 @@ class EvalTrainSettings(ResolveSettings):
     train_kwargs: dict[str, Any] = Field(default_factory=dict)
 
     @model_validator(mode="after")
-    def default_target_name(
+    def _default_target_name(
         self,
     ) -> Self:
         """Fill the task-specific target name when the caller omits one."""
--- a/src/pragmata/core/settings/eval_settings.py
+++ b/src/pragmata/core/settings/eval_settings.py
@@ -1,13 +1,19 @@
 """Evaluation workflow settings."""
 
 from pathlib import Path
-from typing import Any
+from typing import Any, Self
 from uuid import uuid4
 
-from pydantic import Field
+from pydantic import Field, PositiveInt, model_validator
 
 from pragmata.core.schemas.annotation_task import Task
 from pragmata.core.settings.settings_base import ResolveSettings
+
+DEFAULT_EVAL_TRAIN_TARGET_NAMES: dict[Task, str] = {
+    Task.RETRIEVAL: "Retrieval evaluation",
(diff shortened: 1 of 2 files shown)


lramir14 (approved):
Thanks for the nice work. Everything looks consistent, with one minor doubt. When we use: ´@model_validator(mode="after")
def default_target_name(self) -> Self: ´ 

The new default_target_name validator is declared without a leading underscore, then this exposes it as part of the public API of evaltrainsettings, and it seems inconsistent with the convention used in annotation_settings.py when using model validator. Would it make sense to use the underscore for consistency or was it by design?

saschagobel:
> Thanks for the nice work. Everything looks consistent, with one minor doubt. When we use: ´@model_validator(mode="after") def default_target_name(self) -> Self: ´
> 
> The new default_target_name validator is declared without a leading underscore, then this exposes it as part of the public API of evaltrainsettings, and it seems inconsistent with the convention used in annotation_settings.py when using model validator. Would it make sense to use the underscore for consistency or was it by design?

Good point on consistency. This was not intended as public API: Pydantic does not expose the validator as a model field or in `model_dump()`, so the configurable settings surface is unchanged. But since the validator is an implementation detail and `annotation_settings.py` uses underscore-prefixed validator names, I agree it is cleaner to rename it to `_default_target_name`.

lramir14 (approved):
Thanks for your reply and the change. I see no other issues. Approved.

PR #259, opened 2026-06-23
feat: add tlmtc train adapter

saschagobel:
### Summary

Adds the eval training boundary for calling `tlmtc.train_tlmtc`.

The adapter exposes a `pragmata`-facing `train_evaluator(...)` function, merges dedicated `pragmata` train settings with `train_kwargs`, validates reserved keyword collisions, and imports `tlmtc` lazily so the optional eval dependency is only required when training is invoked.

### Key changes

- Add `src/pragmata/core/eval/tlmtc_adapters.py`.
- Add `train_evaluator(...)` as the internal Pragmata call boundary for evaluator training.
- Forward dedicated train arguments to `tlmtc.train_tlmtc`:
  - `raw_csv`
  - `work_dir`
  - `target_name`
  - `checkpoint`
  - `proxy_checkpoint`
  - `scale_learning_rate`
  - `sequence_length`
  - `trust_remote_code`
- Preserve `train_kwargs` as the pass-through surface for other `tlmtc` train options.
- Reject `train_kwargs` keys that would override dedicated Pragmata-managed arguments.
- Import `tlmtc` lazily and surface an install hint when the eval extra is missing.

### Tests

- Add adapter tests covering:
  - merging dedicated train args with `train_kwargs`
  - forwarding the merged call to `tlmtc.train_tlmtc`
  - returning the raw `tlmtc.train_tlmtc` result
  - rejecting every dedicated adapter argument when repeated in `train_kwargs`
  - lazy import behavior when `tlmtc` is missing

### DIFF
--- a/src/pragmata/core/eval/tlmtc_adapters.py
+++ b/src/pragmata/core/eval/tlmtc_adapters.py
@@ -1,0 +1,83 @@
+"""tlmtc call boundary for eval workflows."""
+
+from pathlib import Path
+from typing import Any
+
+_RESERVED_TRAIN_KWARGS = frozenset(
+    {
+        "raw_csv",
+        "work_dir",
+        "target_name",
+        "checkpoint",
+        "proxy_checkpoint",
+        "scale_learning_rate",
+        "sequence_length",
+        "trust_remote_code",
+    }
+)
+
+
+def train_evaluator(
+    *,
+    raw_csv: Path,
+    work_dir: Path,
+    target_name: str,
+    checkpoint: str,
+    proxy_checkpoint: str,
+    scale_learning_rate: bool,
+    sequence_length: int,
+    trust_remote_code: bool,
+    train_kwargs: dict[str, Any],
+) -> Any:
+    """Train a pragmata evaluator model through tlmtc.
+
+    Args:
+        raw_csv: tlmtc-compatible labeled training CSV.
+        work_dir: Base eval work directory passed through to tlmtc.
+        target_name: Display name used in tlmtc logs and reports.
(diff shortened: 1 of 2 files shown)


src/pragmata/core/eval/tlmtc_adapters.py
```diff
@@ -0,0 +1,81 @@
...
+    Raises:
+        ImportError: If tlmtc is not installed.
+        ValueError: If ``train_kwargs`` attempts to override a dedicated
+            Pragmata-owned train argument.
+    """
+    overlapping_keys = _RESERVED_TRAIN_KWARGS.intersection(train_kwargs)
+    if overlapping_keys:
+        overlapping = ", ".join(sorted(overlapping_keys))
+        raise ValueError(
+            f"train_kwargs must not override core tlmtc train settings: {overlapping}. "
```
lramir14: In this docstring it seems like tlmtc is enforcing the boundary but if I understand correctly pragmata is the one doing this. So perhaps the docstring could be "train_kwargs must not override pragmata-managed train settings: {overlapping}."

lramir14 (approved):
Thanks for your work, well implemented and approved. Two minor comments:

_RESERVED_TRAIN_KWARGS and the train_evaluator signature need to be kept in sync manually — worth documenting to make future maintainers aware of the coupling.

And I think we should consider adding a test for empty train_kwargs that asserts captured_kwargs contains exactly the eight dedicated arguments. The existing merge test adds run_id, batch_size, and verbosity on top, so a accidentally dropped or added key in train_args would go unnoticed.

PR #260, opened 2026-06-24
docs(eval): eval scoring design doc

henrycgbaker:
## Summary

Design doc for the **eval scoring layer**: turning row-level label columns into per-query values and corpus metrics with a confidence interval on every reported metric, exposed via the public API + a new CLI subcommand, reusing the existing IAA uncertainty machinery.

Adds `docs/design/eval-scoring-metrics.md` and an index entry in `docs/design/README.md`. Docs-only - no code changes.

## Proposed shape

- **`core/annotation/uncertainty.py`** (new) — shared `wilson_interval` + `percentile_bootstrap`, owned by the annotation domain (where `bootstrap_alpha` already lives) and imported from eval; refactor `iaa.py:bootstrap_alpha` to reuse.
- **`core/eval/metrics.py`** (new) — pure, I/O-free per-query metric functions matching the taxonomy, plus corpus aggregation.
- **`pragmata.eval.score(...)`** (in existing `api/eval.py`) — the single orchestration point: resolve settings/paths, select input, read + validate, group by query, compute point + CI per metric, assemble the report, write `*_scores.json`. No separate core `run_scoring` layer.
- **`pragmata eval score`** — new subcommand on the existing `eval` Typer group; a thin 1:1 wrapper over the API.
- **Output contract** — nested `MetricScore` (point + CI + method + n) per metric, replacing bare `Rate` floats; reports also record run-level `ci_level` and `n_bootstrap_resamples`.
- **Input contract** — dedicated `*_SCORE_SCHEMA` (not the train schemas); retrieval requires `record_uuid`/`rank`/`chunk_id`, grounding/generation require `record_uuid`.

Uncertainty: Wilson for the 11 proportions (10 unconditional + `conditional_fabrication_rate`), query-level percentile bootstrap for the 5 continuous metrics; query is the resampling (cluster) unit. CIs estimate sampling uncertainty over query examples only — not annotator disagreement, model label error, or benchmark bias.

Input selection has three mutually-exclusive modes — `labeled_input_path`, annotation `export_id`, prediction `prediction_run_id`; `score_id` is an output identifier only. `K` is inferred from the ranked rows, not exposed as a parameter.

## Status

Revised per review. Remaining open decisions are called out in the doc: input precedence/fallback, incomplete/degenerate-data policy, and the `find_latest_prediction_run` resolver.


docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
+# Eval scoring layer: translate labels into corpus metrics with uncertainty
+
+## Summary
+
+Builds the eval scoring stage 
+- turns per-row labels (either manually annotated or evaluator-predicted) into corpus metrics defined in the [metrics taxonomy](../methodology/metrics-taxonomy.md)
```
saschagobel: "per-row labels" is a bit misleading as the metric grain is query/example-level. I'd rephrase to "turn row-level label columns into per-query values and corpus-level metrics", with retrieval called out as chunk rows grouped by query.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
+# Eval scoring layer: translate labels into corpus metrics with uncertainty
+
+## Summary
+
+Builds the eval scoring stage 
+- turns per-row labels (either manually annotated or evaluator-predicted) into corpus metrics defined in the [metrics taxonomy](../methodology/metrics-taxonomy.md) 
+- attach a confidence interval to every reported metric
```
saschagobel: Confidence intervals need a precise scope statement. These CIs estimate sampling uncertainty over query examples; they do not capture annotator disagreement, evaluator/model label error, or benchmark construction bias. That distinction should be explicit before making "CI on every metric" part of the contract.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+ [tlmtc train / predict  (EXTERNAL, ADR-0002)]              → per-row predicted labels
+   ▼
+ ╔════════════════════════════════════════╗
+ ║  SCORING  (per-row labels -> metrics)  ║   MISSING - this issue
+ ╚════════════════════════════════════════╝
+   ▼
+ *_scores.json  (core/schemas/eval_output.py)           NEEDS UPDATING CUFRENTLY:
+                                                              shapes only, bare 0–1 floats,
+                                                              no uncertainty, no compute code
+```
```
saschagobel: This background is outdated. The current flow should start at `annotation export` and then branch: human-labeled exports can go directly into `pragmata eval score`, while evaluator-based scoring goes through `pragmata eval train` followed by `pragmata eval predict` and then `pragmata eval score`. Scoring is the missing piece; train/predict are already in development and should be represented as part of the eval workflow rather than as an external tlmtc-only block.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+
+- report shapes exist (`eval_output.py`) but each metric is a bare `Rate = float[0,1]` and nothing computes them.
```
saschagobel, line 33: This is accurate, but worth phrasing more precisely: paths, settings, and report schemas are already in place; what is missing is the scoring computation itself and the API/CLI wiring around it.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+- no eval CLI (`cli/commands/` has only `annotation.py`, `querygen.py`).
```
saschagobel: Correct, but the eval CLI should be framed as an eval command group. pragmata eval train and pragmata eval predict are expected to land first, and pragmata eval score should extend that surface rather than being designed independently.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+- IAA already reports bootstrap CIs on Krippendorff's alpha - `core/annotation/iaa.py:bootstrap_alpha` (seeded, computes percentiles, configurable `n_resamples`), surfaced through`iaa_report.py` (`ci_lower/ci_upper/ci_level/ n_bootstrap_resamples`), driven by `api/annotation_iaa.py:compute_iaa` -> `iaa_runner.py:run_iaa`.
+
+## Goals
+
+- compute all metrics (retrieval 6, grounding 5, generation 5) from per-row labels.
+- attach CIs (boostrap sampling uncertainty).
+- public API (`score_eval`??)
```
saschagobel: I would avoid `score_eval` as the public user-facing API name. Since eval is the tool namespace, `from pragmata.eval import score` / `pragmata.eval.score(...)` is the cleaner shape.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+- CLI (`pragmata eval score`??)
+- shared `core/stats.py` helpers, w/ IAA refactored to reuse
```
saschagobel: A shared helper may make sense, but `core/stats.py` feels too broad for just Wilson intervals and percentile bootstrap. I'd either narrow the name, or place the helper under core/eval or core/annotation and import it from the other side once the actual shared surface is clear.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+- public API (`score_eval`??)
+- CLI (`pragmata eval score`??)
+- shared `core/stats.py` helpers, w/ IAA refactored to reuse 
+
+## Non-goals (named, deferred)
+
+- evaluator label noise: cross-encoder is imperfect; propagating its classification error is only relevant on the predict-from-model path (human-label scoring has none) and is a materially larger design/owned by tlmtc? Deferred.
+- run-to-run comparison / significance: ee emit per-run CIs only; comparing two runs rigorously needs a *paired* difference-CI (overlap of two independent CIs is **not** a
+  significance test). Downstream / out of scope.
+- bias-corrected-&-accelerated intervals: we chose percentile; BCa is the upgrade only if coverage proves poor.
```
saschagobel: The non-goals read a bit ad hoc. I'd keep only boundaries that clarify likely implementation ambiguity for this scoring layer. Otherwise this section can grow endlessly with things we are not implementing now.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+
+## Design
+
+### Grain and resampling unit
+
+```
+ corpus metric (reported) <- mean over queries <- per-query value  <─ per-row labels
```
saschagobel: The right-to-left arrows are readable as "is derived from", but pipeline diagrams in this doc would be easier to scan left-to-right. Suggest flipping this to:

per-row labels -> per-query metric values -> mean over queries -> reported corpus metric

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+- **Retrieval:** per-chunk labels -> per-query `@k` scalar -> mean over queries
```
saschagobel: This is directionally correct for retrieval, but "scalar" is a bit underspecified. Each retrieval metric computes its own per-query @K value from that query?s ranked chunks, and the corpus metric is the mean over those per-query values.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+
+- unit everything averages over is **query** 
+- query is also the resampling unit (resample whole queries w/ replacement, keeping each query's chunks attached (cluster
+bootstrap). Resampling chunks would break the `/k` denominator and rank order and understate variance (= the hierarchical caveat in ADR-0012))
+
+### Uncertainty method - per metric
+
+| Family | Metrics | Method | n |
+|---|---|---|---|
+| Proportion (binary per query) | all grounding (5), all generation (5), `sufficiency_hit_at_k` | **Wilson** closed-form | #queries |
```
saschagobel: This row seems to double-count conditional_fabrication_rate: it says "all grounding (5)" here, then gives conditional fabrication its own row below. I'd make this "non-conditional grounding metrics (4), all generation metrics (5), sufficiency_hit_at_k".

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+| Conditional proportion | `conditional_fabrication_rate` | **Wilson** on the cited subset | #cited queries |
+| Continuous mean over queries | `ndcg_at_k`, `mean_reciprocal_rank_at_k`, `topical_precision_at_k`, `sufficiency_rate_at_k`, `misleading_context_rate_at_k` | **query-level percentile bootstrap** | #queries |
+
+NB: ^^^`method` is recorded per metric. Wilson chosen for proportions because it is well-behaved at small n (cf wioth bootstrapping a proportion degenerates); bootstrap for the
+continuous metrics because they have no clean closed form.
+
+### New module: `core/stats.py`
```
saschagobel: I'm not yet convinced this belongs as `core/stats.py`. The proposed shared surface is currently just uncertainty helpers for IAA/eval reporting, and `stats` is a very broad module name. The comparison to `atomic_io.py`, `csv_io.py`, and `types.py` does not quite settle the ownership question. Those are broadly infrastructural; Wilson intervals and percentile bootstrap are more specifically reporting/uncertainty utilities. I'd keep this under one owning domain (`core/annotation` or `core/eval`) and import it from the other, at least until a broader core-level stats surface emerges.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+    *, n_resamples: int = 1000, ci: float = 0.95, seed: int | None = None,
+) -> tuple[float, float]:
+    """Percentile bootstrap CI. Resamples unit indices [0, n_units) with replacement,
+    applies `statistic` to each resample, drops NaN replicates (as IAA does)."""
+```
+
+
+### Output contract - `core/schemas/eval_output.py`
+
+Introduce nested per-metric model (JSON):
```
saschagobel: I like replacing bare `Rate` values with a nested `MetricScore`. It makes the output contract much more useful for downstream consumers because the point estimate, uncertainty method, and effective denominator travel together.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+
```
saschagobel, line 92: If we add per-metric CIs, the score report schemas should also record the run-level uncertainty settings. `ci_level` applies to every metric, and `n_bootstrap_resamples` applies to metrics with `method == "percentile_bootstrap"`. These should be added to `RetrievalScoreReport`, `GroundingScoreReport`, and `GenerationScoreReport` so the JSON artifacts are self-describing.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+  matching the taxonomy formulas, plus corpus aggregation.
+- **`core/eval/scoring.py`** (new) - orchestration, mirrors `iaa_runner.py:run_iaa`: validate -> group by query -> compute point + CI per metric (Wilson or bootstrap) -> assemble the task report -> write JSON.
+
+```python
+def run_scoring(
+    frame: pd.DataFrame, *, task: Task,
+    top_k: int | None = None, n_resamples: int = 1000, ci: float = 0.95, seed: int | None = None,
+) -> RetrievalScoreReport | GroundingScoreReport | GenerationScoreReport: ...
+```
+^^^ pure compute - takes data itself
```
saschagobel: `core/eval/metrics.py` is the right scope for the pure metric computations: per-query formulas, corpus aggregation, and any small metric-specific helpers. Keeping this deterministic and I/O-free will make it straightforward to test

I'm skeptical of adding `core/eval/scoring.py` as a second orchestration layer. The described responsibilities (validate -> group by query -> compute point + CI -> assemble report) are close to what the public `pragmata eval score` API must already orchestrate. If core has a `run_scoring()` that does all of this and the API wraps it again, the boundary between business logic and orchestration becomes blurry.

`run_scoring(...)` feels too close to the API surface. I'd prefer smaller core functions with clear ownership: validation in the schema/import layer, query grouping/preparation in a focused eval helper, metric computation in `core/eval/metrics.py`, uncertainty helpers in their own narrow module, and report assembly in the API command path. That keeps the API as the orchestration point instead of wrapping a second orchestration abstraction.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+### Public API - `api/eval_score.py` (new), mirrors `compute_iaa`
```
saschagobel: As noted above, once `pragmata eval train` and `pragmata eval predict` are exposed through `api/eval.py`, scoring should extend that same eval API surface as a thin orchestrator. I don't think this needs a separate `api/eval_score.py` module unless the implementation grows enough to justify splitting internals later.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+
+
+the i/o wrapper:
+
+```python
+def score_eval(
```
saschagobel: I'd name this `score`, not `score_eval`, if it lives on the `eval` API surface. The tool namespace already provides the context: `pragmata.eval.score(...)`.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+    *, base_dir: str | Path | Unset = UNSET,
+    score_id: str | None = None,
+    labeled_input_path: str | Path | None = None,   # Needed?
```
saschagobel: `labeled_input_path` is needed for the direct scoring branch: annotation exports or externally prepared labeled data can go straight into scoring without going through prediction. We may also want an export_id convenience selector for annotation exports, but that should not replace direct path input if external labeled data should remain scoreable.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+    prediction_run_id: str | None = None,    # Needed?
```
saschagobel: `prediction_run_id` is needed for the branch where scoring consumes labels produced by `pragmata eval predict`. This should refer to a `pragmata` prediction run, even if the underlying prediction output is tlmtc-managed. `EvalScoreSettings` already documents "latest prediction run" fallback, but that resolver does not exist yet; it could be specified/implemented in eval_paths.py, probably following the same pattern as `find_latest_annotation_export_id` in `eval_paths.py`.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+the i/o wrapper:
+
+```python
+def score_eval(
+    *, base_dir: str | Path | Unset = UNSET,
+    score_id: str | None = None,
+    labeled_input_path: str | Path | None = None,   # Needed?
+    prediction_run_id: str | None = None,    # Needed?  
+    task: Task,
+    top_k: int | None = None,
```
saschagobel: Why is `top_k` an API parameter here? The labeled retrieval data should already represent a concrete cutoff, so `K` should be inferred from the ranked rows in the scoring input. Post-hoc truncation would change the evaluated artifact and bias the metric, so I would avoid exposing `top_k` unless there is a separate, clearly justified use case.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+    n_resamples: int = 1000, ci: float = 0.95, seed: int | None = None,
+    config_path: str | Path | Unset = UNSET,
+) -> RetrievalScoreReport | GroundingScoreReport | GenerationScoreReport:
+    """Resolve EvalScoreSettings + score paths, read the labeled/predicted frame,
+run_scoring inside error_log, write *_scores.json, return the report."""
```
saschagobel: As described above: This wrapper should be the orchestration point, not just a shell around `run_scoring`.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+Reuses `EvalScoreSettings` (`settings/eval_settings.py`), `resolve_eval_score_paths`
+(`paths/eval_paths.py`), `validate_eval_score_frame` (`eval_input.py`), `error_log`.
+
+### CLI - `cli/commands/eval.py` (new) + wire into `cli/app.py`
+
+- typer app (same as annotatiojn & querygen)
+- subcommand `score` ??
```
saschagobel: Agree with adding an eval Typer app and a `score` subcommand. That mirrors the intended Python API shape (`eval.score`) and fits the expected `pragmata eval train/predict/score` command group.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+
+```
+pragmata eval score --task retrieval \
+  [--import-run-id ID ???] [--export-run-id ID] \
+  [--base-dir DIR] \
+  [--n-resamples 1000] [--ci 0.95] [--seed N] [--config FILE]
```
saschagobel, line 163: The CLI arguments should mirror the Python API rather than introduce a separate shape. If the CLI exposes `--ci` (`--ci-level`?) and `--seed`, those should also be explicit API parameters, since the CLI should be a thin wrapper around the API.

Input selection should reflect the two scoring branches discussed above: direct labeled input versus prediction-output scoring. Concretely, I'd expect selectors along the lines of `--labeled-input-path`, `--export-id`, and `--prediction-run-id`.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+```
+
+- run-id here would be the annotation's export_id = input selector. Would need to considr more how this takes human-labeled vs predicted-labeled (see the above comments in the api block)
```
saschagobel: Yes, this needs more explicit input-mode design. Human-labeled scoring can come from a direct `labeled_input_path` or annotation `export_id`; prediction-based scoring should come from `prediction_run_id`. Those are distinct modes and should not be collapsed into a generic run-id.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+- score/export-id would be where the results are written  = this runs handle... could remove?
```
saschagobel: I think this is mixing input and output identifiers. Annotation `export_id` / prediction `run_id` select the input; `score_id` identifies where score artifacts are written under `eval/scores/<score_id>/`. I'd keep `score_id` as an optional output identifier, defaulting to the generated value already provided by `EvalScoreSettings`, and avoid using `export_id` for output naming.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+- decide the above api / cli shapes
+- input-contract dependency (decision(s) needed):
+  - retrieval scoring contract currently lacks the columns the `@k` metrics need. `RETRIEVAL_*_SCHEMA` (`eval_input.py`) is `(query, chunk, labels)`; `strict=False` = lets extras pass but nothing is required/ordered. To score retrieval we need, per row:
+    - `record_uuid` - group chunks into queries (the per-query unit).
+    - `rank` (1...K) - required by NDCG@K and MRR@K and any top-K truncation.
+    - `chunk_id` - stable identity within a query (already a dup-key in `transforms.py`).
+- So extend the retrieval scoring schema to 
+  1. require `record_uuid`, `rank`, `chunk_id` (w/o `rank`,NDCG/MRR cannot be computed)
+  2.  define how `K` is set (explicit `--top-k` vs inferred per-query chunkcount)???.
+- grounding/generation need `record_uuid`
```
saschagobel: I'd make this a scoring-specific input contract rather than extending the train schemas. validate_eval_score_frame currently reuses _TRAIN_SCHEMAS_BY_TASK, but scoring has stricter structural requirements than training: retrieval needs record_uuid, rank, and chunk_id, while grounding/generation need record_uuid. I suggest adding dedicated *_SCORE_SCHEMA definitions in eval_input.py and have validate_eval_score_frame use those, so training inputs are not unnecessarily constrained by scoring-only metadata.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+  [--base-dir DIR] \
+  [--n-resamples 1000] [--ci 0.95] [--seed N] [--config FILE]
+```
+
+- run-id here would be the annotation's export_id = input selector. Would need to considr more how this takes human-labeled vs predicted-labeled (see the above comments in the api block)
+- score/export-id would be where the results are written  = this runs handle... could remove?
+
+### TODO
+
+- decide the above api / cli shapes
```
saschagobel: After the API/CLI discussion above, this should no longer be an open TODO. The design should settle on one eval API/CLI shape and document the input selectors and precedence rules explicitly.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+- input-contract dependency (decision(s) needed):
+  - retrieval scoring contract currently lacks the columns the `@k` metrics need. `RETRIEVAL_*_SCHEMA` (`eval_input.py`) is `(query, chunk, labels)`; `strict=False` = lets extras pass but nothing is required/ordered. To score retrieval we need, per row:
+    - `record_uuid` - group chunks into queries (the per-query unit).
+    - `rank` (1...K) - required by NDCG@K and MRR@K and any top-K truncation.
+    - `chunk_id` - stable identity within a query (already a dup-key in `transforms.py`).
+- So extend the retrieval scoring schema to 
+  1. require `record_uuid`, `rank`, `chunk_id` (w/o `rank`,NDCG/MRR cannot be computed)
+  2.  define how `K` is set (explicit `--top-k` vs inferred per-query chunkcount)???.
```
saschagobel: `K` should be inferred from the ranked rows in the scoring input, not supplied as `--top-k`. Exposing `top_k` suggests a tunable choice where the scored artifact already has a fixed cutoff, and a wrong or post-hoc value would either bias the metric or force avoidable error handling.

docs/design/eval-scoring-metrics.md
```diff
@@ -0,0 +1,174 @@
...
+pragmata eval score --task retrieval \
+  [--import-run-id ID ???] [--export-run-id ID] \
+  [--base-dir DIR] \
+  [--n-resamples 1000] [--ci 0.95] [--seed N] [--config FILE]
+```
+
+- run-id here would be the annotation's export_id = input selector. Would need to considr more how this takes human-labeled vs predicted-labeled (see the above comments in the api block)
+- score/export-id would be where the results are written  = this runs handle... could remove?
+
+### TODO
```
saschagobel: I'd add two items to this list:

1. input selection semantics. Scoring appears to have three input selectors: direct `labeled_input_path`, annotation `export_id`, and `prediction_run_id`. The design should define which are mutually exclusive, which are required, and what (if any) fallback applies if none is supplied.

2. incomplete and degenerate scoring data. Retrieval metrics assume complete ranked chunk labels per query; if some chunks are unlabeled, we need a deliberate policy: fail with an informative message, skip affected queries, or compute an alternative fallback metric with clear caveats. Similarly, all-0/all-1 or otherwise degenerate labels may make some estimates uninformative and should be handled explicitly. Ideally the validation/guard logic is reusable between eval train and eval score where the constraints overlap.

saschagobel (commented):
Good design doc overall; it attaches well to the existing eval work and the proposed scoring layer is directionally right. My comments are mostly about tightening API/CLI boundaries, input contracts, uncertainty/reporting semantics, and resolving the remaining TODOs/open decisions before implementation.

saschagobel (approved):

PR #261, opened 2026-06-30
refactor: rename tlmtc train adapter

saschagobel:
### Summary

Renames the internal `tlmtc` train adapter so it does not conflict with the upcoming public eval API name.

The public API will expose `train_evaluator(...)`, while the core adapter now uses `run_tlmtc_train(...)` to describe its actual responsibility: invoking `tlmtc.train_tlmtc` with `pragmata`-managed arguments plus passthrough `train_kwargs`.

### Key changes

- Rename `train_evaluator(...)` to `run_tlmtc_train(...)` in `tlmtc_adapters.py`.
- Update the reserved-kwargs sync comment to reference `run_tlmtc_train`.
- Update adapter tests to import and call `run_tlmtc_train`.
- Rename test helpers and signature-derived constants to match the adapter name.
- Keep adapter behavior unchanged.

### Tests

- Update `tlmtc` adapter tests for the new function name.
- Verify reserved-kwargs validation still derives dedicated arguments from the adapter signature.
- Confirm the adapter test suite still passes.

### DIFF
--- a/src/pragmata/core/eval/tlmtc_adapters.py
+++ b/src/pragmata/core/eval/tlmtc_adapters.py
@@ -5,7 +5,7 @@ from typing import Any
 
 import pandas as pd
 
-# Keep this in sync with train_evaluator's dedicated arguments.
+# Keep this in sync with run_tlmtc_train's dedicated arguments.
 _RESERVED_TRAIN_KWARGS = frozenset(
     {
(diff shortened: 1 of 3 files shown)


lramir14 (approved):
Thanks for the nice refactoring. All clean, I dont see any issues. Approved.

PR #262, opened 2026-06-30
fix: ensure train creates eval tool root

saschagobel:
### Summary

Adds the eval tool root to train path resolution so `eval train` can create `<base_dir>/eval` before handing that directory to `tlmtc` as its `work_dir`.

`pragmata` still does not create `tlmtc`-owned train output directories. It only ensures the eval tool root exists; `tlmtc` remains responsible for creating its own artifact structure below that root.

### Key changes

- Add `tool_root` to `EvalTrainPaths`.
- Add `EvalTrainPaths.ensure_dirs()` for creating the eval tool directory scaffold.
- Resolve `tool_root` as `<base_dir>/eval` for:
  - direct labeled training input
  - annotation-export-backed training input
- Keep train input CSV resolution unchanged.

### Tests

- Update eval train path expectations to include `tool_root`.
- Add coverage that `EvalTrainPaths.ensure_dirs()` creates the eval tool root and returns `self`.

### DIFF
--- a/src/pragmata/core/paths/eval_paths.py
+++ b/src/pragmata/core/paths/eval_paths.py
@@ -16,13 +16,23 @@ class EvalTrainPaths:
     """Resolved labeled input path and source provenance for eval train.
 
     Attributes:
+        tool_root: Root directory for eval artifacts. Passed to tlmtc as
+            ``work_dir`` for training.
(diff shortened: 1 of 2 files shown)


lramir14 (approved):
Thanks for the nice work. Approved.

PR #265, opened 2026-07-01
chore: bump tlmtc to 0.3

saschagobel:
### Summary

Bumps the eval extra to require `tlmtc` 0.3 and updates the lockfile accordingly.

This version is needed for the eval train API work because `tlmtc` now supports dataframe input for training and prediction, allowing `pragmata` to pass transformed eval frames directly without writing intermediate CSV files.

### Key changes

- Update the `eval` optional dependency from `tlmtc[full]>=0.2,<1.0` to `tlmtc[full]>=0.3,<1.0`.
- Update `uv.lock` from `tlmtc` 0.2.0 to 0.3.0.
- Preserve the existing `<1.0` upper bound.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ type = [
 ]
 annotation = ["argilla>=2.0,<3.0",]
 querygen = ["sentence-transformers>=3.0,<6.0",]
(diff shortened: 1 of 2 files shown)


lramir14 (approved):
Approved.

PR #266, opened 2026-07-01
feat: add train evaluator API

saschagobel:
### Summary

Adds the public Python API surface for training `pragmata` eval models. `train_evaluator` now resolves eval train settings, selects the labeled input source, imports and validates the task-specific training data, transforms it into a `tlmtc` training frame, and delegates training through the eval `tlmtc` adapter.

### Key changes

- Add `pragmata.api.eval.train_evaluator`.
- Support direct labeled CSV input through `labeled_data_path`.
- Support annotation export input through:
  - explicit `export_id`
  - latest matching annotation export for the selected `task`
- Resolve and create the eval tool root before handing control to `tlmtc`.
- Import and validate the selected task CSV.
- Transform `pragmata` eval columns into `tlmtc` train columns.
- Pass the transformed dataframe directly to `run_tlmtc_train`.
- Forward curated train settings:
  - `target_name`
  - `checkpoint`
  - `proxy_checkpoint`
  - `scale_learning_rate`
  - `sequence_length`
  - `trust_remote_code`
- Preserve `train_kwargs` for all other `tlmtc`-owned training options.
- Add the public `pragmata.eval` namespace.
- Expose `eval` from the top-level `pragmata` package.

### Notes

This PR adds only the Python API. CLI wiring is intentionally left for a follow-up PR.

The API returns the `tlmtc` train result directly, so callers receive the resolved training run paths and generated artifact locations owned by `tlmtc`.

### Tests

- Added unit tests for train evaluator API orchestration.
- Cover direct labeled CSV input.
- Cover config plus explicit override precedence.
- Cover selected annotation export path resolution.

### DIFF
--- a/src/pragmata/__init__.py
+++ b/src/pragmata/__init__.py
@@ -5,9 +5,9 @@ Only curated, stable symbols should be exposed here.
 
 import logging
 
-from . import annotation, querygen
+from . import annotation, eval, querygen
 from .api import get_version
 
-__all__ = ["annotation", "get_version", "querygen"]
+__all__ = ["annotation", "eval", "get_version", "querygen"]
 
 logging.getLogger(__name__).addHandler(logging.NullHandler())
--- a/src/pragmata/api/eval.py
+++ b/src/pragmata/api/eval.py
@@ -64,7 +64,7 @@ def train_evaluator(
     """
     settings = EvalTrainSettings.resolve(
         config=load_config_file(config_path) if isinstance(config_path, (str, Path)) else None,
-        env=None, # Environment-derived settings are not wired for train_evaluator yet.
+        env=None,  # Environment-derived settings are not wired for train_evaluator yet.
         overrides={
             "base_dir": base_dir,
             "labeled_data_path": labeled_data_path,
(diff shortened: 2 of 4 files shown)


lramir14 (approved):
thanks for the nice work. The logic and orchestration look solid. However, I have a doubt: export_annotations and gen_queries both wrap their core execution in error_log(...), so any failure during the run gets captured in errors.log under the tool root. Here, train_evaluator doesn't use error_log anywhere and any failure would not be captured by the errors.log. So I am wondering if this was intentional by design?

saschagobel:
> thanks for the nice work. The logic and orchestration look solid. However, I have a doubt: export_annotations and gen_queries both wrap their core execution in error_log(...), so any failure during the run gets captured in errors.log under the tool root. Here, train_evaluator doesn't use error_log anywhere and any failure would not be captured by the errors.log. So I am wondering if this was intentional by design?

Good catch. `error_log(...)` currently only attaches an ERROR+ file handler for `pragmata.*` log records while the wrapped block is active. It does not log escaping exceptions by itself.

For `train_evaluator`, adding the wrapper would therefore be a no-op at the moment: the code called inside this path raises exceptions directly, but does not emit `logger.error(...)` / `logger.exception(...)` records that `error_log(...)` could capture. So this would only be a cosmetic consistency change.

I think evaluator training error logging should be handled more comprehensively in a follow-up PR.

PR #267, opened 2026-07-01
test: add train evaluator integration coverage

saschagobel:
### Summary

Adds integration coverage for the public eval train API.

The new tests exercise `pragmata.eval.train_evaluator` through the real import, transform, adapter, and `tlmtc` training path using `tlmtc`. The fixture data uses realistic retrieval-style query/chunk pairs with coherent multilabel annotations.

### Key changes

- Add eval train integration tests under `TestTrainEvaluator`.
- Cover direct labeled CSV input.
- Cover explicit annotation export selection via `export_id`.
- Cover latest annotation export selection for the requested task.
- Verify returned `tlmtc` result paths point to persisted train artifacts.
- Run one tiny transfer-learning smoke test with:
  - `google/bert_uncased_L-2_H-128_A-2`
  - `train_epochs=1`
  - `hyperparameter_tuning=False`
  - `threshold_optimization=False`
- Keep export-selection tests fast with `transfer_learning=False`.
- Register the `eval` pytest marker.
- Include the eval extra in the dev environment and update lock metadata.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -42,7 +42,7 @@ anthropic = ["langchain-anthropic>=1.4,<2.0"]
 google-genai = ["langchain-google-genai>=4.2,<5.0"]
 full = ["pragmata[annotation,querygen,eval]"]
 everything = ["pragmata[full,cohere,deepseek,openai,anthropic,google-genai]"]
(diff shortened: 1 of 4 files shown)


lramir14 (approved):
Thanks for the work, the integration coverage is thorough. Approved.

PR #268, opened 2026-07-01
feat(annotation): wire completeness into export (columns + meta summary)

henrycgbaker:

**This PR (3/6):** Split out of #247 (was 1001 lines / XL). Wires the panel-completeness calculator (#247) into the retrieval export path. Stacked on #247.

---

### Schema
- `core/schemas/annotation_export.py`:
  - `RetrievalExportRow` += `panel_complete: bool`, `n_annotated_chunks: int`, `n_submitted_chunks: int`, `n_discarded_chunks: int`, `n_records_seen: int`. All defaulted; appended at the tail so existing column order is preserved.
  - New `KBucketStat` (per-K-bucket panel counts) and `CompletenessSummary` (aggregates + `by_k_bucket` cross-tab + full per-K histogram) Pydantic models.
  - `AnnotationExportMeta` += `completeness_summary: CompletenessSummary | None` and `completeness_status: Literal["ok","failed","not_requested"]`.

### Single-walk retrieval pipeline
- `core/annotation/export_fetcher.py`: new `walk_retrieval_records` (one Argilla scroll per dataset, no status filter — sees discards-only records too so the integrity check has the full record set) + `RetrievalRecordSnapshot` dataclass + `fetch_retrieval_from_records` (status filter applied in Python at typed-row construction).
- `core/annotation/export_runner.py`: when `Task.RETRIEVAL in tasks`, walks once and feeds both the typed-row projection and the completeness aggregator. `compute_completeness` failures degrade the sidecar (sets `completeness_status="failed"`) without losing the already-fetched export.

### API
- `annotation/__init__.py`: lazy re-exports for `CompletenessSummary`, `KBucketStat`.

### Tests
- `test_export_runner.py`: `write_export_csv` stamps completeness columns by `record_uuid`; sidecar carries `completeness_summary`; `completeness_status` ok/failed/not_requested set explicitly.
- `test_annotation_export.py`: `RetrievalExportRow` field-order locked.

## Why

The retriever is threshold-based, so dropping partial panels biases the metrics toward small K (MNAR). Surfacing both per-row predicate columns AND a bucketed sidecar aggregate lets the eval pipeline either filter complete panels per-K or apply condensed-list scoring without re-querying the data layer.

The single-walk shape (one retrieval scroll feeds both fetch_task and compute_completeness) costs one extra in-memory pass over already-fetched records but spares the Argilla server a second full scroll per export — meaningful on the in-flight batches (~hundreds of panels).

The STRICT `panel_complete` policy question (which shape to ship — this PR assumes (A) from #247) is discussed on #247, since that's where the predicate is actually computed; this PR just surfaces it as a column.

## Test plan
- [ ] `uv run python -m pytest tests/unit/core/annotation/test_export_runner.py tests/unit/core/schemas/test_annotation_export.py`

### DIFF
--- a/src/pragmata/core/annotation/export_fetcher.py
+++ b/src/pragmata/core/annotation/export_fetcher.py
@@ -180,6 +180,60 @@ def walk_retrieval_records(client: rg.Argilla, settings: AnnotationSettings) -> 
     return snapshots
 
 
+def fetch_retrieval_from_records(
+    snapshots: list[RetrievalRecordSnapshot],
+    user_lookup: dict[UUID, str],
+    *,
+    include_discarded: bool,
+) -> list[tuple[AnnotationModel, list[LogicalConstraint]]]:
+    """Project pre-walked retrieval snapshots into typed rows.
+
+    The status filter is applied in Python (per user-response) since the
+    walk skipped the query-level filter to feed the completeness pass too.
+    Mirrors ``fetch_task``'s retrieval branch but reuses ``_build_row`` so
+    typed-row construction has one source of truth.
+    """
+    accepted = {"submitted", "discarded"} if include_discarded else {"submitted"}
+    check_constraints = CONSTRAINT_CHECKERS[Task.RETRIEVAL]
+    rows: list[tuple[AnnotationModel, list[LogicalConstraint]]] = []
+    missing_uuid_count = 0
+    for snap in snapshots:
+        if not snap.record_uuid:
+            missing_uuid_count += 1
+            # Mirror fetch_task: keep going, the row still gets emitted with
+            # an empty record_uuid (the export uses this signal too).
+        record = snap.record
+        created_at: datetime = record._model.updated_at or record._model.inserted_at
+        inserted_at: datetime = record._model.inserted_at
+        language: str | None = record.metadata.get("language")
+        record_status: str = record.status
+        for user_id, response_status, answers in snap.response_user_pairs:
+            if response_status not in accepted:
+                continue
+            base = {
+                "record_uuid": snap.record_uuid,
+                "annotator_id": user_lookup.get(user_id, str(user_id)),
+                "language": language,
+                "calibration": snap.calibration,
+                "inserted_at": inserted_at,
+                "created_at": created_at,
+                "record_status": record_status,
+                "response_status": response_status,
+                "discard_reason": answers.get("discard_reason"),
+                "discard_notes": answers.get("discard_notes"),
+                "notes": answers.get("notes") or "",
+            }
+            row = _build_row(Task.RETRIEVAL, base=base, answers=answers, fields=record.fields, metadata=record.metadata)
+            violations = check_constraints(row) if response_status == "submitted" else []
+            rows.append((row, violations))
+    if missing_uuid_count:
+        logger.warning(
+            "task=retrieval: %d record(s) missing record_uuid metadata — included with empty string",
+            missing_uuid_count,
+        )
+    return rows
+
+
 def fetch_task(
     client: rg.Argilla,
     settings: AnnotationSettings,
--- a/src/pragmata/core/annotation/export_runner.py
+++ b/src/pragmata/core/annotation/export_runner.py
@@ -6,16 +6,28 @@ import logging
 from dataclasses import dataclass
 from datetime import UTC, datetime
 from pathlib import Path
-from typing import TYPE_CHECKING
+from typing import TYPE_CHECKING, Any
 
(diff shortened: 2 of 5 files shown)


src/pragmata/core/annotation/export_runner.py
```diff
@@ -198,22 +239,49 @@ def run_export(
...
+        else:
+            task_rows[task] = fetch_task(client, settings, task, user_lookup, include_discarded=include_discarded)
+
+    completeness_report: CompletenessReport | None = None
+    completeness_status = "not_requested"
+    if retrieval_snapshots is not None:
+        try:
+            completeness_report = compute_completeness_from_records(retrieval_snapshots)
+            completeness_status = "ok"
+        except Exception:
```
ddimmery, line 274: The generic exception should be made more specific (e.g. `KeyboardInterrupt` gets captured by the base `Exception` class).

src/pragmata/core/schemas/annotation_export.py
```diff
@@ -92,10 +92,35 @@ class GenerationAnnotation(AnnotationBase):
...
+    # n_submitted_chunks AND n_discarded_chunks. So
+    # n_submitted + n_discarded >= n_annotated (equality when no mixed chunks).
+    # n_mixed_chunks = n_submitted + n_discarded - n_annotated.
+    n_annotated_chunks: int = 0  # union: chunks with any terminal response
+    n_submitted_chunks: int = 0  # chunks with any submitted response (used by panel_complete)
+    n_discarded_chunks: int = 0  # chunks with any discarded response
+    # n_records_seen = distinct chunk-records observed for this panel. Used
+    # for the integrity check (records_seen != n_retrieved_chunks → records
+    # lost or duplicated upstream). Surfaced per-row so consumers can spot
+    # corrupted panels in dataframe pipelines without re-joining the sidecar.
```
ddimmery, line 122: If `compute_completeness_from_records` raises an error, this will be indistinguishable from an successful complete with empty data. Would there be a good way to surface this? e.g. a Sentinel value indicating an error?

src/pragmata/core/annotation/export_fetcher.py
```diff
@@ -92,6 +100,142 @@ def _build_row(
...
+    scroll per dataset instead of two.
+    """
+    workspace_name, purposes = resolve_task_purposes(settings, Task.RETRIEVAL)
+    snapshots: list[RetrievalRecordSnapshot] = []
+    for calibration in purposes:
+        ds_name = dataset_name(Task.RETRIEVAL, calibration=calibration, dataset_id=settings.dataset_id)
+        dataset = client.datasets(ds_name, workspace=workspace_name)
+        if dataset is None:
+            continue
+        for record in dataset.records(with_responses=True):
```
ddimmery, line 164: This collects all responses and then filters here, rather than like with grounding/generation which filter with `query = rg.Query(filter=rg.Filter([("response.status", "in", statuses)]))`. Shouldn't matter, but might as well be consistent.

src/pragmata/core/annotation/export_fetcher.py
```diff
@@ -22,6 +23,13 @@
 
 logger = logging.getLogger(__name__)
 
+# A chunk-record's annotation is "terminal" (no longer pending) when at least
+# one of its responses has one of these statuses. Discarded counts as covered
+# for metrics purposes — it's an explicit decision, not a hole. Shared by
+# fetch_task's status-filter callers, completeness, and panel_status so the
+# three never drift on the metric definition.
+TERMINAL_STATUSES = frozenset({"submitted", "discarded"})
```
ddimmery: Is this used? Maybe in later diffs?

src/pragmata/core/annotation/export_runner.py
```diff
@@ -139,6 +171,8 @@ def assemble_export_meta(
     n_annotators: dict[Task, int],
     calibration_enabled: dict[Task, bool],
     constraint_summary: dict[str, int],
+    completeness_summary: CompletenessSummary | None = None,
+    completeness_status: str = "not_requested",
```
ddimmery: I think you can drop the type-checker ignore on line 209 if you type this as `Literal["ok","failed","not_requested"]`

ddimmery (approved):

src/pragmata/core/annotation/export_fetcher.py
henrycgbaker: intentionally different: walk_retrieval_records needs to see unfiltered so the completeness pass sees chunk-records w/ no terminal response (needed fo integrity check and K counting)

PR #269, opened 2026-07-01
feat(annotation): status API + CLI command

henrycgbaker:

**This PR (5/6):** Split out of #248 (was 975 lines / L). Exposes the panel-status core engine (#248) as a CLI command and API function. Stacked on #248.

---

## Scope

- **New `pragmata annotation status` CLI**: reads live retrieval datasets and reports per-panel completeness across prod + cal in one call, plus the all-task record-progress summary (total/completed across retrieval/grounding/generation) as an aligned table, grouped by task by default, with `--by-workspace` / `--by-dataset` for finer breakdowns. Pure read; no Argilla mutations.
- CLI: `--workspace` to scope; credentials via `--api-url/--api-key` or env.
- `api/annotation_status.py`: `report_status()` wraps `compute_panel_status` + `compute_task_progress` for the CLI/API surface.

### API
- `annotation/__init__.py`: lazy re-exports for `HeadlineTotals`, `PanelStatus`, `StatusReport`, `report_status`.

### Tests
- `test_status_api.py`: `report_status()` assembly of panel + progress reports.
- `test_cli_annotation.py`: CLI output formatting, `--workspace` scoping, `--by-workspace`/`--by-dataset` breakdowns.

The `--tag-partial-panels` advisory write ships in #249.

## Test plan
- [ ] `uv run python -m pytest tests/unit/api/test_status_api.py tests/unit/cli/test_cli_annotation.py`

### DIFF
--- a/src/pragmata/annotation/__init__.py
+++ b/src/pragmata/annotation/__init__.py
@@ -15,30 +15,36 @@ from typing import TYPE_CHECKING
 
 __all__ = [
     "ExportResult",
+    "HeadlineTotals",
     "IaaReport",
     "ImportResult",
     "Locale",
     "SetupResult",
+    "StatusReport",
     "Task",
     "UserSpec",
     "compute_iaa",
     "export_annotations",
     "import_records",
+    "report_status",
     "setup",
     "teardown",
 ]
 
 _LAZY: dict[str, tuple[str, str]] = {
     "ExportResult": ("pragmata.core.annotation.export_runner", "ExportResult"),
+    "HeadlineTotals": ("pragmata.core.annotation.panel_status", "HeadlineTotals"),
     "IaaReport": ("pragmata.core.schemas.iaa_report", "IaaReport"),
     "ImportResult": ("pragmata.api.annotation_import", "ImportResult"),
     "Locale": ("pragmata.core.schemas.annotation_task", "Locale"),
     "SetupResult": ("pragmata.core.annotation.setup", "SetupResult"),
+    "StatusReport": ("pragmata.core.annotation.panel_status", "StatusReport"),
     "Task": ("pragmata.core.schemas.annotation_task", "Task"),
     "UserSpec": ("pragmata.core.settings.annotation_settings", "UserSpec"),
     "compute_iaa": ("pragmata.api.annotation_iaa", "compute_iaa"),
     "export_annotations": ("pragmata.api.annotation_export", "export_annotations"),
     "import_records": ("pragmata.api.annotation_import", "import_records"),
+    "report_status": ("pragmata.api.annotation_status", "report_status"),
     "setup": ("pragmata.api.annotation_setup", "setup"),
     "teardown": ("pragmata.api.annotation_setup", "teardown"),
 }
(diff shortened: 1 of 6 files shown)


saschagobel:
`_render_table` feels like generic CLI presentation formatting rather than annotation-command behavior. Since `cli/parsing.py` already keeps reusable CLI helpers outside command modules, maybe this should live in a small `pragmata.cli.formatting` / output helper instead and be imported here. Fine to keep local if this is intentionally private to this one status command, but extracting it would keep the command module boundary cleaner.

src/pragmata/api/annotation_status.py
```diff
@@ -0,0 +1,53 @@
...
+
+    Returns:
+        ``StatusReport`` with the all-task ``progress`` summary plus the
+        retrieval per-panel facts.
+    """
+    url = api_url if isinstance(api_url, str) else os.environ.get("ARGILLA_API_URL")
+    key = api_key if isinstance(api_key, str) else resolve_api_key("argilla")
+    client = resolve_argilla_client(url, key)
+    progress = compute_task_progress(client, workspace=workspace)
+    report = compute_panel_status(client, workspace=workspace).with_progress(progress)
```
ddimmery, line 44: Each of these separately crawl the Argilla data. This shouldn't be a problem, but I suppose you could get a weird race condition that leaves them slightly inconsistent. I think in practice this is only an efficiency issue that isn't really worth fixing unless something comes up.

src/pragmata/annotation/__init__.py
```diff
@@ -17,17 +17,21 @@
...
     "ExportResult",
+    "HeadlineTotals",
     "IaaReport",
     "ImportResult",
     "KBucketStat",
     "Locale",
     "PanelCompleteness",
+    "PanelStatus",
     "SetupResult",
+    "StatusReport",
```
ddimmery: `ProgressReport` and `ProgressRow` should probably also be exported, since they're constituent components of `StatusReport`. Consuming code that works with this might want to use these types, so they should probably be exported.

src/pragmata/cli/commands/annotation.py
```diff
@@ -214,6 +214,102 @@ def export_command(
...
+    # By task (default). The retrieval row carries panel-completeness; other
+    # tasks are single-record and have no panels ("-").
+    task_rows: list[tuple[str, ...]] = []
+    for row in prog.by_task:
+        if row.task == "retrieval":
+            panels = (_num(report.n_panels), _num(report.n_complete), _num(report.n_overlap_satisfied))
+        else:
+            panels = ("–", "–", "–")
+        task_rows.append((row.label, _num(row.total), _num(row.completed), _pct(row.completed, row.total), *panels))
+    for line in _render_table(["TASK", "TOTAL", "COMPLETED", "%", "PANELS", "COMPL", "OVERLAP"], task_rows, "lrrrrrr"):
```
ddimmery: `COMPLETED` versus `COMPL` is very confusing. I would suggest being more specific on the latter with `PANEL-COMPL` or something.

ddimmery (approved):

src/pragmata/annotation/__init__.py
saschagobel: Should we really export `ProgressReport` and `ProgressRow` here?

I’m wondering whether we should be cautious about expanding `pragmata.annotation` further, since the public annotation surface already seems larger than it strictly needs to be. My intuition is that nested/component types should only be re-exported when users are expected to construct them or depend on them directly. Since these are accessible through the returned `StatusReport`, I’d lean toward keeping them internal unless there is a concrete external use case.

PR #270, opened 2026-07-02
feat: add eval train command

saschagobel:
### Summary

Adds CLI support for training `pragmata` eval models.

The new `pragmata eval train-evaluator` command delegates to the public `pragmata.eval.train_evaluator` API, preserving the API-owned orchestration while exposing the train workflow through the Typer CLI.

### Key changes

- Add `src/pragmata/cli/commands/eval.py`.
- Register the `eval` command group on the root CLI app.
- Add `eval train-evaluator` options
- Map omitted options to `UNSET`.
- Parse `--train-kwargs` as JSON through the existing CLI parsing helper.
- Print a concise run summary with the returned `run_id` and `run_directory`.

### Tests

- Add unit tests for the eval train CLI command.
- Cover root `eval` command registration.
- Cover registered `eval train-evaluator --help`.
- Cover omitted-option `UNSET` mapping.
- Cover explicit option forwarding to `pragmata.eval.train_evaluator`.
- Cover boolean flag forwarding.
- Cover JSON parsing for `--train-kwargs`.
- Cover CLI run summary output.

### DIFF
--- a/tests/unit/cli/test_cli_eval.py
+++ b/tests/unit/cli/test_cli_eval.py
@@ -1,6 +1,5 @@
 """Tests CLI commands for evaluation workflows."""
 
-import re
 from pathlib import Path
 from typing import Any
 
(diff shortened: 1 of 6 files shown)


tests/unit/cli/test_cli_eval.py
```diff
@@ -0,0 +1,222 @@
...
+import pytest
+from typer.testing import CliRunner
+
+from pragmata.api import UNSET
+from pragmata.cli.app import app
+from pragmata.cli.commands.eval import eval_app
+
+runner = CliRunner()
+
+ANSI_ESCAPE_RE = re.compile(r"\x1b\[[0-9;]*m")
```
lramir14: this is a copy from test_cli_querygen.py. Would it make sense then to define a shared contest.py file and then import it instead of re-defining this function across several scripts?

lramir14 (approved):
Thank you for the nice work. The only minor comment I have is the ANSI_ESCAPE_RE duplication function here and in test_cli_querygen.py, so would it make sense to pull it into a conftest.py script or a helper module to avoid re-defining? One additional comment: the run summary omits the model output location, so echoing model_dir could potentially be beneficial for the user in case they dont know run_dir internal layout to find the trained model.

tests/unit/cli/test_cli_eval.py
saschagobel: Done

PR #271, opened 2026-07-04
feat: persist pragmata-owned train metadata

saschagobel:
### Summary

Persists `pragmata`-owned metadata for completed eval training runs.

This is needed because `tlmtc` owns the training artifacts but has no concept of `pragmata` eval tasks. By writing a small `pragmata_train.meta.json` sidecar with the `run_id`, `task`, and optional `annotation_export_id`, later predict workflows can select the latest compatible model for a task and validate explicit run IDs against the requested task.

### Key changes

- Add `EvalTrainMeta` to `src/pragmata/core/schemas/eval_output.py`.
- Add `resolve_eval_train_meta_path` to `src/pragmata/core/paths/eval_paths.py`.
- Add `src/pragmata/core/eval/export.py`.
- Persist `pragmata_train.meta.json` after successful `run_tlmtc_train`.
- Keep the public `train_evaluator` return value unchanged.
- Store the sidecar beside the `tlmtc` train run artifacts under `eval/train_outputs/<run_id>/`.

### Tests

- Add unit tests for the eval train metadata schema.
- Add unit coverage for the train metadata path resolver.
- Add unit coverage for exporting the metadata sidecar.
- Update eval API tests to cover metadata persistence orchestration.
- Update eval integration tests to assert the sidecar is written for direct input, explicit export selection, and latest export selection.

### DIFF
--- a/src/pragmata/api/eval.py
+++ b/src/pragmata/api/eval.py
@@ -3,12 +3,14 @@
 from pathlib import Path
 from typing import Any
 
+from pragmata.core.eval.export import export_eval_train_meta
 from pragmata.core.eval.imports import import_eval_train_frame
 from pragmata.core.eval.tlmtc_adapters import run_tlmtc_train
 from pragmata.core.eval.transforms import build_tlmtc_frame
-from pragmata.core.paths.eval_paths import resolve_eval_train_paths
+from pragmata.core.paths.eval_paths import resolve_eval_train_meta_path, resolve_eval_train_paths
 from pragmata.core.paths.paths import WorkspacePaths
 from pragmata.core.schemas.annotation_task import Task
+from pragmata.core.schemas.eval_output import EvalTrainMeta
 from pragmata.core.settings.eval_settings import EvalTrainSettings
 from pragmata.core.settings.settings_base import UNSET, Unset, load_config_file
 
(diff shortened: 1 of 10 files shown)


lramir14 (approved):
Thanks for the nice work. I am only flagging a potential issue with the integration between Pragmata and TLMTC. "atomic_write_json" requires its parent directory to already exist, and every other caller in the repo satisfies that itself via an explicit ensure_dirs() on a pragmata-owned path bundle. export_eval_train_meta is the exception — it writes into train_outputs/<run_id>/, a directory pragmata never creates itself; it only exists because tlmtc happens to create it as a side effect of training. Would it be worth adding a defensive path.parent.mkdir(parents=True, exist_ok=True) before the write, so pragmata satisfies its own utility's contract directly instead of relying on tlmtc's internal behavior?

saschagobel:
> Thanks for the nice work. I am only flagging a potential issue with the integration between Pragmata and TLMTC. "atomic_write_json" requires its parent directory to already exist, and every other caller in the repo satisfies that itself via an explicit ensure_dirs() on a pragmata-owned path bundle. export_eval_train_meta is the exception — it writes into train_outputs/<run_id>/, a directory pragmata never creates itself; it only exists because tlmtc happens to create it as a side effect of training. Would it be worth adding a defensive path.parent.mkdir(parents=True, exist_ok=True) before the write, so pragmata satisfies its own utility's contract directly instead of relying on tlmtc's internal behavior?

Good point about making the `atomic_write_json` precondition explicit. I avoided creating the parent directory here because `train_outputs/<run_id>/` is the `tlmtc`-owned run directory; if it is missing after training returns, `pragmata` should fail clearly rather than create an orphan run directory. I added an explicit parent-dir check before writing the `pragmata` sidecar so the contract is local without taking over `tlmtc` directory ownership.

PR #272, opened 2026-07-13
chore: migrate evaluator training to tlmtc 0.4

saschagobel:
### Summary

Migrates evaluator training to `tlmtc` 0.4 and replaces the removed `full` dependency extra with the training-specific `train` extra.

The default evaluator checkpoints move from EuroBERT to mmBERT because the EuroBERT checkpoints have proven fragile and require `trust_remote_code=True`. The mmBERT alternatives are smaller, fully multilingual, based on a modern BERT architecture, support sequences of up to 8,192 tokens, and provide stronger performance without requiring remote checkpoint code.

The dedicated `trust_remote_code` setting is removed so normal training inherits tlmtc's safer `False` default, while advanced callers can still override it through `train_kwargs`.

### Changes

- Update the eval dependency to `tlmtc[train]>=0.4,<1.0`.
- Refresh the lockfile to resolve `tlmtc` 0.4.0 and its training dependency surface.
- Change the default target checkpoint to `jhu-clsp/mmBERT-base`.
- Change the default proxy checkpoint to `jhu-clsp/mmBERT-small`.
- Remove `trust_remote_code` from `EvalTrainSettings`.
- Remove the dedicated `trust_remote_code` argument from the evaluator API and CLI.
- Stop explicitly forwarding `trust_remote_code` through the tlmtc adapter.
- Allow explicit `trust_remote_code` overrides to pass through `train_kwargs`.
- Update evaluator documentation and tests for the new defaults and API surface.

### Testing

- Updated settings, API, CLI, adapter, and integration coverage.
- Covered forwarding `trust_remote_code` through `train_kwargs`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,7 +34,7 @@ type = [
 ]
 annotation = ["argilla>=2.0,<3.0",]
 querygen = ["sentence-transformers>=3.0,<6.0",]
-eval = ["tlmtc[full]>=0.3,<1.0",]
+eval = ["tlmtc[train]>=0.4,<1.0",]
 cohere = ["langchain-cohere>=0.5,<1.0"]
 deepseek = ["langchain-deepseek>=1.0,<2.0"]
 openai = ["langchain-openai>=1.2,<2.0"]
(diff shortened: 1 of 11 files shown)


lramir14 (approved):
Thanks for the nice work. I see the required changes are applied. Approved.

PR #273, opened 2026-07-14
feat(annotation): add shared uncertainty helpers (Wilson + bootstrap)

henrycgbaker:
## Summary

First PR of eval-scoring stack (design: `docs/design/eval-scoring-metrics.md`). Extracts ci primitives into shared module so eval scoring and IAA share one implementation:

- New `core/annotation/uncertainty.py` w/ `wilson_interval(successes, n, *, ci)` and generic `percentile_bootstrap(n_units, statistic, *, n_resamples, ci, seed)` (cluster/percentile bootstrap over a resampling unit).
- Refactors `iaa.py:bootstrap_alpha` to delegate to `percentile_bootstrap`. The RNG call sequence is preserved, so results are bit-identical — locked by a golden-value regression test captured from the pre-refactor code.
- NB `wilson_interval` has no caller yet; it is the primitive the incoming scoring layer uses for the proportion metrics.

Module lives under `core/annotation` (where the bootstrap logic originated); eval will import from here. Kept dependency-light — NumPy plus the stdlib normal quantile, no SciPy.

## Test plan

- [x] `pytest tests/unit/core/annotation/test_uncertainty.py` — Wilson reference value (8/10 → ~[0.490, 0.943]), clamping at all-0/all-1, bootstrap seed reproducibility, NaN-replicate drop, golden regression lock.
- [x] `pytest tests/unit/core/annotation/test_iaa.py tests/unit/core/annotation/test_iaa_runner.py` — existing IAA suite unaffected by the refactor.
- [x] Drove `pragmata annotation iaa` end-to-end: CI values render, seed-reproducible, widen with `--ci`, shift with `--seed`, `--n-resamples 1` edge handled.
- [x] CI passes.

### DIFF
--- a/src/pragmata/core/annotation/uncertainty.py
+++ b/src/pragmata/core/annotation/uncertainty.py
@@ -1,0 +1,91 @@
+"""Shared uncertainty helpers for annotation and eval reporting.
+
+Confidence-interval primitives reused across reporting stacks: IAA bootstraps
+Krippendorff's alpha (:func:`percentile_bootstrap`), and eval scoring attaches a
+CI to every metric - Wilson intervals for proportions
+(:func:`wilson_interval`), percentile bootstrap for continuous per-query means.
+
+Kept dependency-light: NumPy plus the stdlib normal quantile, no SciPy. Lives
+under ``core.annotation`` because the bootstrap logic originated in ``iaa.py``;
+eval imports from here. Promote to a broader ``core.stats`` only if a wider
+stats surface later emerges.
+"""
+
+from collections.abc import Callable
+from statistics import NormalDist
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+def wilson_interval(successes: int, n: int, *, ci: float = 0.95) -> tuple[float, float]:
+    """Wilson score interval for a binomial proportion.
+
+    Preferred over the normal approximation at small ``n`` and for extreme
+    proportions (all-0 / all-1): it stays within ``[0, 1]`` and remains a
+    positive-width interval instead of collapsing, which is why proportion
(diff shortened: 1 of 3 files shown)


src/pragmata/core/annotation/uncertainty.py
```diff
@@ -0,0 +1,91 @@
...
+        successes: Number of positive outcomes (``0 <= successes <= n``).
+        n: Number of trials (the effective denominator).
+        ci: Confidence level (e.g. 0.95 for a 95% interval).
+
+    Returns:
+        ``(ci_lower, ci_upper)`` clamped to ``[0, 1]``. Returns
+        ``(nan, nan)`` when ``n <= 0``.
+    """
+    if n <= 0:
+        return (float("nan"), float("nan"))
```
ddimmery, line 43: Might as well be defensive on `0 <= successes <= n` as well.

src/pragmata/core/annotation/uncertainty.py
```diff
@@ -0,0 +1,91 @@
...
+"""
+
+from collections.abc import Callable
+from statistics import NormalDist
+
+import numpy as np
+from numpy.typing import NDArray
+
+
+def wilson_interval(successes: int, n: int, *, ci: float = 0.95) -> tuple[float, float]:
```
ddimmery: it's purely aesthetic, but I would prefer the argument to be `alpha` instead of this `ci` parameter. `alpha = 1-ci`, so it's defining the same thing, just I think the terminology will be more standard/clearer.

ddimmery (approved):
Looks good, minor nits.

Tests are nice.

PR #274, opened 2026-07-14
feat(eval): add dedicated eval score input schemas

henrycgbaker:
## Summary

2nd PR of eval-scoring stack (design: `docs/design/eval-scoring-metrics.md`). Scoring has stricter structural requirements than training, so it gets its own input contract instead of reusing the train schemas:

- Adds `RETRIEVAL_SCORE_SCHEMA` / `GROUNDING_SCORE_SCHEMA` / `GENERATION_SCORE_SCHEMA`. Every task requires `record_uuid` (the per-query grouping unit); retrieval additionally requires `chunk_id` and `chunk_rank`.
- Switches `validate_eval_score_frame` onto using these dedicated schemas (prev reused `_TRAIN_SCHEMAS_BY_TASK`) -> training inputs are no longer constrained by scoring-only metadata.
- `K` is inferred from the ranked rows at scoring time, so only `rank >= 1` is enforced = no upper bound, no hard-coded K.
- 
## Test plan

- [x] `pytest tests/unit/core/schemas/test_eval_input.py` - per-task valid frames; `record_uuid` required for all tasks; `chunk_id`/`chunk_rank` required for retrieval only; non-positive rank rejected; rank coercion; and a distinctness test proving one frame is train-valid but score-invalid.
- [x] `pytest tests/unit/core/schemas/ tests/unit/core/eval/` - 301 green; no downstream regressions.
- [x] CI passes.

### DIFF
--- a/src/pragmata/core/schemas/eval_input.py
+++ b/src/pragmata/core/schemas/eval_input.py
@@ -97,8 +97,8 @@ def _identity_column_schemas(
 
     Every task needs ``record_uuid`` to group rows into the per-query unit that
     metrics average over. Retrieval additionally needs ``chunk_id`` (stable
-    per-chunk identity within a query) and ``chunk_rank`` (the 1-based rank the
-    design doc refers to as ``rank``, required by NDCG@K and MRR@K).
+    per-chunk identity within a query) and ``chunk_rank`` (the 1-based rank
+    within the query, required by NDCG@K and MRR@K).
     """
     columns: dict[str, pa.Column] = {
         "record_uuid": pa.Column(str, nullable=False, required=True, checks=_NON_BLANK_STRING),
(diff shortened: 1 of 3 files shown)


src/pragmata/core/schemas/eval_input.py
```diff
@@ -85,6 +90,25 @@ def _label_column_schemas(
...
 
+def _identity_column_schemas(
+    task: Task,
+) -> dict[str, pa.Column]:
+    """Structural grouping/identity columns that scoring requires but training does not.
+
+    Every task needs ``record_uuid`` to group rows into the per-query unit that
+    metrics average over. Retrieval additionally needs ``chunk_id`` (stable
+    per-chunk identity within a query) and ``chunk_rank`` (the 1-based rank the
+    design doc refers to as ``rank``, required by NDCG@K and MRR@K).
```
ddimmery: It's kind of odd to have the docstrings refer to other documentation as something that it is explicitly inconsistent with.

If we're ok making changes to design docs after they've been merged (I am), I would prefer to just change the language everywhere in the design doc to ensure consistency.

ddimmery (approved):

PR #275, opened 2026-07-14
feat(eval): attach per-metric uncertainty to eval score reports

henrycgbaker:
## Summary

Turns score reports' bare rates into uncertainty-carrying metrics (design: `docs/design/eval-scoring-metrics.md`):

- adds nested `MetricScore` model: `point`, `ci_lower`, `ci_upper` (all on the `[0,1]` scale), `method` (`"wilson"` | `"bootstrap"`), and `n` (the effective denominator — queries, or the cited subset for a conditional metric). A model validator enforces `ci_lower <= ci_upper` (equality allowed, for degenerate bootstrap intervals on a constant metric).
- changes every metric field on `RetrievalScoreReport` / `GroundingScoreReport` / `GenerationScoreReport` from bare `Rate` to a `MetricScore` (NB: `conditional_fabrication_rate` stays optional).
- adds run-level `ci_level` to all 3 reports -> jsons are self-describing. The bootstrap resample count is intentionally not stored at the report level: it is meaningful only for `method == "bootstrap"` metrics (none, for the all-Wilson grounding/generation reports), and `MetricScore.method` already records the method per metric.
- `CiLevel` is defined locally (matching the file's existing `Rate` alias) rather than imported from the IAA report schema -> keeps eval output contract decoupled.

## Test plan

- [x] `pytest tests/unit/core/schemas/test_eval_output.py` — `MetricScore` validation (method literal, positive `n`, `[0,1]` bounds, CI bound ordering incl. degenerate + inverted, frozen, extra-forbid); every metric field requires a `MetricScore` not a bare float; `ci_level` open-interval required; nested JSON round-trip; optional conditional stays `None`-able.
- [x] CI passes.

## Doc audit

Updated `docs/design/eval-scoring-metrics.md`: report schemas record only `ci_level`, not `n_bootstrap_resamples` (the IAA reference on line 35 is unchanged — IAA keeps the field, since its CIs are all bootstrap).

### DIFF
--- a/src/pragmata/core/schemas/eval_output.py
+++ b/src/pragmata/core/schemas/eval_output.py
@@ -3,7 +3,7 @@
 from datetime import UTC, datetime
 from typing import Annotated, Literal
 
-from pydantic import BaseModel, ConfigDict, Field, PositiveInt
+from pydantic import BaseModel, ConfigDict, Field, PositiveInt, model_validator
 
 from pragmata.core.schemas.annotation_task import Task
 
@@ -29,6 +29,14 @@ class MetricScore(BaseModel):
     ci_upper: Rate
     method: Literal["wilson", "bootstrap"]
     n: PositiveInt
+
+    @model_validator(mode="after")
+    def _check_ci_ordering(self) -> "MetricScore":
+        # A degenerate interval (lower == upper) is valid: a bootstrap CI on a
+        # constant metric collapses to the point estimate. Only inversion is wrong.
+        if self.ci_lower > self.ci_upper:
+            raise ValueError(f"ci_lower ({self.ci_lower}) must not exceed ci_upper ({self.ci_upper})")
+        return self
 
 
 class EvalTrainMeta(BaseModel):
--- a/src/pragmata/core/schemas/eval_output.py
+++ b/src/pragmata/core/schemas/eval_output.py
@@ -8,6 +8,27 @@ from pydantic import BaseModel, ConfigDict, Field, PositiveInt
 from pragmata.core.schemas.annotation_task import Task
 
 type Rate = Annotated[float, Field(ge=0.0, le=1.0)]
+# Confidence level in the open interval (0, 1); same semantics as the IAA report's CiLevel.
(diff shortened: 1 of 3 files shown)


src/pragmata/core/schemas/eval_output.py
```diff
@@ -8,6 +8,27 @@
...
+    metrics, ``bootstrap`` for continuous ones), and ``n`` is the effective
+    denominator the estimate is over: the number of queries, or the cited
+    subset for a conditional metric.
+    """
+
+    model_config = ConfigDict(frozen=True, extra="forbid")
+
+    point: Rate
+    ci_lower: Rate
+    ci_upper: Rate
```
ddimmery, line 29: It would be nice to validate `ci_upper > ci_lower`

src/pragmata/core/schemas/eval_output.py
```diff
@@ -67,8 +92,10 @@ class GenerationScoreReport(BaseModel):
     notes: str = ""
     created_at: datetime
     n_examples: PositiveInt
-    proper_action_rate: Rate
-    on_topic_rate: Rate
-    helpfulness_rate: Rate
-    incompleteness_rate: Rate
-    unsafe_content_rate: Rate
+    ci_level: CiLevel
+    n_bootstrap_resamples: PositiveInt
```
ddimmery: This is confusing since all metrics below could conceivably use Wilson intervals, correct? It's doubly confusing because `MetricScore` is talking about how data _was_ generated (i.e. elsewhere in code), not prospectively how it _should_ be generated. Does this even need to be here?

ddimmery (approved):

src/pragmata/core/schemas/eval_output.py
henrycgbaker: you're correct. For grounding/generation = all binary, so Wilson. For retrieval, there's also continuous metrics, so bootstrap - but even there resample count is really a property of individual bootstrap CIs, which MetricScore.method already records -> dropping n_bootstrap_resamples from the report schemas as you suggest

PR #277, opened 2026-07-15
docs: define prediction output ingestion for scoring

saschagobel:
### Summary

Clarifies how `tlmtc` prediction outputs should be resolved, imported, and normalized for `pragmata` scoring.

The scoring workflow receives `tlmtc`’s generic `text` and `text_pair` columns in prediction artifacts, and should restore the task-specific `pragmata` column names at the score import boundary.

### Rationale

Duplicating task-specific text columns in prediction artifacts alongside `text` and `text_pair` would increase artifact size and create two equivalent sources of truth.

The mapping is deterministic because scoring knows the requested task. Applying the inverse mapping during scoring-input ingestion keeps `tlmtc` artifacts lean while ensuring they satisfy `pragmata`’s task-specific score schemas.

### Changes

- Require `import_eval_score_frame` as the score ingestion and validation boundary.
- Require prediction-derived scoring inputs to restore task-specific text columns before score validation.
- Reuse `TEXT_COLUMNS_BY_TASK` as the single source of truth for the inverse mapping.
- Assign score input selection and path resolution to `core/paths/eval_paths.py`, mirroring the training workflow.
- Require resolved score paths to retain enough provenance to distinguish prediction artifacts from `pragmata`-shaped inputs.

### DIFF
--- a/docs/design/eval-scoring-metrics.md
+++ b/docs/design/eval-scoring-metrics.md
@@ -169,6 +169,15 @@ No `--top-k` (see above). Input mode follows the same three selectors as the API
 
 Scoring has stricter structural requirements than training, so it gets its own `*_SCORE_SCHEMA` definitions in `eval_input.py` rather than reusing `_TRAIN_SCHEMAS_BY_TASK`; `validate_eval_score_frame` switches to those. This keeps training inputs from being constrained by scoring-only metadata.
 
+Prediction outputs remain tlmtc-shaped: `build_tlmtc_frame(..., mode="predict")` renames the task text columns to `text` and `text_pair`, and tlmtc preserves those generic columns in `predictions.csv`. When scoring a prediction output, the score import path must restore the Pragmata column names before calling `validate_eval_score_frame`. Reuse `TEXT_COLUMNS_BY_TASK[task]` from `eval_input.py` as the single source of truth for the inverse mapping:
+
+```python
+text_column, text_pair_column = TEXT_COLUMNS_BY_TASK[task]
+frame = frame.rename(columns={"text": text_column, "text_pair": text_pair_column})
+```
+
+Identity and provenance columns pass through prediction unchanged. Do not duplicate the task-specific text columns alongside `text` and `text_pair` in prediction inputs or artifacts.
+
 - **retrieval** (`RETRIEVAL_SCORE_SCHEMA`) requires, per row:
   - `record_uuid` - group chunks into queries (the per-query unit).
   - `chunk_rank` (1...K) - required by NDCG@K and MRR@K; without it those metrics cannot be computed.
--- a/docs/design/eval-scoring-metrics.md
+++ b/docs/design/eval-scoring-metrics.md
@@ -169,7 +169,9 @@ No `--top-k` (see above). Input mode follows the same three selectors as the API
 
 Scoring has stricter structural requirements than training, so it gets its own `*_SCORE_SCHEMA` definitions in `eval_input.py` rather than reusing `_TRAIN_SCHEMAS_BY_TASK`; `validate_eval_score_frame` switches to those. This keeps training inputs from being constrained by scoring-only metadata.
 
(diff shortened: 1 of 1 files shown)


docs/design/eval-scoring-metrics.md
```diff
@@ -169,6 +169,19 @@ No `--top-k` (see above). Input mode follows the same three selectors as the API
...
+
+```python
+text_column, text_pair_column = TEXT_COLUMNS_BY_TASK[task]
+frame = frame.rename(columns={"text": text_column, "text_pair": text_pair_column})
+```
+
+Identity and provenance columns pass through prediction unchanged. Do not duplicate the task-specific text columns alongside `text` and `text_pair` in prediction inputs or artifacts.
+
+Input-source selection and path resolution belong in `core/paths/eval_paths.py`, mirroring `resolve_eval_train_paths`. The score path resolver must return one concrete input CSV path and retain enough source provenance to distinguish a tlmtc prediction artifact from already Pragmata-shaped input. The API then passes the resolved input to `import_eval_score_frame`. The exact path-bundle and importer signatures are implementation details for the scoring PR.
```
henrycgbaker, line 183: In my feature branch where I'm working on this (not yet in PR) I was proposing to reduced this provenance tracking complexity; I didn't see what it buys us and adds complexity.

What I was proposing was that `score()` collapses to a single source-agnostic `input_path` (one path with schema validation/enforcement/normalisation, no `export_id`/`prediction_run_id` selectors etc). Instead if we record just the `input_path` in the metadata we can recover whether it was human annotated or synthetically predicted.

henrycgbaker (approved):
Approve, pending decision to the parallel design question around collapsing `score()` to a single source agnostic `input_path()` (a separate PR implementation).

docs/design/eval-scoring-metrics.md
saschagobel: Thanks, this is helpful context. I think there may be two related but distinct questions here:

1. Should scoring operate internally on one resolved input path?
2. Should the public API expose only `input_path`, or also offer IDs that resolve to that path?

This PR proposes the first while retaining `export_id` and the prediction/evaluator ID as convenience selectors. Path resolution collapses the selected source to one concrete CSV before import and scoring.

A few questions and reservations about removing the ID selectors entirely:

- A path-only API works technically, but it requires users to know Pragmata’s workspace layout and locate the correct artifact themselves. For example, they would need to derive and pass paths such as:

  - `annotation/exports/<export_id>/<task>.csv`
  - `eval/prediction_outputs/<evaluator_run_id>/predictions.csv`
Is requiring that workspace knowledge desirable for artifacts that Pragmata created and already knows how to resolve?

- The IDs seem like natural public handles because the producing commands expose them and Pragmata owns the corresponding artifact layouts. Users could pass an `export_id` or `evaluator_run_id`, while the path layer would still resolve that ID immediately to the same single concrete input path used internally. Would retaining this convenience actually add meaningful complexity to the scoring implementation beyond the resolver?

- A path-only score API would also diverge from eval train, which accepts either a direct path or an annotation export ID. Predict similarly uses an evaluator run ID to select a Pragmata-managed model artifact. Is there a reason scoring should require users to navigate internal paths instead of following the same ID-based workflow conventions?

- How would the path-only interface work in a containerized or remote deployment, where the caller’s path may differ from the container-visible path?

- Can provenance reliably be recovered from `input_path`?

- Prediction outputs are tlmtc-shaped while manual and annotation inputs are Pragmata-shaped. If the importer is fully source-agnostic, would it infer the required normalization from column names?

- Because the existing design specifies path and ID selectors, would the single-path proposal update the API, CLI, settings, input-selection section, and metadata contract accordingly?

- Which complexity does removing the public selectors eliminate beyond the relatively small path-resolution layer? The scoring implementation itself can still operate on one resolved path regardless of how users select it.

I may be missing context from the implementation branch, but retaining IDs at the public boundary appears compatible with using one resolved, source-agnostic path internally.

PR #278, opened 2026-07-15
feat(eval): add per-query metrics and query grouping

henrycgbaker:

---

## Summary

4th in eval-scoring stack (design: `docs/design/eval-scoring-metrics.md`). Adds the pure computation layer that turns row-level labels into the per-query values each corpus metric averages over:

- `core/eval/grouping.py` — groups a validated score frame into per-query values for all three tasks. Retrieval is the special case: chunk rows are grouped by `record_uuid` and ordered by `chunk_rank` into per-query ranked lists; grounding/generation are one row per query.
- `core/eval/metrics.py` — the metric primitives (precision@K, sufficiency hit/rate@K, misleading-context rate@K, MRR@K, NDCG@K, and the proportion helpers) plus `corpus_mean`, computed on the per-query values.
- No API/CLI wiring here — both modules are dormant at merge and consumed by the scoring API (next PR in the stack). No dependency on the uncertainty work in #275; this is an independent sibling.

## Test plan

- [x] `pytest tests/unit/core/eval/test_grouping.py tests/unit/core/eval/test_metrics.py` — 33 passed (per-query grouping across tasks, retrieval rank ordering, each metric primitive incl. edge cases, corpus mean).
- [x] CI passes.

### DIFF
--- a/src/pragmata/core/eval/grouping.py
+++ b/src/pragmata/core/eval/grouping.py
@@ -1,0 +1,89 @@
+"""Group validated eval score frames into the per-query values metrics consume.
+
+The scoring resampling unit is the query (``record_uuid``). Retrieval rows are
+grouped into queries and each query's chunks ordered by ``chunk_rank`` before the
+per-query retrieval formulas run; grounding and generation carry one row per
+query, so their per-query values are the row-level labels directly.
+
+Each function returns ``{report_field_name: [per-query value, ...]}`` in a stable
+query order (first appearance), which the scoring layer turns into point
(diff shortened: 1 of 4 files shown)


ddimmery (approved):
This looks fine, but I'm a bit concerned about the lack of validation, e.g. of ranks. I think this is a bigger problem than this diff, though.

henrycgbaker:
> This looks fine, but I'm a bit concerned about the lack of validation, e.g. of ranks. I think this is a bigger problem than this diff, though.

raised as issue #287

PR #279, opened 2026-07-15
feat(eval): add pragmata.eval.score scoring API

henrycgbaker:

---

## Summary

thin `pragmata.eval.score` API orchestration:

- **`api/eval.py: score(...)`** - pure orchestration, no implementation logic: resolve settings → resolve input (`resolve_eval_score_input`) → ingest (`import_eval_score_frame`) → assemble (`build_score_report`) → write `*_scores.json`, wrapped in `error_log`. All maths lives in `core/`.
- **`EvalScoreSettings`** - extends main's stub: renames `labeled_input_path`→`path`, adds `export_id`, `prediction_id`, and the uncertainty params (`n_resamples`, `ci`, `seed`). Selectors are mutually exclusive (enforced in `resolve_eval_score_input`), with a latest-export fallback when none is given.
- All optional API params use the `UNSET` sentinel (matching `train_evaluator`) so config/env/default resolution works.
- **`pragmata.eval.score`** exported alongside `train_evaluator`.

## NB
- `prediction_id` raises `NotImplementedError` (lands with `eval predict`); interim no-selector fallback is the latest annotation export.

### DIFF
--- a/src/pragmata/api/eval.py
+++ b/src/pragmata/api/eval.py
@@ -1,14 +1,23 @@
 """API orchestration for evaluation workflows."""
 
+from datetime import UTC, datetime
 from pathlib import Path
 from typing import Any
 
+from pragmata.api._error_log import error_log
 from pragmata.core.eval.export import export_eval_train_meta
-from pragmata.core.eval.imports import import_eval_predict_frame, import_eval_train_frame
+from pragmata.core.eval.imports import (
+    import_eval_predict_frame,
+    import_eval_score_frame,
+    import_eval_train_frame,
+)
+from pragmata.core.eval.scoring import ScoreReport, build_score_report
 from pragmata.core.eval.tlmtc_adapters import run_tlmtc_predict, run_tlmtc_train
 from pragmata.core.eval.transforms import build_tlmtc_frame
 from pragmata.core.paths.eval_paths import (
     resolve_eval_predict_paths,
+    resolve_eval_score_input,
+    resolve_eval_score_paths,
     resolve_eval_train_meta_path,
     resolve_eval_train_paths,
     resolve_eval_train_run_id,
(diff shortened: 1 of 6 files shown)


src/pragmata/api/eval.py
```diff
@@ -194,3 +207,98 @@ def predict_labels(
...
+            ``"generation"``).
+        n_resamples: Bootstrap iterations for the continuous metrics. Defaults to 1000.
+        ci: Confidence level for every interval. Defaults to 0.95.
+        seed: Optional RNG seed for reproducible bootstrap intervals.
+        config_path: Path to a YAML configuration file.
+
+    Returns:
+        The task-specific score report, also written to ``*_scores.json``.
+
+    Raises:
```
ddimmery, line 254: Also raises `ValueError` from `_require_at_most_one_selector` according to Claude.

src/pragmata/api/eval.py
```diff
@@ -194,3 +207,98 @@ def predict_labels(
...
+        report = build_score_report(
+            frame,
+            task=settings.task,
+            ci=settings.ci,
+            n_resamples=settings.n_resamples,
+            seed=settings.seed,
+            source=resolved.source,
+            created_at=datetime.now(UTC),
+        )
+        match settings.task:
```
ddimmery, line 297: could use a default case to provide informative errors if a Task were added.

src/pragmata/api/eval.py
```diff
@@ -194,3 +207,98 @@ def predict_labels(
...
+            "task": task,
+            "n_resamples": n_resamples,
+            "ci": ci,
+            "seed": seed,
+        },
+    )
+    workspace = WorkspacePaths.from_base_dir(settings.base_dir)
+    score_paths = resolve_eval_score_paths(workspace=workspace, score_id=settings.score_id).ensure_dirs()
+
+    with error_log(score_paths.tool_root):
```
ddimmery: This resolves to `eval/` rather than under anything score specific. This is also how annotation modules work, but not how querygen works. This is an inconsistency worth fixing at some point. Probably easier to just fix rather than filing an issue.

ddimmery (approved):
The design doc should be updated to align with this diff:
- `labeled_input_path` and `prediction_run_id` should be updated
- Precedence open question was resolved, should be updated

PR #280, opened 2026-07-15
feat(eval): add pragmata eval score CLI subcommand

henrycgbaker:

---

## Summary

`pragmata eval score` CLI subcommand - the top of the eval-score stack, a thin Typer wrapper 1:1 over `pragmata.eval.score`.

- `eval score --task ... [--path P | --export-id ID | --prediction-id ID] [--score-id] [--base-dir] [--n-resamples] [--ci] [--seed] [--config]` maps each flag to the API (UNSET when omitted), then renders each metric as `point [lo, hi] (method, n)` (`n/a` for an absent conditional metric).
- Adds `ScoreReport.metric_scores()` on the three report schemas (display-ordered `(name, metric)` pairs) - the CLI's render helper, folded here (its only consumer) rather than restacking the schema slice.

## Finalised naming (this stack)

Selector and provenance-kind are consistent per concept:

| selector / CLI flag | provenance `kind` |
|---|---|
| `--path` | `direct_path` |
| `--export-id` | `annotation_export` |
| `--prediction-id` | `model_prediction` |

### DIFF
--- a/src/pragmata/cli/commands/eval.py
+++ b/src/pragmata/cli/commands/eval.py
@@ -163,7 +163,7 @@ def score_command(
     typer.echo(f"\n{report.task.value} scores (n={report.n_examples}, {report.ci_level:.0%} CI):")
     for name, metric in report.metric_scores():
         if metric is None:
-            typer.echo(f"  {name}: n/a (no cited examples)")
+            typer.echo(f"  {name}: n/a (not computed)")
         else:
             typer.echo(
                 f"  {name}: {metric.point:.3f} "
--- a/src/pragmata/cli/commands/eval.py
+++ b/src/pragmata/cli/commands/eval.py
@@ -6,7 +6,7 @@ from pragmata import eval
 from pragmata.api import UNSET
 from pragmata.cli.parsing import parse_cli_value
 
-eval_app = typer.Typer(help="Evaluation commands.")
+eval_app = typer.Typer(help="Evaluation commands.", no_args_is_help=True)
 
 
 @eval_app.command("train-evaluator")
--- a/src/pragmata/cli/commands/eval.py
+++ b/src/pragmata/cli/commands/eval.py
@@ -86,3 +86,86 @@ def train_evaluator_command(
     typer.echo(f"run_id: {result.paths.run_id}")
     typer.echo(f"run_directory: {result.paths.run_dir}")
     typer.echo(f"model_directory: {result.paths.model_dir}")
+
+
+@eval_app.command("score")
+def score_command(
+    task: str | None = typer.Option(
+        None,
+        "--task",
+        help="Eval task to score: retrieval, grounding, or generation.",
+    ),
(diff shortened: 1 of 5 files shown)


src/pragmata/core/schemas/eval_output.py
```diff
@@ -89,6 +89,17 @@ class RetrievalScoreReport(BaseModel):
...
 
+    def metric_scores(self) -> list[tuple[str, MetricScore | None]]:
+        """The report's metrics in display order, each paired with its field name."""
+        return [
+            ("topical_precision_at_k", self.topical_precision_at_k),
+            ("sufficiency_hit_at_k", self.sufficiency_hit_at_k),
+            ("sufficiency_rate_at_k", self.sufficiency_rate_at_k),
+            ("misleading_context_rate_at_k", self.misleading_context_rate_at_k),
+            ("mean_reciprocal_rank_at_k", self.mean_reciprocal_rank_at_k),
+            ("ndcg_at_k", self.ndcg_at_k),
```
ddimmery, line 100: These are already defines as fields in the `RetrievalScoreReport` class, so this is kind of a redefinition as you could infer these based on the `RetrievalScoreReport.model_fields` and check for `MetricScore` typed fields. This is also not particularly simple, so I think I'd leave this as-is unless you see a cleaner approach.

tests/unit/cli/test_cli_eval.py
```diff
@@ -79,7 +82,7 @@ def fake_train_evaluator(**kwargs: Any) -> _TrainResult:
 
         monkeypatch.setattr("pragmata.eval.train_evaluator", fake_train_evaluator)
 
-        result = runner.invoke(eval_app, [])
+        result = runner.invoke(eval_app, ["train-evaluator"])
```
ddimmery, line 98: The bare invocation of `pragmata eval` has changed. I don't know of any documentation on this, but if it exists it should be updated to note.

src/pragmata/cli/commands/eval.py
```diff
@@ -86,3 +86,86 @@ def train_evaluator_command(
...
+        n_resamples=UNSET if n_resamples is None else n_resamples,
+        ci=UNSET if ci is None else ci,
+        seed=UNSET if seed is None else seed,
+        config_path=UNSET if config_path is None else config_path,
+    )
+
+    typer.echo(f"\n{report.task.value} scores (n={report.n_examples}, {report.ci_level:.0%} CI):")
+    for name, metric in report.metric_scores():
+        if metric is None:
+            typer.echo(f"  {name}: n/a (no cited examples)")
```
ddimmery: This is correct because the only optional metric is `conditional_fabrication_rate`, but that assumption probably shouldn't be made here. I'd make the message more generic and less informative to resolve this generally. Something like `n/a (not computed)` or `n/a (undefined for this corpus)`

ddimmery (approved):

PR #282, opened 2026-07-17
feat(eval): add core score report assembly with ScoreInputSource

henrycgbaker:
## Summary

Adds a non-IO scoring layer that assembles task-specific score reports.

- **`core/eval/scoring.py`** - `build_score_report(...)`: point estimates via `metrics.corpus_mean` + per-query formulas; CIs via `core.annotation.uncertainty` (Wilson for proportion metrics, percentile bootstrap for continuous retrieval metrics). IO-free.
- **`ScoreInputSource`** provenance model (`core/schemas/eval_output.py`): `kind` (`direct_path` | `annotation_export` | `model_prediction`), `ref` (the selector's value), `resolved_path` (concrete CSV read, workspace-relative). `kind` also drives ingestion - only `model_prediction` inputs need tlmtc text-column restoration.

## Contract note

Score reports carry a required **`source: ScoreInputSource`** instead of the merged-but-provisional `annotation_export_id: str | None`. Provenance records *how* the input was selected + the resolved path (i.e. distinguishes prediction vs pragmata-shaped input). `EvalTrainMeta` keeps `annotation_export_id`.

## Doc audit

No doc update needed - internal scoring layer + schema contract; no user-visible CLI/flag/string surface changed.

### DIFF
--- a/src/pragmata/core/eval/scoring.py
+++ b/src/pragmata/core/eval/scoring.py
@@ -1,0 +1,123 @@
+"""Assemble task-specific eval score reports from labeled per-query values.
+
+This is the scoring layer referenced by ``core/eval/metrics.py``: point estimates
+come from ``metrics.corpus_mean`` and the per-query formulas, while each metric's
+confidence interval comes from the shared ``core.annotation.uncertainty`` helpers
+(Wilson for proportion metrics, percentile bootstrap for the continuous retrieval
+metrics). It is the report-assembly step the API orchestrates - pure and I/O-free:
+the caller resolves and reads the frame and writes the report.
+"""
+
+from datetime import datetime
+
+import numpy as np
+import pandas as pd
+from numpy.typing import NDArray
+
+from pragmata.core.annotation.uncertainty import percentile_bootstrap, wilson_interval
+from pragmata.core.eval import grouping, metrics
+from pragmata.core.schemas.annotation_task import Task
+from pragmata.core.schemas.eval_output import (
+    GenerationScoreReport,
+    GroundingScoreReport,
+    MetricScore,
+    RetrievalScoreReport,
+    ScoreInputSource,
+)
+
+ScoreReport = RetrievalScoreReport | GroundingScoreReport | GenerationScoreReport
+
+
+def _wilson_metric(values: list[float] | NDArray, *, alpha: float) -> MetricScore:
+    """MetricScore for a proportion metric: point = proportion, CI = Wilson."""
+    array = np.asarray(values)
+    n = len(array)
+    lower, upper = wilson_interval(int(array.sum()), n, alpha=alpha)
+    return MetricScore(point=metrics.corpus_mean(values), ci_lower=lower, ci_upper=upper, method="wilson", n=n)
+
+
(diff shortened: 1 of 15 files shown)


src/pragmata/core/eval/scoring.py
```diff
@@ -0,0 +1,123 @@
...
+            unsupported_claim_rate=wilson(values["unsupported_claim_rate"]),
+            contradiction_rate=wilson(values["contradiction_rate"]),
+            citation_presence_rate=wilson(values["citation_presence_rate"]),
+            conditional_fabrication_rate=_conditional_metric(
+                grouping.conditional_fabrication_units(frame), alpha=alpha
+            ),
+        )
+
+    values = grouping.generation_per_query_values(frame)
+    return GenerationScoreReport(
```
ddimmery: It's easy to have the fall-through just raise an error for an unknown/unsupported task type rather than assume Generation.

src/pragmata/core/eval/scoring.py
```diff
@@ -0,0 +1,123 @@
...
+    def wilson(values: list[float]) -> MetricScore:
+        return _wilson_metric(values, alpha=alpha)
+
+    if task == Task.RETRIEVAL:
+        values = grouping.retrieval_per_query_values(frame)
+        return RetrievalScoreReport(
+            source=source,
+            created_at=created_at,
+            n_examples=len(values["topical_precision_at_k"]),
+            top_k=int(frame["chunk_rank"].max()),
```
ddimmery, line 91: I know we've done a good bit of discussion on this, but I'd like to see an issue written out describing limitations of our inference on `k` throughout the pragmata (even without a general solution in place). Also, is `chunk_rank` indexed from 0 or from 1? This could be an off-by-one error depending on the answer.

src/pragmata/core/eval/scoring.py
```diff
@@ -0,0 +1,123 @@
...
+    n_resamples: int,
+    seed: int | None,
+    source: ScoreInputSource,
+    created_at: datetime,
+) -> ScoreReport:
+    """Assemble the task-specific report from per-query values (pure, I/O-free)."""
+    alpha = 1.0 - ci
+
+    def bootstrap(values: list[float]) -> MetricScore:
+        return _bootstrap_metric(values, alpha=alpha, n_resamples=n_resamples, seed=seed)
```
ddimmery, line 78: The seed is the same for every metric. I don't think this will cause any major issues: each marginal bootstrap will be marginally valid (with large enough n_resamples), and we never do anything with the joint. I think it's worth adding a defensive comment here but moving on (separate seeds aren't obviously better, but if anything changes in the future, this is behavior to have flagged).

ddimmery (approved):
minor notes

src/pragmata/core/eval/scoring.py
henrycgbaker: `chunk_rank` is `(Field(ge=1)` /  Argilla min=1) -> top_k = `frame["chunk_rank"].max()` correctly equals numb of retrieved chunks

PR #283, opened 2026-07-17
feat(eval): add score-input resolution to eval_paths

henrycgbaker:

---

## Summary

Adds score-input selection + path resolution to `core/paths/eval_paths.py`, and makes eval input selection mutually exclusive across both train and score.

- **`resolve_eval_score_input(...)`** → `EvalScoreInput` bundle: the concrete labeled CSV to read plus its `ScoreInputSource` provenance (`kind` / `ref` / `resolved_path`), so the caller and report record how the input was selected without re-resolving.
- **Mutually exclusive selectors.** `path`, `export_id`, and `prediction_id` are alternatives: at most one may be given, more than one raises, and no selector resolves to the latest annotation export for the task. Enforced by a shared `_require_at_most_one_selector` helper.
- **`resolve_eval_train_paths`** uses the same helper, so `labeled_data_path` vs `export_id` follows identical rules.
- **`EvalScoreInput`** dataclass (input CSV + provenance) and **`provenance_path(...)`** (records the input relative to the workspace, or absolute when outside it).

### DIFF
--- a/src/pragmata/core/paths/eval_paths.py
+++ b/src/pragmata/core/paths/eval_paths.py
@@ -109,6 +109,21 @@ def find_latest_annotation_export_id(
     return latest_export_id
 
 
+def _require_at_most_one_selector(**selectors: object | None) -> None:
+    """Reject ambiguous input selection: at most one selector may be set.
+
+    Eval input selectors (a direct path, an export id, a prediction id) are
+    mutually exclusive - there is no precedence between them. Passing more than
+    one raises; passing none is allowed (callers fall back to the latest export).
+    """
+    given = [name for name, value in selectors.items() if value is not None]
+    if len(given) > 1:
+        raise ValueError(
+            f"At most one input selector may be given, got {', '.join(given)}. "
+            "Pass a single selector, or none to use the latest annotation export."
+        )
+
+
 def resolve_eval_train_paths(
     *,
     workspace: WorkspacePaths,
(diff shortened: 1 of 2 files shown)


src/pragmata/core/paths/eval_paths.py
```diff
@@ -358,5 +373,122 @@ def resolve_eval_score_paths(
...
+        case Task.GROUNDING:
+            input_csv = export_paths.grounding_annotation_csv
+        case Task.GENERATION:
+            input_csv = export_paths.generation_annotation_csv
+
+    if not input_csv.is_file():
+        raise FileNotFoundError(
+            f"Annotation export {resolved_export_id!r} does not contain the requested {task.value} CSV. "
+            f"Expected {input_csv}."
+        )
```
ddimmery, line 476: It seems like this same pattern is used in `find_latest_annotation_export_id`, `resolve_eval_train_paths` and `resolve_eval_score_input`. It would be better to pull out a shared helper to do this.

src/pragmata/core/paths/eval_paths.py
```diff
@@ -358,5 +373,122 @@ def resolve_eval_score_paths(
...
+            f"Expected annotation_export.meta.json at {export_paths.export_meta_json}."
+        )
+
+    match task:
+        case Task.RETRIEVAL:
+            input_csv = export_paths.retrieval_annotation_csv
+        case Task.GROUNDING:
+            input_csv = export_paths.grounding_annotation_csv
+        case Task.GENERATION:
+            input_csv = export_paths.generation_annotation_csv
```
ddimmery: A catch-all case to explicitly throw an error here rather than falling through and failing at the next step would be good.

src/pragmata/core/paths/eval_paths.py
```diff
@@ -358,5 +373,122 @@ def resolve_eval_score_paths(
...
+    """Resolve the labeled CSV consumed by eval score and its provenance.
+
+    At most one input selector may be given: a direct ``path``, an ``export_id``,
+    or a ``prediction_id``. With no selector, the latest valid annotation export
+    for the task is used, mirroring ``resolve_eval_train_paths``.
+
+    Direct paths and annotation exports are already Pragmata-shaped and score
+    without further preparation. Scoring a prediction run is not yet supported:
+    ``pragmata eval predict`` has landed, but it has no run-addressable output
+    layout and the prediction-run path resolver (``find_latest_prediction_run``)
```
ddimmery: I assume this is later in the stack?

ddimmery (approved):

src/pragmata/core/paths/eval_paths.py
henrycgbaker: Not in this stack, tracked in #303

PR #284, opened 2026-07-17
feat(eval): add import_eval_score_frame with majority consolidation

henrycgbaker:

---

## Summary

Adds `import_eval_score_frame` as the eval score ingestion boundary, and consolidates multi-annotator inputs into one row per scoring unit.

- **`import_eval_score_frame`** (`core/eval/imports.py`): `read -> (restore tlmtc text columns for prediction inputs) → validate → consolidate → guard`.
- **Majority consolidation** reuses the train-side row consolidation, generalized into a shared **`consolidate_labels_by_majority`** (`core/eval/transforms.py`): per-label strict majority; exact 50/50 tie falls back to a representative row; a no-op when units are already unique. `build_tlmtc_frame` (train) uses the same helper.
- **`_guard_unique_scoring_units`** remains as the post-collapse invariant - it hard-errors only on a residual duplicate the collapse cannot resolve (e.g. two `chunk_id`s sharing a `chunk_rank`).

### DIFF
--- a/src/pragmata/core/eval/imports.py
+++ b/src/pragmata/core/eval/imports.py
@@ -5,7 +5,14 @@ from pathlib import Path
 import pandas as pd
 
 from pragmata.core.schemas.annotation_task import Task
-from pragmata.core.schemas.eval_input import validate_eval_predict_frame, validate_eval_train_frame
+from pragmata.core.schemas.eval_input import (
+    TEXT_COLUMNS_BY_TASK,
+    EvalInputSchemaError,
+    validate_eval_predict_frame,
+    validate_eval_score_frame,
+    validate_eval_train_frame,
+)
+from pragmata.core.schemas.eval_output import ScoreInputSource
 
 
 def import_eval_train_frame(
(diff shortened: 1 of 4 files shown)


tests/unit/core/eval/test_imports.py
```diff
@@ -158,3 +163,119 @@ def test_import_eval_predict_frame_rejects_invalid_csv(
...
+                "q1,c1,1,1,0,r1,ch1,1",
+            ],
+        )
+
+        frame = import_eval_score_frame(path=path, task=Task.RETRIEVAL, source=self._PREDICTION)
+
+        assert {"query", "chunk"}.issubset(frame.columns)
+        assert "text" not in frame.columns and "text_pair" not in frame.columns
+        assert frame.loc[0, "query"] == "q1"
+        assert frame.loc[0, "chunk"] == "c1"
```
ddimmery, line 281: I assume later in the stack we'll get a round trip test?

src/pragmata/core/eval/imports.py
```diff
@@ -42,3 +50,73 @@ def import_eval_predict_frame(
...
+    Raises:
+        EvalInputSchemaError: If the frame violates the score contract or retains a
+            duplicate scoring unit after majority consolidation.
+    """
+    frame = pd.read_csv(path, encoding="utf-8")
+    if source.kind == "model_prediction":
+        frame = _restore_pragmata_text_columns(frame, task=task)
+    validated = validate_eval_score_frame(frame, task=task)
+    consolidated = consolidate_labels_by_majority(validated, task=task)
+    _guard_unique_scoring_units(consolidated, task=task)
```
ddimmery, line 97: this should only be necessary for retrieval, but that's not a problem.

src/pragmata/core/eval/transforms.py
```diff
@@ -70,17 +70,20 @@ def build_tlmtc_frame(
...
-    Tied labels fall back to the selected source row. When an observed row matches
-    all strict-majority labels, that row is selected to preserve non-label metadata
-    deterministically.
+    """Collapse repeated annotation units to one row via per-label majority consensus.
+
+    Shared by eval train and score ingestion: multiple annotator rows for the same
+    scoring unit (retrieval ``(record_uuid, chunk_id)``; grounding/generation
+    ``record_uuid``) are reduced to a single row. Each label with a strict majority
+    (> half positive) is set independently; a tied label (exact 50/50 on an even
+    number of annotators) falls back to the selected source row's value. When an
```
ddimmery, line 88: Is there a test on the tied case? Would be useful to ensure consistent behavior.

ddimmery (approved):

tests/unit/core/eval/test_imports.py
henrycgbaker: adding to do this in issue #303 that covers landing prediction-run scoring

src/pragmata/core/eval/transforms.py
henrycgbaker: test_build_tlmtc_frame_consolidates_training_duplicates_by_earliest_fallback_on_tie (test_transforms.py:264)

PR #285, opened 2026-07-17
feat: add public label prediction workflow

saschagobel:
### Summary

Adds the public `pragmata.eval.predict_labels()` Python workflow for predicting task-specific evaluation labels from unlabeled CSV inputs.

`pragmata` resolves and validates the prediction input, selects a task-compatible evaluator, maps `pragmata` task columns to `tlmtc`'s input contract, and delegates inference to the existing `tlmtc` prediction adapter. The raw `tlmtc` `PredictResult` is returned unchanged.

Only the unlabeled data path and evaluation task are required. If no evaluator run ID is supplied through the call or configuration, `pragmata` selects the latest evaluator compatible with the requested task.

### Changes

- Add the public `predict_labels()` API to `pragmata.api.eval`.
- Resolve prediction settings from an optional YAML configuration and explicit API overrides.
- Preserve `UNSET` semantics so omitted API arguments can fall back to configuration values.
- Resolve the eval workspace and explicit unlabeled prediction input path.
- Select a concrete evaluator run ID through `pragmata`'s task-aware evaluator selector.
- Import and validate the task-specific unlabeled CSV.
- Transform `pragmata` text columns to `tlmtc`'s `text` and `text_pair` contract.
- Pass the transformed in-memory DataFrame, eval work directory, concrete evaluator ID, and `tlmtc`-owned prediction options to the prediction adapter.
- Return `tlmtc`'s `PredictResult` without rewriting its paths or generated artifacts.
- Re-export `predict_labels` through the public `pragmata.eval` namespace.

### Testing

- Covered direct prediction from an explicit unlabeled CSV and evaluator run ID.
- Covered task propagation through evaluator selection, input import, and dataframe transformation.
- Covered forwarding of the transformed DataFrame, concrete evaluator ID, eval work directory, and prediction options.
- Covered automatic task-compatible evaluator selection when no evaluator ID is configured.
- Covered configuration merging and explicit API override precedence.
- Covered returning the raw adapter result unchanged.

### DIFF
--- a/src/pragmata/api/eval.py
+++ b/src/pragmata/api/eval.py
@@ -4,14 +4,19 @@ from pathlib import Path
 from typing import Any
 
 from pragmata.core.eval.export import export_eval_train_meta
-from pragmata.core.eval.imports import import_eval_train_frame
-from pragmata.core.eval.tlmtc_adapters import run_tlmtc_train
+from pragmata.core.eval.imports import import_eval_predict_frame, import_eval_train_frame
+from pragmata.core.eval.tlmtc_adapters import run_tlmtc_predict, run_tlmtc_train
 from pragmata.core.eval.transforms import build_tlmtc_frame
-from pragmata.core.paths.eval_paths import resolve_eval_train_meta_path, resolve_eval_train_paths
+from pragmata.core.paths.eval_paths import (
+    resolve_eval_predict_paths,
+    resolve_eval_train_meta_path,
+    resolve_eval_train_paths,
+    resolve_eval_train_run_id,
+)
 from pragmata.core.paths.paths import WorkspacePaths
 from pragmata.core.schemas.annotation_task import Task
 from pragmata.core.schemas.eval_output import EvalTrainMeta
-from pragmata.core.settings.eval_settings import EvalTrainSettings
+from pragmata.core.settings.eval_settings import EvalPredictSettings, EvalTrainSettings
 from pragmata.core.settings.settings_base import UNSET, Unset, load_config_file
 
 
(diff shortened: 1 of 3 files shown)


lramir14 (approved):
Nice work. I only have one minor comment: the `predict_labels` docstring states outputs land under `<base_dir>/eval/prediction_outputs/<evaluator_run_id>/`, but nothing in pragmata builds that path. The path`resolve_eval_predict_paths` only resolves the flat `eval/` root, and `run_tlmtc_predict` forwards `work_dir` straight to tlmtc without touching the filesystem. So that nested layout is tlmtc's own output convention, not something pragmata owns or guarantees. In this way, the docstring would benefit from some rewording to attribute it explicitly to tlmtc's prediction contract (and ideally verifying it against tlmtc's docs). Otherwise the docstring can silently go stale if tlmtc changes its layout, and nothing in the suite would catch it since the tests mock the adapter and don't assert on `result.paths`.

Approved.

saschagobel:
> Nice work. I only have one minor comment: the `predict_labels` docstring states outputs land under `<base_dir>/eval/prediction_outputs/<evaluator_run_id>/`, but nothing in pragmata builds that path. The path`resolve_eval_predict_paths` only resolves the flat `eval/` root, and `run_tlmtc_predict` forwards `work_dir` straight to tlmtc without touching the filesystem. So that nested layout is tlmtc's own output convention, not something pragmata owns or guarantees. In this way, the docstring would benefit from some rewording to attribute it explicitly to tlmtc's prediction contract (and ideally verifying it against tlmtc's docs). Otherwise the docstring can silently go stale if tlmtc changes its layout, and nothing in the suite would catch it since the tests mock the adapter and don't assert on `result.paths`.
> 
> Approved.

Good point, thanks. I removed the concrete output-directory layout from predict_labels, since that convention is owned by `tlmtc` and only creates avoidable maintenance surface here. I made the same adjustment to train_evaluator for consistency.

PR #286, opened 2026-07-17
test: cover predict_labels end to end

saschagobel:
### Summary

Adds end-to-end integration coverage for the public `pragmata.eval.predict_labels()` workflow.

The tests exercise the real `tlmtc.predict_tlmtc()` dependency using a small evaluator trained from the existing tiny BERT checkpoint. The evaluator is trained once and reused across prediction cases to keep the integration suite proportionate.

### Changes

- Add a dedicated `TestPredictLabels` integration-test class.
- Add task-specific unlabeled retrieval input with identity and provenance columns.
- Train one reusable prediction-capable evaluator through the public eval training API.
- Run prediction through the public `pragmata.eval.predict_labels()` API and real `tlmtc` inference.
- Cover explicit evaluator selection.
- Cover automatic selection of the latest task-compatible evaluator while ignoring a globally newer evaluator for another task.
- Cover explicit task mismatch rejection before TLMTC inference.
- Validate the returned TLMTC `PredictResult` paths.
- Validate generated `probabilities.csv` and `predictions.csv` artifacts.
- Verify that identity and provenance columns are retained while task text columns are mapped to `text` and `text_pair`.
- Verify binary predictions and bounded probability values.

### DIFF
--- a/tests/integration/test_eval.py
+++ b/tests/integration/test_eval.py
@@ -1,4 +1,4 @@
-"""Integration tests for the public evaluator training surface."""
+"""Integration tests for public evaluator workflows."""
 
 import json
 from datetime import UTC, datetime
@@ -10,9 +10,11 @@ import pytest
 
 from pragmata import eval
 from pragmata.core.paths.annotation_paths import resolve_export_paths
+from pragmata.core.paths.eval_paths import resolve_eval_train_meta_path
 from pragmata.core.paths.paths import WorkspacePaths
 from pragmata.core.schemas.annotation_export import AnnotationExportMeta
 from pragmata.core.schemas.annotation_task import Task
+from pragmata.core.schemas.eval_output import EvalTrainMeta
 
 pytestmark = [pytest.mark.integration, pytest.mark.eval]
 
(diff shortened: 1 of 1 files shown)


lramir14 (approved):
Looks great. 
One non-blocking note: the decoy sidecars in `test_selects_latest_task_compatible_evaluator` persist in the shared `base_dir`. Safe today, but a future test in that workspace that triggers automatic selection could resolve to a decoy with no backing model. Might be worth isolating that test on a function-scoped workspace (as the mismatch test already does) or noting the caveat in a comment.  

Approved.

saschagobel:
> Looks great. One non-blocking note: the decoy sidecars in `test_selects_latest_task_compatible_evaluator` persist in the shared `base_dir`. Safe today, but a future test in that workspace that triggers automatic selection could resolve to a decoy with no backing model. Might be worth isolating that test on a function-scoped workspace (as the mismatch test already does) or noting the caveat in a comment.
> 
> Approved.

I now remove the metadata-only evaluator candidates in a finally block so they cannot affect later tests sharing the module-scoped workspace.

PR #295, opened 2026-07-18
feat: add predict-labels CLI command

saschagobel:
### Summary

Adds the `pragmata eval predict-labels` command for predicting task-specific evaluation labels with a trained evaluator.

The command is a thin wrapper around the public `pragmata.eval.predict_labels()` API. It preserves configuration fallback for omitted options, parses `tlmtc`-owned prediction arguments from JSON, and prints the artifact paths returned by `tlmtc`.

### Changes

- Register `predict-labels` on the existing eval Typer application.
- Expose unlabeled input path, evaluator run ID, task, workspace, configuration, and prediction keyword options.
- Map omitted CLI options to Pragmata's `UNSET` sentinel.
- Parse `--predict-kwargs` through the existing CLI value parser.
- Delegate exclusively to `pragmata.eval.predict_labels()`.
- Print the selected evaluator ID, prediction directory, probabilities path, and predictions path.

### Testing

- Covered eval help output listing both `train-evaluator` and `predict-labels`.
- Covered the complete `predict-labels` help surface.
- Covered omitted options mapping to `UNSET`.
- Covered explicit option forwarding and JSON prediction-kwargs parsing.
- Covered completion output using the returned prediction artifact paths.

### DIFF
--- a/src/pragmata/cli/commands/eval.py
+++ b/src/pragmata/cli/commands/eval.py
@@ -86,3 +86,53 @@ def train_evaluator_command(
     typer.echo(f"run_id: {result.paths.run_id}")
     typer.echo(f"run_directory: {result.paths.run_dir}")
     typer.echo(f"model_directory: {result.paths.model_dir}")
+
+
+@eval_app.command("predict-labels")
+def predict_command(
+    unlabeled_data_path: str | None = typer.Option(
+        None,
+        "--unlabeled-data-path",
+        help="Path to the task-specific unlabeled eval prediction CSV.",
+    ),
+    evaluator_run_id: str | None = typer.Option(
+        None,
+        "--evaluator-run-id",
+        help="Evaluator training run identifier. If omitted, the latest task-compatible evaluator is used.",
+    ),
+    task: str | None = typer.Option(
+        None,
+        "--task",
+        help="Eval task to predict for: retrieval, grounding, or generation.",
+    ),
+    base_dir: str | None = typer.Option(
+        None,
+        "--base-dir",
+        help="Workspace base directory. Defaults to current working directory.",
+    ),
+    config_path: str | None = typer.Option(
+        None,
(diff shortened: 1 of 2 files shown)


lramir14 (approved):
Thanks for the nice work. Approving with minor comments.

1. test_bare_invocation_shows_help on main asserts the available commands (train-evaluator, score), so perhaps it would be worth updating it to include predict-labels, or it'll be incomplete after merge.
2. TestTrainEvaluatorCommand covers invocation through the root app, but TestPredictLabelsCommand only tests the sub-app directly. Perhaps worth covering symmetrically.
3. --task is a free str with valid values only in the help text, so --task retreival passes straight through to the API. Does predict_labels() error clearly, or would a Typer Enum be worth adding? Same applies to train-evaluator and score, so worth sharing any validation rather than duplicating.

saschagobel:
> Thanks for the nice work. Approving with minor comments.
> 
> 1. test_bare_invocation_shows_help on main asserts the available commands (train-evaluator, score), so perhaps it would be worth updating it to include predict-labels, or it'll be incomplete after merge.
> 2. TestTrainEvaluatorCommand covers invocation through the root app, but TestPredictLabelsCommand only tests the sub-app directly. Perhaps worth covering symmetrically.
> 3. --task is a free str with valid values only in the help text, so --task retreival passes straight through to the API. Does predict_labels() error clearly, or would a Typer Enum be worth adding? Same applies to train-evaluator and score, so worth sharing any validation rather than duplicating.

Thanks! I updated the merged command-list assertion to include predict-labels and added symmetric root-app help coverage for the prediction command. I kept --task as a string at the CLI boundary because train, predict, and score already share centralized validation through the Pydantic settings’ Task enum; invalid values fail with the allowed values listed explicitly. This avoids duplicating the same validation across the three CLI commands.

PR #296, opened 2026-07-18
ci: add package build validation

saschagobel:
### Summary

Adds package build validation to the regular CI workflow so packaging failures are detected before release tags are created.

### Changes

- Add an independent CI job that builds the wheel and source distribution with `uv build`.
- Verify that the lockfile is current before building.
- Update all CI jobs to use `astral-sh/setup-uv@v8.2.0`.


lramir14 (approved):
thanks for the nice work. I have no issues. Approved.

PR #297, opened 2026-07-18
ci: prepare trusted PyPI release workflow

saschagobel:
### Summary

Replaces the build-only release placeholder with a gated workflow for running release preflight checks, building package distributions, and publishing tagged releases to PyPI through Trusted Publishing.

Integration tests run before the build, while publishing remains restricted to version-tag pushes. Manual dispatches can run either the integration suites alone or the full preflight including the package build.

### Changes

- Add manual release preflight dispatch with an optional package build.
- Run annotation, query-generation, and evaluation integration suites as parallel matrix jobs.
- Install each suite's dedicated optional dependency extra.
- Start and clean up the Docker-based Argilla stack only for annotation tests.
- Require release tags to reference commits contained in `main`.
- Require all integration suites to pass before building distributions.
- Verify that the release tag matches the package version.
- Check that the lockfile is current before testing and building.
- Build source and wheel distributions with `uv build`.
- Validate distribution metadata with Twine.
- Transfer distributions between build and publish jobs using workflow artifacts.
- Publish tagged releases through the `pypi` GitHub environment and PyPI Trusted Publishing.
- Limit OIDC token permissions to the publishing job.
- Prevent manually dispatched preflight runs from publishing packages.

### Testing

- Verified the workflow YAML syntax.
- Verified the lockfile with `uv lock --check`.
- Built the source and wheel distributions successfully.
- Validated both distributions with Twine.
- Ran the query-generation integration suite: 3 passed.
- Ran the evaluation integration suite: 3 passed.
- Ran the annotation integration suite against a fresh local stack: 19 passed and 3 failed.
- Confirmed that the existing annotation failures block the downstream build and publishing jobs until resolved.


.github/workflows/release.yml
```diff
@@ -1,37 +1,178 @@
-name: Release (Build Only)
+name: Release
 
 on:
+  workflow_dispatch:
+    inputs:
+      build_package:
+        description: "Run full release preflight including package build"
```
ddimmery, line 7: I think this should be more explicit that it _does not_ release, only a push with `v*` tags actually publishes.

.github/workflows/release.yml
```diff
@@ -1,37 +1,178 @@
...
+        run: |
+          docker compose \
+            --project-name pragmata-release-${{ github.run_id }}-${{ strategy.job-index }} \
+            --file deploy/annotation/docker-compose.dev.yml \
+            --env-file deploy/annotation/.env.dev.example \
+            --profile all-bundled \
+            up --detach --pull always --wait --remove-orphans
+
+      - name: Run integration tests
+        run: uv run --python ${{ env.PYTHON_VERSION }} --frozen pytest -o addopts= -m "integration and ${{ matrix.suite.marker }}" -q
```
ddimmery, line 85: As far as I can tell, `PRAGMATA_TEST_ARGILLA_URL` and `PRAGMATA_TEST_ARGILLA_API_KEY` are never exported, so this will currently fail. 

The Makefile defines these explicitly just before running the test, but the CI version does not.

.github/workflows/release.yml
```diff
@@ -1,37 +1,178 @@
...
+        run: |
+          TAG_VERSION="${GITHUB_REF_NAME#v}"
+          PACKAGE_VERSION="$(uv run --no-sync python - <<'PY'
+          import tomllib
+          from pathlib import Path
+
+          pyproject = tomllib.loads(Path("pyproject.toml").read_text())
+          print(pyproject["project"]["version"])
+          PY
+          )"
```
ddimmery, line 128: At this point, afaict the project hasn't been installed because there has been no `uv sync`: see https://github.com/astral-sh/setup-uv#does-setup-uv-also-install-my-project-or-its-dependencies-automatically

.github/workflows/release.yml
```diff
@@ -1,37 +1,178 @@
...
+
+    steps:
+      - name: Download package distributions
+        uses: actions/download-artifact@v8
+        with:
+          name: python-package-distributions
+          path: dist/
+
+      - name: Publish package distributions to PyPI
+        uses: pypa/gh-action-pypi-publish@v1.14.0
```
ddimmery, line 178: Out of an abundance of caution, could pin this by the commit hash rather than version tag (in case `pypa` is compromised).

.github/workflows/release.yml
```diff
@@ -1,37 +1,178 @@
...
+      - name: Build package
+        run: uv build
+
+      - name: Check build artifacts
+        run: |
+          echo "Built files:"
+          ls -l dist/
+
+      - name: Check package metadata
+        run: uvx --from twine twine check dist/*
```
ddimmery, line 147: could make this `--strict`

ddimmery (commented):
For the overall design, I have two points of discussion:
- For running integration tests, since they require making sure things like environment variables are setup, it would be good to have one canonical call (e.g. `make test-integration`) so that nothing gets mistakenly out of sync.
- In case of a flaky failure of the test-suite, the current design will require a new version-tag before trying again (there's no way to release on command). These seem plausible, given (a) Github infra and (b) the dependence on tlmtc[train] / sentence-transformers. It seems this is a real requirement, so any huggingface downtime/instability could also make this flake.

My preference would be to have the testing use the Makefile definitions for consistency, add a ~weekly cadence on automated integration tests. I think it depends on how you want release processes to work for the other question.

I also don't see a reason why this shouldn't also gate on lint/mypy/unit tests, even if those should already be satisfied.

saschagobel:
> * In case of a flaky failure of the test-suite, the current design will require a new version-tag before trying again (there's no way to release on command). These seem plausible, given (a) Github infra and (b) the dependence on tlmtc[train] / sentence-transformers. It seems this is a real requirement, so any huggingface downtime/instability could also make this flake.

Publishing on manual dispatch is intentionally disabled; a pushed version tag is the release authorization. A flaky tagged run does not require a new tag: GitHub reruns retain the original GITHUB_SHA and GITHUB_REF, so the failed job and its dependents can be rerun against the same tag. The Hugging Face dependency can indeed cause a transient failure, but that is recoverable through such a rerun.

saschagobel:
> ... add a ~weekly cadence on automated integration tests.

I’d prefer to keep scheduled integration monitoring separate from the release workflow: this workflow’s purpose is to gate releases, and it already runs the complete integration suite for every release. If proactive integration-health monitoring becomes useful, we can add a dedicated workflow with an appropriate cadence rather than coupling it to release behavior.

saschagobel:
> For the overall design, I have two points of discussion:
> 
> * For running integration tests, since they require making sure things like environment variables are setup, it would be good to have one canonical call (e.g. `make test-integration`) so that nothing gets mistakenly out of sync.
> * In case of a flaky failure of the test-suite, the current design will require a new version-tag before trying again (there's no way to release on command). These seem plausible, given (a) Github infra and (b) the dependence on tlmtc[train] / sentence-transformers. It seems this is a real requirement, so any huggingface downtime/instability could also make this flake.
> 
> My preference would be to have the testing use the Makefile definitions for consistency, add a ~weekly cadence on automated integration tests. I think it depends on how you want release processes to work for the other question.
> 
> I also don't see a reason why this shouldn't also gate on lint/mypy/unit tests, even if those should already be satisfied.

Thanks, I've updated the workflow to use make test-integration as the canonical integration-test entry point and added separate lint, mypy, unit, and packaging gates. Publishing remains tag-only intentionally. I’ve left scheduled integration monitoring out of this release workflow.

ddimmery (approved):
This resolves my concerns!

PR #298, opened 2026-07-19
fix: cast label columns to int64 before majority-vote consolidation

lramir14:
## Summary
Casts source label columns to `int64` in `build_tlmtc_frame()` before majority-vote consolidation. Real annotation exports produce `bool`-typed label columns, and `_consolidate_training_rows()` assigns `int64` majority-vote overrides back into them, which raised `pandas.errors.LossySetitemError` in both directions depending on the duplicate-row path.

## Changes

* Cast the task's label columns to `int64` in `build_tlmtc_frame()`, before `_consolidate_training_rows()` runs.
* No changes to `_consolidate_training_rows()` itself — overrides and source columns now agree in dtype from the start.
* Produces `tlmtc`-compatible `int` `label_*` columns, since `MULTILABEL_SCHEMA` doesn't reliably coerce `bool`.

**Note:** a related bug, `build_tlmtc_frame()` passing all annotation columns to `tlmtc`, where `df_preprocess()`'s `dropna()` drops every row due to legitimately-null non-label columns, is being fixed upstream in `tlmtc` 0.4.2 and is intentionally not included here.

## Testing

* Added a regression test with boolean-dtype labels (`pd.array(..., dtype="boolean")`) across duplicate rows, exercising the majority-vote path that previously raised `LossySetitemError`.
* Full `tests/unit/core/eval/test_transforms.py` suite passes (18/18).

**Status:** Ready for review


src/pragmata/core/eval/transforms.py
```diff
@@ -48,7 +48,12 @@ def build_tlmtc_frame(
     if mode not in {"train", "predict"}:
         raise ValueError(f"Unsupported eval transform mode: {mode!r}.")
 
-    consolidated_frame = _consolidate_training_rows(frame, task=task) if mode == "train" else frame
+    if mode == "train":
+        label_source_columns = LABEL_COLUMNS_BY_TASK[task]
+        frame = frame.astype({column: "int64" for column in label_source_columns})
+        consolidated_frame = _consolidate_training_rows(frame, task=task)
+    else:
+        consolidated_frame = frame
```
saschagobel, line 55: This looks correct but I’d avoid introducing `label_source_columns` and use the mapping directly:

```python
if mode == "train":
    frame = frame.astype(
        {column: "int64" for column in LABEL_COLUMNS_BY_TASK[task]}
    )
    consolidated_frame = _consolidate_training_rows(frame, task=task)
else:
    consolidated_frame = frame
```
This keeps the fix local, avoids a second name similar to the existing `source_label_columns`, and guarantees integer labels before consolidation.

tests/unit/core/eval/test_transforms.py
```diff
@@ -415,3 +415,33 @@ def test_build_tlmtc_frame_rejects_unknown_mode() -> None:
 
     with pytest.raises(ValueError, match="Unsupported eval transform mode"):
         build_tlmtc_frame(frame, task=Task.RETRIEVAL, mode="score")  # type: ignore[arg-type]
+
+
+def test_build_tlmtc_frame_casts_bool_labels_before_majority_consolidation(
```
saschagobel, line 420: Good regression coverage for duplicate consolidation. I’d suggest shortening the docstring to align with the existing module style, using `assert_frame_equal` to check all label values and dtypes together, and adding a small non-duplicate boolean-label test. The latter ensures integer normalization for every training frame, not only when consolidation runs.

saschagobel (approved):
Thanks, this looks good, just two non-blocking suggestions.

src/pragmata/core/eval/transforms.py
lramir14: Thanks for the feedback. Implemented.

tests/unit/core/eval/test_transforms.py
lramir14: implemented. Requesting review.

saschagobel (approved):

PR #299, opened 2026-07-19
chore: prepare package metadata for PyPI

saschagobel:
### Summary

Prepares the package metadata for an eventual initial PyPI release without changing the current package version.

The project now declares its license using the current SPDX-based metadata format and provides focused discovery metadata, supported-environment classifiers, and links to the project’s documentation and development resources.

### Changes

- Replace the deprecated license table with the SPDX expression `MIT`.
- Explicitly include `LICENSE.md` in package distributions.
- Add focused keywords for RAG evaluation and query generation.
- Classify the project as Alpha-stage, console-accessible, and OS-independent.
- Add intended-audience, supported-Python, artificial-intelligence, information-analysis, and Python-library classifiers.
- Add homepage, documentation, repository, issue-tracker, and release-note URLs.
- Require Hatchling 1.27 or newer for PEP 639 and Core Metadata 2.4 support.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -8,8 +8,29 @@ authors =[
   { name = "Luis Fernando Ramírez Ruiz" },
 ]
 readme = "README.md"
-license = { file = "LICENSE.md" }
+license = "MIT"
+license-files = ["LICENSE.md"]
 requires-python = ">=3.12"
+keywords = [
+    "rag",
+    "retrieval-augmented generation",
+    "evaluation",
+    "query generation",
+]
+classifiers = [
+    "Development Status :: 3 - Alpha",
+    "Environment :: Console",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Information Technology",
+    "Intended Audience :: Science/Research",
+    "Operating System :: OS Independent",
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -8,8 +8,29 @@ authors =[
...
-license = { file = "LICENSE.md" }
+license = "MIT"
+license-files = ["LICENSE.md"]
 requires-python = ">=3.12"
+keywords = [
+    "rag",
+    "retrieval-augmented generation",
+    "evaluation",
+    "query generation",
+]
```
lramir14: Wouldn't the project benefit if we also add the annotation interface?

lramir14 (approved):
Thanks for the update and implementation. All is good, only a minor comment on the key words. Approved.

pyproject.toml
saschagobel: Good point, the annotation component should be represented. Since it is specifically tailored to RAG rather than general-purpose annotation, I added “RAG annotation.” I also qualified the evaluation and query-generation keywords so they consistently reflect the project’s scope.

PR #300, opened 2026-07-21
fix(annotation): isolate integration tests on ephemeral Argilla stack

henrycgbaker:
### Summary

There were 3 failing integration tests failed on shared state & the annotation integration suite ran destructive setup/teardown against whatever was on `localhost:6900` (the live annotation stack). This makes suite run against an ephemeral, port-isolated stack that is never the live one, and fixes the failures.

### Changes

- **Fix the 3 failing tests (isolation bugs):**
  - Per-test `base_dir` gives each test a fresh partition manifest (previously all shared one under cwd), fixing the record-count and idempotency assertions.
  - The per-workspace `min_submitted` carve-out test now feeds its topology through a YAML config, so the dataset that `import_records` lazily creates actually gets the intended `min_submitted` (`setup_workspaces` only creates workspaces).
- **Ephemeral, isolated test stack:** tests resolve the target via `PRAGMATA_TEST_ARGILLA_URL` / `PRAGMATA_TEST_ARGILLA_API_KEY`. `make test-integration` / `test-all` / `test-stack` stand up a throwaway Compose project on a non-default port (`:6901`) and tear it down (`-v`) on exit. The `-p` project scope keeps containers/volumes disjoint from the live `annotation` stack on `:6900`.
- **Fail-closed guard:** unless `PRAGMATA_TEST_ARGILLA_URL` is set, a run that selects annotation tests **aborts at collection with a clear `pytest.UsageError` (exit 4)** - safe (raised before any fixture, so the destructive teardown never fires) and loud (a CI gate cannot pass having silently run zero annotation tests). Ordinary unit runs are unaffected: the default `-m 'not integration'` filter removes annotation items before the guard sees them.
- **Cohesion:** rename `make test` → `test-unit` (unit-only, no stack); share `client`/`base_dir` via a single `tests/integration/conftest.py` (also fixes a readiness-gate drift where the calibration client skipped the stack-ready check).

### Test plan

- [x] `make test-unit` - 1192 passed
- [x] `make test-integration` against ephemeral `:6901` - all annotation tests pass; ephemeral stack torn down; live `:6900` untouched (API still 200)
- [x] Fail-closed verified: with `PRAGMATA_TEST_ARGILLA_URL` unset, `pytest -m integration` errors with `UsageError` (exit 4) and runs no tests; `--setup-plan` confirms no fixture instantiates before the abort
- [x] `ruff check` clean
- [x] CI passes

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,16 @@
 COMPOSE_FILE := deploy/annotation/docker-compose.dev.yml
 ENV_FILE     := deploy/annotation/.env
 ENV_EXAMPLE  := deploy/annotation/.env.dev.example
-TEST_PROJECT := pragmata-test
 
-.PHONY: docker-up docker-up-external-pg docker-up-external-es docker-up-external docker-down docker-down-clean docker-stop docker-logs docker-status ensure-env lint type-check test-stack test test-integration test-all ci
+# One throwaway stack for all test targets — an isolated Compose project on a
+# non-default host port so its destructive setup/teardown can never touch the
+# live stack (Compose project 'annotation') on :6900. The -p project scope
+# always wins over the shared compose file, so containers/volumes stay disjoint.
+# test-stack and test-integration/test-all share it (they never run at once).
+TEST_STACK_PROJECT := pragmata-test
+TEST_STACK_PORT    := 6901
+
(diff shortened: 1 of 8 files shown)


saschagobel (approved):
All good, thanks.

PR #304, opened 2026-07-28
feat(eval): score prediction runs via prediction_id

### LINKED ISSUE #303: eval: support scoring prediction runs
## Summary

- `pragmata eval score` cannot yet score a prediction run
- `resolve_eval_score_input` accepts a `prediction_id` selector but raises `NotImplementedError`
- wiring it up needs a score-input resolver (and a metadata sidecar for task/provenance) that don't exist yet

## Current state

- **Prediction output is already run-addressable.** `predict_labels` (`src/pragmata/api/eval.py`) writes `eval/prediction_outputs/<run_id>/predictions.csv` (+ `probabilities.csv`) - asserted in `tests/integration/test_eval.py:404-413`.
- **But that dir is keyed by the *evaluator* run id**, not a distinct prediction-run id.`run_tlmtc_predict` forwards `evaluator_run_id` to tlmtc as `run_id`  (`src/pragmata/core/eval/tlmtc_adapters.py:138`). Two predictions from the same evaluator  would collide under one dir.
- **Predict persists no pragmata metadata.** There is `EvalTrainMeta`(`src/pragmata/core/schemas/eval_output.py`) but no `EvalPredictMeta`, so a scored prediction run has no sidecar to validate task/provenance against (cf. `EvalTrainMeta` beside `pragmata_train.meta.json`).
- **The import layer is ready.** `import_eval_score_frame`  (`src/pragmata/core/eval/imports.py`) already branches on `source.kind == "model_prediction"`  and restores task text columns. No change needed there.

## Design decision 

**`prediction_id` = evaluator run id**, driven by a tlmtc constraint (verified against tlmtc 0.4.2 source).

`predict_tlmtc(run_id=...)` takes a **single** `run_id` that does double duty: it selects the trained evaluator (`train_outputs/<run_id>/`) *and* names the output (`prediction_outputs/<run_id>/`, `tlmtc/paths.py:369`). There is no separate output-id parameter

This mirrors pragmata's existing **train** identity exactly: reuse tlmtc's `run_id` and drop a `pragmata_predict.meta.json` beside tlmtc's artifacts, resolved directly by the explicit `prediction_id`. `predict-labels` CLI (#295) already exposes this addressing (`prediction_outputs/<train-run-id>/`).

**Accepted trade-off:** re-predicting with the same evaluator overwrites its output dir - this is tlmtc's own behaviour and is documented, not worked around. A distinct prediction-run id would be a tlmtc feature request, not a pragmata hack.

## Work

1. **Predict-side identity.** Add an `EvalPredictMeta` schema (`run_id` = evaluator run id, `task`, `created_at`, input provenance) and have `predict_labels` persist `pragmata_predict.meta.json`  into `prediction_outputs/<run_id>/` beside the tlmtc artifacts (mirrors the train-meta pattern).
2. **Resolver.** Replace the `NotImplementedError` branch in `resolve_eval_score_input` (`src/pragmata/core/paths/eval_paths.py`) with resolution of `prediction_outputs/<prediction_id>/predictions.csv`, returning `EvalScoreInput(source=ScoreInputSource(kind="model_prediction", ...))` — the exact kind the importer already expects. Resolve by the explicit `prediction_id` directly (no glob); the run's `pragmata_predict.meta.json` is read only to confirm it was produced for the requested task. No latest-prediction finder is added: `prediction_id` is always explicit, and the no-selector fallback stays "latest annotation export", so there is no caller for one (thanks @saschagobel).
3. **CLI.** `--prediction-id` already exists on `pragmata eval score` (added in #280); it simply starts working. Minimal/no CLI change.
4. **Tests.**

henrycgbaker:
## Summary

- `pragmata eval score --prediction-id <id>` now works: `resolve_eval_score_input` resolves a completed prediction run's `predictions.csv` and returns `ScoreInputSource(kind="model_prediction", ...)`, replacing the `NotImplementedError` branch. The importer already handled that kind, so no import-layer change was needed.
- Gives prediction runs a discoverable identity: new `EvalPredictMeta` schema, and `predict_labels` persists a `pragmata_predict.meta.json` sidecar into `prediction_outputs/<run_id>/` beside the tlmtc artifacts (mirrors the existing train-meta pattern). `prediction_id` is the evaluator run id, per tlmtc's single-`run_id` constraint. `unlabeled_data_path` is a required field: prediction input is mandatory on `EvalPredictSettings` and resolved before inference, so a completed run always records which input it labeled.
- Resolution reads that sidecar and rejects a run produced for a different task (`ValueError`), rather than silently scoring mismatched labels. Missing sidecar or missing `predictions.csv` raise `FileNotFoundError`.
- `export_eval_train_meta` generalised to `export_eval_meta(meta: EvalTrainMeta | EvalPredictMeta, path)` so train and predict share one writer without widening the type boundary.
- Stale documentation brought in line with the implementation: the `--prediction-id` CLI help, the `EvalScoreSettings` docstring, and `docs/design/eval-scoring-metrics.md` no longer describe `prediction_id` as unsupported or the annotation-export fallback as interim. The `eval score` docstring's claim that the selectors have a precedence order is also corrected - they are mutually exclusive, matching Open decision 1.

Deviation from the issue's original plan (I updated #303 body so now aligned): no `find_latest_prediction_run` resolver was added. The no-selector fallback stays "latest annotation export" (unchanged), so nothing needs to glob for a latest prediction run - `--prediction-id` is always explicit.

Closes #303

### DIFF
--- a/src/pragmata/core/eval/export.py
+++ b/src/pragmata/core/eval/export.py
@@ -3,7 +3,7 @@
 from pathlib import Path
 
 from pragmata.core.atomic_io import atomic_write_json
-from pragmata.core.schemas.eval_output import EvalTrainMeta
+from pragmata.core.schemas.eval_output import EvalPredictMeta, EvalTrainMeta
 
 
 def export_eval_train_meta(
@@ -18,5 +18,22 @@ def export_eval_train_meta(
     """
     if not path.parent.is_dir():
         raise FileNotFoundError(f"Eval train metadata parent directory does not exist: {path.parent}")
+
+    atomic_write_json(meta.model_dump(mode="json"), path)
+
+
+def export_eval_predict_meta(
+    meta: EvalPredictMeta,
+    path: Path,
+) -> None:
+    """Write Pragmata-owned evaluator prediction metadata to disk as JSON.
+
+    Args:
+        meta: Validated evaluator prediction metadata to persist.
+        path: Destination path for the JSON sidecar, under the tlmtc
+            prediction-run directory.
+    """
+    if not path.parent.is_dir():
+        raise FileNotFoundError(f"Eval predict metadata parent directory does not exist: {path.parent}")
 
     atomic_write_json(meta.model_dump(mode="json"), path)
--- a/src/pragmata/core/eval/export.py
+++ b/src/pragmata/core/eval/export.py
@@ -2,38 +2,22 @@
 
 from pathlib import Path
 
+from pydantic import BaseModel
+
 from pragmata.core.atomic_io import atomic_write_json
-from pragmata.core.schemas.eval_output import EvalPredictMeta, EvalTrainMeta
 
 
-def export_eval_train_meta(
-    meta: EvalTrainMeta,
-    path: Path,
-) -> None:
-    """Write Pragmata-owned evaluator training metadata to disk as JSON.
+def export_eval_meta(meta: BaseModel, path: Path) -> None:
+    """Write Pragmata-owned evaluator run metadata to disk as JSON.
+
+    Shared by train and predict runs (``EvalTrainMeta`` / ``EvalPredictMeta``);
+    the sidecar is written beside the run's tlmtc artifacts.
 
     Args:
-        meta: Validated evaluator training metadata to persist.
+        meta: Validated evaluator run metadata to persist.
         path: Destination path for the JSON sidecar.
     """
     if not path.parent.is_dir():
-        raise FileNotFoundError(f"Eval train metadata parent directory does not exist: {path.parent}")
-
-    atomic_write_json(meta.model_dump(mode="json"), path)
-
-
-def export_eval_predict_meta(
-    meta: EvalPredictMeta,
-    path: Path,
-) -> None:
-    """Write Pragmata-owned evaluator prediction metadata to disk as JSON.
-
-    Args:
-        meta: Validated evaluator prediction metadata to persist.
-        path: Destination path for the JSON sidecar, under the tlmtc
-            prediction-run directory.
-    """
-    if not path.parent.is_dir():
-        raise FileNotFoundError(f"Eval predict metadata parent directory does not exist: {path.parent}")
+        raise FileNotFoundError(f"Eval metadata parent directory does not exist: {path.parent}")
 
     atomic_write_json(meta.model_dump(mode="json"), path)
--- a/src/pragmata/core/eval/export.py
+++ b/src/pragmata/core/eval/export.py
@@ -2,12 +2,11 @@
 
 from pathlib import Path
 
-from pydantic import BaseModel
-
 from pragmata.core.atomic_io import atomic_write_json
+from pragmata.core.schemas.eval_output import EvalPredictMeta, EvalTrainMeta
 
 
-def export_eval_meta(meta: BaseModel, path: Path) -> None:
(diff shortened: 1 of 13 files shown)


tests/unit/cli/test_cli_eval.py
saschagobel, line 1: Since this PR makes `--prediction-id` functional and updates its CLI test, the actual option help in `src/pragmata/cli/commands/eval.py` also requires an update. It still says “Not yet supported” and refers to the nonexistent eval predict command.

src/pragmata/core/paths/eval_paths.py
```diff
@@ -415,26 +434,30 @@ def resolve_eval_score_input(
     for the task is used, mirroring ``resolve_eval_train_paths``.
 
     Direct paths and annotation exports are already Pragmata-shaped and score
-    without further preparation. Scoring a prediction run (``prediction_id``) is
-    not currently supported and raises ``NotImplementedError`` (see #303). With no
-    selector, the latest annotation export is used.
+    without further preparation. A ``prediction_id`` selects a completed
```
saschagobel, line 437: The remaining repository documentation should match this implementation. 
There are three stale locations:

• EvalScoreSettings lines 105–106 still call the annotation-export fallback interim “until eval predict lands”.
• Its prediction_id documentation at lines 118–119 still says it is unsupported.
• docs/design/eval-scoring-metrics.md lines 152 and 194 still say prediction_id is not wired and that find_latest_prediction_run needs to be added.

These now contradict both this resolver and the agreed design: prediction_id is explicitly resolved, no latest-prediction finder is needed, and the no-selector fallback remains the latest annotation export.

src/pragmata/core/schemas/eval_output.py
```diff
@@ -70,6 +70,22 @@ class EvalTrainMeta(BaseModel):
...
+    the same ``run_id`` it loads the evaluator from, so a run is identified by
+    which evaluator produced it.
+    """
+
+    model_config = ConfigDict(extra="forbid", frozen=True)
+
+    run_id: str
+    created_at: datetime = Field(default_factory=lambda: datetime.now(UTC))
+    task: Task
+    unlabeled_data_path: str | None = None
```
saschagobel: Prediction input is required by `EvalPredictSettings`, resolved before inference, and always available when this metadata is written. Making this optional allows a “completed” sidecar without the input provenance required by the issue. Could this be a required str instead?

src/pragmata/core/eval/export.py
```diff
@@ -2,21 +2,22 @@
...
 from pragmata.core.atomic_io import atomic_write_json
-from pragmata.core.schemas.eval_output import EvalTrainMeta
 
 
-def export_eval_train_meta(
-    meta: EvalTrainMeta,
-    path: Path,
-) -> None:
-    """Write Pragmata-owned evaluator training metadata to disk as JSON.
+def export_eval_meta(meta: BaseModel, path: Path) -> None:
```
saschagobel: I'd suggest to keep the type boundary specific to the metadata this writer actually supports. A union would preserve the shared implementation without weakening type safety:
meta: EvalTrainMeta | EvalPredictMeta

The corresponding test names/docstrings also still refer to the removed export_eval_train_meta and export_eval_predict_meta functions and should be updated with this refactor.

saschagobel (changes requested):
The core implementation looks sound and well tested, but a few inconsistencies remain across the CLI/docs and the new metadata contract. Requesting these focused cleanups before approval.

saschagobel (approved):

PR #305, opened 2026-07-29
feat(eval): guard scoring against incomplete retrieval panels

henrycgbaker:
## Summary

- `eval score` silently accepts partial retrieval panels. 
- -Annotation exports carry `panel_complete` and `n_retrieved_chunks`, but nothing in the scoring path reads either - the score schema doesn't require them, and `grouping.py` averages whatever chunks are present. 
- A partial panel is not a smaller sample of the same quantity, it changes the metric: precision@K over 2 labeled chunks of a K=5 retrieval has the wrong denominator, and the rank-sensitive metrics (MRR, NDCG) are biased upward because annotators work top-down, so unjudged low-rank chunks can never lower the score.

## Behaviour: three mutually exclusive modes

| Mode | Result |
|---|---|
| default | `EvalInputSchemaError` naming the first short panel |
| `--skip-incomplete-panels` | drop short panels, score the rest; the retrieval report records `n_panels_skipped` and the CLI echoes it |
| `--allow-incomplete-panels` | score everything as-is, for callers who accept the bias |

Passing both flags raises (guard + `EvalScoreSettings` validator). Skipping every panel is an error, not an empty success. Frames without a usable K (column absent on direct-path/prediction inputs, or the `-1` pre-backfill sentinel) pass with a `logger.warning`: completeness is unknowable there rather than violated. Grounding/generation are unaffected (no panel notion).

The default stays **fail** because neither remedy is safe to apply on the caller's behalf: skipping changes the population (completion is not random), allowing changes the metric.

## Test plan

- [x] Guard: rejects a partial panel; accepts a complete one; skip drops and counts; skip-with-nothing-left raises; both-flags raises; missing-K and `-1`-sentinel warning paths; grounding unaffected
- [x] Settings: mutual-exclusion validator test
- [x] CLI: help listing, UNSET mapping, delegation for both flags
- [x] Full suite: 1261 passed; ruff format/check and mypy clean

### DIFF
--- a/src/pragmata/api/eval.py
+++ b/src/pragmata/api/eval.py
@@ -248,8 +248,8 @@ def score(
         ci: Confidence level for every interval. Defaults to 0.95.
         seed: Optional RNG seed for reproducible bootstrap intervals.
         allow_incomplete_panels: Score retrieval panels whose labeled chunks do not
-            cover ``n_retrieved_chunks``. Defaults to False: a partial panel changes
-            the @K denominators and biases the rank-sensitive metrics upward.
+            cover ``n_retrieved_chunks``. Defaults to ``False`` because a partial panel
+            changes the @K denominators and biases the rank-sensitive metrics upward.
         config_path: Path to a YAML configuration file.
 
     Returns:
--- a/src/pragmata/api/eval.py
+++ b/src/pragmata/api/eval.py
@@ -220,6 +220,7 @@ def score(
     n_resamples: int | Unset = UNSET,
     ci: float | Unset = UNSET,
     seed: int | Unset = UNSET,
+    skip_incomplete_panels: bool | Unset = UNSET,
     allow_incomplete_panels: bool | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
 ) -> ScoreReport:
@@ -247,9 +248,14 @@ def score(
         n_resamples: Bootstrap iterations for the continuous metrics. Defaults to 1000.
         ci: Confidence level for every interval. Defaults to 0.95.
         seed: Optional RNG seed for reproducible bootstrap intervals.
+        skip_incomplete_panels: Drop retrieval panels whose labeled chunks do not
+            cover ``n_retrieved_chunks`` and score the rest; the retrieval report
+            records the drop count as ``n_panels_skipped``. Defaults to ``False``.
+            Mutually exclusive with ``allow_incomplete_panels``.
         allow_incomplete_panels: Score retrieval panels whose labeled chunks do not
             cover ``n_retrieved_chunks``. Defaults to ``False`` because a partial panel
             changes the @K denominators and biases the rank-sensitive metrics upward.
+            Mutually exclusive with ``skip_incomplete_panels``.
         config_path: Path to a YAML configuration file.
 
     Returns:
@@ -275,6 +281,7 @@ def score(
             "n_resamples": n_resamples,
             "ci": ci,
             "seed": seed,
+            "skip_incomplete_panels": skip_incomplete_panels,
             "allow_incomplete_panels": allow_incomplete_panels,
         },
     )
@@ -289,10 +296,11 @@ def score(
             export_id=settings.export_id,
             prediction_id=settings.prediction_id,
         )
-        frame = import_eval_score_frame(
+        frame, n_panels_skipped = import_eval_score_frame(
             path=resolved.input_csv,
             task=settings.task,
             source=resolved.source,
+            skip_incomplete_panels=settings.skip_incomplete_panels,
             allow_incomplete_panels=settings.allow_incomplete_panels,
         )
         report = build_score_report(
@@ -303,6 +311,7 @@ def score(
             seed=settings.seed,
             source=resolved.source,
             created_at=datetime.now(UTC),
+            n_panels_skipped=n_panels_skipped,
         )
         match settings.task:
             case Task.RETRIEVAL:
--- a/src/pragmata/api/eval.py
+++ b/src/pragmata/api/eval.py
@@ -220,6 +220,7 @@ def score(
     n_resamples: int | Unset = UNSET,
     ci: float | Unset = UNSET,
     seed: int | Unset = UNSET,
+    allow_incomplete_panels: bool | Unset = UNSET,
     config_path: str | Path | Unset = UNSET,
 ) -> ScoreReport:
     """Score labeled eval data into corpus metrics with confidence intervals.
@@ -246,6 +247,9 @@ def score(
         n_resamples: Bootstrap iterations for the continuous metrics. Defaults to 1000.
         ci: Confidence level for every interval. Defaults to 0.95.
         seed: Optional RNG seed for reproducible bootstrap intervals.
+        allow_incomplete_panels: Score retrieval panels whose labeled chunks do not
+            cover ``n_retrieved_chunks``. Defaults to False: a partial panel changes
+            the @K denominators and biases the rank-sensitive metrics upward.
         config_path: Path to a YAML configuration file.
 
     Returns:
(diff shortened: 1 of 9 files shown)


src/pragmata/api/eval.py
```diff
@@ -246,6 +248,14 @@ def score(
...
         ci: Confidence level for every interval. Defaults to 0.95.
         seed: Optional RNG seed for reproducible bootstrap intervals.
+        skip_incomplete_panels: Drop retrieval panels whose labeled chunks do not
+            cover ``n_retrieved_chunks`` and score the rest; the retrieval report
+            records the drop count as ``n_panels_skipped``. Defaults to ``False``.
+            Mutually exclusive with ``allow_incomplete_panels``.
+        allow_incomplete_panels: Score retrieval panels whose labeled chunks do not
+            cover ``n_retrieved_chunks``. Defaults to ``False`` because a partial panel
+            changes the @K denominators and biases the rank-sensitive metrics upward.
+            Mutually exclusive with ``skip_incomplete_panels``.
```
saschagobel, line 258: I do not think we should support `allow_incomplete_panels`. This option knowingly
produces biased retrieval metrics while returning them as ordinary @K scores,
which makes an invalid result look authoritative.

The existing behaviors should be sufficient: fail by default, or explicitly
skip incomplete panels and report how many were excluded. If partial annotation
coverage needs to be inspected, that should be a separate diagnostic rather
than a supported path through the canonical scoring API.

Could we remove `allow_incomplete_panels` and retain only the strict default plus
`skip_incomplete_panels`?

src/pragmata/core/eval/imports.py
```diff
@@ -113,10 +135,96 @@ def _guard_unique_scoring_units(frame: pd.DataFrame, *, task: Task) -> None:
...
 def _reject_duplicates(frame: pd.DataFrame, keys: list[str], task: Task, unit: str) -> None:
     duplicated = frame.duplicated(subset=keys, keep=False)
     if duplicated.any():
         raise EvalInputSchemaError(
             f"Scoring input for {task.value} has {int(duplicated.sum())} row(s) with a duplicate "
             f"{unit} key {tuple(keys)}; each unit must be unique so metric denominators are not double-counted."
         )
+
+
+def _panels_with_known_k(frame: pd.DataFrame) -> pd.DataFrame:
```
saschagobel, line 217: Why do we support retrieval panels with unknown K at all?

The comments describe this primarily as compatibility for records created before
`n_retrieved_chunks` was introduced. Given that the project is still pre-alpha,
is preserving that legacy-data path worth the additional complexity and weaker
scoring guarantees?

Could we instead require a positive `n_retrieved_chunks` for every retrieval
panel and fail validation when it is absent? Existing data could be migrated or
re-exported rather than making the canonical scoring path support unverifiable
panels indefinitely.

src/pragmata/core/eval/imports.py
```diff
@@ -113,10 +135,96 @@ def _guard_unique_scoring_units(frame: pd.DataFrame, *, task: Task) -> None:
...
+    exports). Where no panel carries a known K the completeness of the input is
+    unknowable here, so the frame passes with a warning rather than a hard failure -
+    direct-path and prediction inputs are not required to carry export metadata.
+    """
+    if skip_incomplete and allow_incomplete:
+        raise ValueError("skip_incomplete_panels and allow_incomplete_panels are mutually exclusive; pass at most one.")
+    if task != Task.RETRIEVAL or allow_incomplete:
+        return frame, 0
+    panels = _panels_with_known_k(frame)
+    if panels.empty:
```
saschagobel, line 174: If unknown-K panels remain supported, is there a bug in how mixed inputs are
handled here?

`_panels_with_known_k()` removes unknown-K panels only from the derived `panels`
summary, not from the original scoring `frame`. The warning then fires only when
the summary is completely empty. As a result, when an input contains both known-
and unknown-K panels, the unknown panels are neither checked nor reported but
are still scored.

This is particularly problematic with `skip_incomplete_panels`: all known
incomplete panels can be removed while an unverifiable panel remains, after
which the CLI claims that `n` counts completed panels only.

src/pragmata/core/eval/imports.py
```diff
@@ -113,10 +135,96 @@ def _guard_unique_scoring_units(frame: pd.DataFrame, *, task: Task) -> None:
...
+
+    ``n_retrieved_chunks`` is absent from direct-path and prediction inputs, and is the
+    ``-1`` sentinel on export records predating the metadata backfill; both mean "K
+    unknown here", so such panels cannot be checked for completeness.
+    """
+    if "n_retrieved_chunks" not in frame.columns:
+        return pd.DataFrame(columns=["n_chunks", "k"])
+    per_query = frame.groupby("record_uuid").agg(
+        n_chunks=("chunk_id", "nunique"),
+        k=("n_retrieved_chunks", "max"),
```
saschagobel, line 228: Suggestion: since this logic consumes `n_retrieved_chunks`, could we also add
that column to `RETRIEVAL_SCORE_SCHEMA` instead of allowing it to pass through
as unvalidated extra metadata?

Ideally, the schema would require a positive integer, with an additional
panel-level check that all rows for a `record_uuid` carry the same K. This would
keep the validation contract aligned with the data the scorer relies on here.

src/pragmata/core/eval/imports.py
```diff
@@ -113,10 +135,96 @@ def _guard_unique_scoring_units(frame: pd.DataFrame, *, task: Task) -> None:
...
 
 
+def _guard_complete_panels(
+    frame: pd.DataFrame,
+    *,
+    task: Task,
+    skip_incomplete: bool,
+    allow_incomplete: bool,
+) -> tuple[pd.DataFrame, int]:
+    """Apply the incomplete-retrieval-panel policy; return the frame and drop count.
```
saschagobel, line 145: Suggestion: could we reduce this docstring substantially? It currently repeats
the API documentation, CLI behavior, error message, and implementation details.

A short description of the policy, the default behavior, and the meaning of the
returned drop count should be sufficient here. In particular, the detailed bias
rationale does not need to be repeated at every layer.

src/pragmata/core/settings/eval_settings.py
```diff
@@ -122,6 +122,10 @@ class EvalScoreSettings(ResolveSettings):
         n_resamples: Bootstrap iterations for the continuous metrics' CIs.
         ci: Confidence level for every reported interval (e.g. 0.95 for 95%).
         seed: Optional RNG seed for reproducible bootstrap intervals.
+        allow_incomplete_panels: Score retrieval panels whose labeled chunks do not
```
saschagobel, line 125: Suggestion: we should document `skip_incomplete_panels` here as well. It is a
public setting defined below, but the Attributes section currently documents
only `allow_incomplete_panels`.

src/pragmata/core/eval/scoring.py
```diff
@@ -67,8 +67,13 @@ def build_score_report(
     seed: int | None,
     source: ScoreInputSource,
     created_at: datetime,
+    n_panels_skipped: int = 0,
 ) -> ScoreReport:
-    """Assemble the task-specific report from per-query values (pure, I/O-free)."""
+    """Assemble the task-specific report from per-query values (pure, I/O-free).
+
+    ``n_panels_skipped`` is recorded on the retrieval report only; the other tasks
+    have no panel notion and reject the field by schema.
```
saschagobel, line 75: This does not appear to match the implementation. For non-retrieval tasks,
`n_panels_skipped` is not passed to the report constructor, so it is ignored
rather than rejected by the schema.

saschagobel (changes requested):
Thanks for adding safeguards around retrieval panel completeness. The overall direction makes sense, but I’m requesting changes because the API contract and input-validation guarantees need to be tightened before this is ready to merge.

